PR #13, opened 2022-12-17
feat(german): add first german intents

mib1185:
add first few intents


tests/de/homeassistant_HassToggle.yaml
```diff
@@ -0,0 +1,11 @@
+language: de
+tests:
+  - sentences:
+      - "schalte den Küchenschalter"
+      - "schalte den Küchenschalter um"
+      - "betätige den Küchenschalter"
+      - "bitte schalte den Küchenschalter um"
```
balloob, line 7: Butte is a skip word and doesn’t need to be added in sentences

mib1185: ah ok, I thought to test it at least once 🙈 
Will remove it

sentences/de/_common.yaml
```diff
@@ -0,0 +1,15 @@
+language: de
+responses:
+  errors:
+    no_intent: "Entschuldigung, das habe ich nicht verstanden"
+    no_area: "Den Bereich {{ area }} gibt es nicht"
+    no_domain: "{{ domain }} gibt es nicht in dem Bereich {{ area }}"
```
syssi, line 6:
```suggestion
    no_domain: "{{ domain }} gibt es nicht im Bereich {{ area }}"
```

tests/de/_common.yaml
```diff
@@ -0,0 +1,23 @@
+language: de
+areas:
+- name: Küche
+  id: kitchen
+- name: Wohnstube
```
syssi, line 5:
```suggestion
- name: Wohnzimmer
```

Doesn't make a difference here. IMO this is the more common word.

sentences/de/_common.yaml
mib1185: this merged version should only be used, when you talk about an not specific defined noun (_mostly a generic place description_), but when the noun is well known or more specific defined, then you should not merge it.

examples:
### merged version, because generic noun/place
- Ich war im Kino 
- Wir haben uns im Stadtzentrum getroffen

### not merged, because well known or more specific defined noun/place
- In dem Supermarkt würde ich nicht einkaufen
- In diesem Supermarkt würde ich nicht einkaufen
- In dem Winter 2010 war es sehrt kalt

tests/de/_common.yaml
mib1185: it depends, in which region you are living 🙈 
But yes you're right, "Wohnzimmer" is the common

spacegaier: I'd agree with that. Wohnzimmer is more common.

sentences/de/_common.yaml
syssi: I understand your explanation. My suggestion is just a gut feeling because the sentence "Ventilator gibt es nicht in dem Bereich Küche" sounds strange compared to "Ventilator gibt es nicht im Bereich Küche".

Feel free to close the conversation without touching the PR.

mib1185: @spacegaier any toughs on this? 🙂

spacegaier: I am leaning towards the proposal from syssi. I get your point from a purely grammatical point of view, but from a gut feeling perspective "in dem Bereich Küche" sounds actually a bit weird. And since we are talking about conversational language "im" feels more suitable to me.

mib1185: Ok, you both have convinced me

mib1185:
@syssi @spacegaier many thanks for your valuable inputs 🙂 i've also added some sentences for the fan and light handling. Hope it's ok to request you for a review

spacegaier:
Ich bin gerade aufm Sprung :) => I am going to have another look tomorrow evening most likely.

tests/de/homeassistant_HassToggle.yaml
balloob: Oh snap, you're right, this is a test. I didn't notice that. You should keep it here 🙈

mib1185:
@balloob tests are failing after c24505fdd63ac0a4063f20504b6b2d5eaef46771 with
`ERROR: /home/runner/work/intents/intents/tests/test_language_sentences.py Imports are incorrectly sorted and/or formatted.`

thecode:
This was merged with `isort` error, I will try to add a pre-commit to fix such errors locally

synesthesiam:
I fixed the error in a different branch, so I went ahead with the merge here. Running `script/format` in the pre-commit phase should do it for the future.

PR #15, opened 2022-12-18
Update Dutch test coverage

balloob:
Update Dutch coverage to help unblock #14

### DIFF
--- a/sentences/nl/cover_HassOpenCover.yaml
+++ b/sentences/nl/cover_HassOpenCover.yaml
@@ -6,14 +6,13 @@ intents:
           - "open <name>"
           - "open <name> in <area>"
       - sentences:
-          - "open [de] garage deur"
-          - "open [de] garage"
+          - "open [de] garage [deur]"
           - "(mag | doe) [de] garage open"
         slots:
           device_class: "garage_door"
       - sentences:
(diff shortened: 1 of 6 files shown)


sentences/nl/cover_HassOpenCover.yaml
```diff
@@ -6,14 +6,13 @@ intents:
...
-          - "open [de] garage deur"
-          - "open [de] garage"
+          - "open [de] garage [deur]"
           - "(mag | doe) [de] garage open"
         slots:
           device_class: "garage_door"
       - sentences:
           - open (het gordijn | de gordijnen | de luxaflex | het rolluik) in <area>
-          - mogen de gordijen open in <area>
+          - mogen de gordijnen open in <area>
```
balloob, line 15: I found an actual bug :)

tests/nl/_common.yaml
```diff
@@ -21,3 +21,7 @@ entities:
     id: fan.plafond
     area: woonkamer
     domain: fan
+  - name: Het Stofje
```
balloob, line 24: Not happy with this. It's a stupid name. However, we have a bunch of names hardcoded that include common names. Couldn't think of anything better.

PR #73, opened 2022-12-21
Add .vscode to .gitignore

berendhaan:
Add .vscode folder to .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ htmlcov
 venv/
 .mypy_cache/
 *.egg-info/
+.vscode/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ htmlcov
 venv/
 .mypy_cache/
 *.egg-info/
+.vscode/
```
frenck, line 16: Is should not exclude `tasks.json` as this repository provides that one.

```suggestion

# Visual Studio Code
.vscode/*
!.vscode/tasks.json
```

berendhaan: The `settings.json` was the trigger for me, can also only exclude that.

PR #74, opened 2022-12-21
[NL] Add some Dutch sentences for climate

berendhaan:
Just some simple few sentences to start with. The naming conventions weren't very clear to me, may need some work.


.gitignore
```diff
@@ -13,3 +13,4 @@ htmlcov
 venv/
 .mypy_cache/
 *.egg-info/
+.vscode/
```
frenck, line 16: This isn't related to this PR and should not be in here.

TheFes:
Where you actually say `temparatuur` you should not add the the angle brackets (`<` and `>`)
So:

berendhaan:
Got feedback from @TheFes , removed the brackets around the word 'temperatuur'

PR #80, opened 2022-12-22
Added some new sentences and tests for EN US

waltlillyman:
US English speakers can be flexible in use of the article, "the", especially with voice assistants.
I've tried to include & exclude it in my sentences and tests, based on what I've heard here in the heartland of the mid-western USA, St. Louis, Missouri.


.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
balloob, line 1: Could you remove this file from your PR. We have a PR open to have VSCode ignore this.

balloob (approved):
Nice!

waltlillyman:
Merged, so moving on.

waltlillyman:
Reopened !

sentences/en/climate_HassClimateGetTemperature.yaml
```diff
@@ -5,3 +5,4 @@ intents:
       - sentences:
           - "<what_is> [the] <temp>"
           - "<what_is> [the] <temp> in <area>"
+          - "<what_is> [the] <area> <temp>"
```
synesthesiam, line 8: Hi @waltlillyman, this is definitely not obvious but `<area>` and `<name>` already include an optional `[the]`: https://github.com/home-assistant/intents/blob/e848342fed0328e312ca9efca4e283dd5ee8a546/sentences/en/_common.yaml#L42

synesthesiam (changes requested):

sentences/en/climate_HassClimateGetTemperature.yaml
synesthesiam: We will also need to add at least as many test cases as there are input sentences for each intent :+1:

waltlillyman: Ok, I can resume tomorrow AM, US CT!

synesthesiam: Thanks! :slightly_smiling_face:

waltlillyman: Made those corrections and tested locally successfully.

waltlillyman:
@synesthesiam @balloob I removed all redundant [the] from before {name} and {area}, and ensured that each sentence had at least one corresponding test case. I ran tests locally successfully. Please take a look!

sentences/en/climate_HassClimateGetTemperature.yaml
balloob: This is not solved yet. We should remove `[the]` as it's included in the `<area>` expansion.

```suggestion
          - "<what_is> <area> <temp>"
```

PR #83, opened 2022-12-22
Updated Hungarian

nagyrobi:


synesthesiam:
@tomatow can you take a look here? The `<name>` expansion rule is missing in `_common.yaml`. If not needed, `<name>` should be changed to `{name}` so it refers to the list HA will generate.

Also, there must be at least as many tests as there are sentence templates.

PR #99, opened 2022-12-23
Use better fixture name for cover

balloob:
Use a common fixture name for the cover entity for both Dutch and English.

CC @TheFes @waltlillyman


waltlillyman (approved):
Thank you!

Issue #26, opened 2022-12-19
[DE] Expansion rules should use native language


ludeeus:
https://github.com/home-assistant/intents/blob/e3e111b499bb76c72b69a2cbd160daec79266abe/sentences/de/_common.yaml#L11-L18

<https://github.com/home-assistant/intents/blob/main/sentences/README.md#sentence-templates>

PR #105, opened 2022-12-23
[DE] use native languag for expansion rules

mib1185:
- fixes: #26
- remove unused expansion rules
- simplified the "all variants" rules
- extend the former `do` rule


sentences/de/fan_HassTurnOff.yaml
```diff
@@ -3,19 +3,19 @@ intents:
...
       - sentences:
-          - "schalte <fan_variants> <area> <turn_off>"
-          - "schalte <area> <fan_variants> <turn_off>"
-          - "stoppe <fan_variants> <area> [<turn_off>]"
-          - "stoppe <area> <fan_variants> [<turn_off>]"
-          - "(mach | mache) <fan_variants> <area> <turn_off>"
-          - "(mach | mache) <area> <fan_variants> <turn_off>"
+          - "schalte <luefter> <area> <ab_aus>"
+          - "schalte <area> <luefter> <ab_aus>"
+          - "stoppe <luefter> <area> [<ab_aus>]"
```
kroimon, line 8: I know this was there before, but <ab_aus> really makes no sense after "stoppe", does it?

mib1185: this is already in work - see https://github.com/home-assistant/intents/pull/92#discussion_r1055915978

balloob: It's fine to sometimes catch too much. We also don't always have correct articles. If someone where to say this, and it doesn't make much sense, the intention would still be turning the fan off.

PR #106, opened 2022-12-23
Update Slovenian errors

lukahra:


sentences/sl/_common.yaml
```diff
@@ -1,12 +1,49 @@
...
+      - in: "f"
+        out: "fahrenheit"
+expansion_rules:
+  name: "{name}"
+  area: "{area}"
+  what_is: "(kaj je | kaj je to | čemu je namenjen | kateri)"
+  is_it: "(ali je)"
+  brightness: "{brightness:brightness_pct} [percent]"
+  turn_on: "( vklop | start | začni )"
+  turn_off: "( izklop | stop | končaj )"
```
balloob, line 47: We should only add lists or expansion rules when they are being used in sentences.
```suggestion
lists: {}
expansion_rules: {}
```

PR #111, opened 2022-12-23
[CA] Initial Catalan language support

### LINKED ISSUE #46: Catalan sentences that should be supported.
If you don't feel comfortable contributing to the YAML in this repository, you can post your **Catalan** (`ca`) sentences that match [our supported intents](https://github.com/home-assistant/intents/blob/main/intents.yaml) below.

duhow:
Fixes #46


synesthesiam:
Thank you @duhow! We need at least as many test sentences as there are sentence templates. Are you able to add some more?

duhow:
Sure thing :)

sentences/ca/_common.yaml
```diff
@@ -1,12 +1,57 @@
...
+    no_intent: Ho sento, no entenc això
+    no_area: No hi ha l'àrea {{ area }}
+    no_domain: No hi ha {{ domain }} a {{ area }}
+    no_device_class: No hi ha {{ device_class }} a {{ area }}
+    no_entity: No hi ha cap dispositiu o entitat que es digui {{ entity }}
+    handle_error: He trobat un error inesperat mentre feia l'intent
+lists:
+  color:
+    values:
+      - blanc
```
balloob, line 13: Colors need to be translated to English. See dutch as example

sentences/ca/_common.yaml
```diff
@@ -1,12 +1,57 @@
...
+      - "fahrenheit"
+      - in: "f"
+        out: "fahrenheit"
+
+expansion_rules:
+  name: "[el | la | l'] {name}"
+  area: "[en el | en la | en | del | de la | al | a la | a] {area}"
+  what_is: "(quin és | quina és | que es)"
+  brightness: "{brightness:brightness_pct} [per cent]"
+  turn_on: "(encén | engega | posa)"
```
balloob, line 46: turn_on should be translated etc

duhow: So `expansion_rules` "action" should have the language name action, rather than the English one?
I'm guessing in the future there will be other common names (besides `name` and `area`), but i'm ok to follow that 👍🏻

duhow:
@synesthesiam @balloob all done, please review.
Had some issue while merging #116 but addressed it, if you can please remove the unwanted lang labels.

sentences/ca/_common.yaml
balloob: On my phone but we have this documented in sentences/README.

sentences/ca/_common.yaml
```diff
@@ -1,12 +1,66 @@
...
+        out: red
+      - in: taronja
+        out: orange
+      - in: groc
+        out: yellow
+      - in: verd
+        out: green
+      - in: blau
+        out: blue
+      - in: "(lila|morat|porpre)"
```
balloob, line 27: Lists cannot use matching syntax. Make an entry per value

balloob: So I guess it _can_ use a matching syntax. @synesthesiam do we want that?

duhow:
@synesthesiam I am running `script/test` in local and it passes, despite the warning here in GitHub Actions...
Please advise.

sentences/ca/_common.yaml
duhow: Waiting for your input / suggestion on how to proceed. @synesthesiam

balloob:
Make sure you have the same version of hassil installed. Try re-running script/setup

balloob:
Also try rebasing on latest main branch.

balloob:
The problem is that you have two overlapping matchers:

 - `light_HassTurnOff.yaml` has `(<apaga> | <tanca>) [el | la] llum [<area>]`
 - `homeassistant_HassTurnOff.yaml` has `<apaga> <name>`

The sentence `apaga llum dormitori` can match both.

sentences/ca/_common.yaml
balloob: For now, it's ok. We need to fix it in a few other places too if we decide to ban it.

PR #122, opened 2022-12-24
Make sure HU outputs English color names

balloob:
Color names need to be outputted in English as per #119. This fixes it for HU.


nagyrobi:
Thank you

PR #127, opened 2022-12-24
Add Catalan translations

ArnyminerZ:


balloob:
There is already a Catalan PR open at #111

PR #131, opened 2022-12-25
Split language intents tests by file

balloob:
This splits language intents tests by file.

Since we only allow lists/expansion rules to be defined in `_common.yaml`, we can limit the intent parser to common + the sentences in the file that we're looking at.

~~This however did not result in a speedup.~~

**Before**

```
48 passed in 11.36s

Slowest:
3.99s setup    tests/test_language_intents.py::test_language_sentences[de]
2.22s setup    tests/test_language_intents.py::test_language_sentences[nl]
1.74s setup    tests/test_language_intents.py::test_language_sentences[hu]
```

**After**

```
288 passed in 38.57s

Slowest:
2.25s call     tests/test_language_intents.py::test_cover_HassOpenCover[de]
1.83s call     tests/test_language_intents.py::test_cover_HassCloseCover[de]
1.76s call     tests/test_language_intents.py::test_climate_HassClimateGetTemperature[de]
```

German is slowest as it's currently the most complete language. Leaving it open as draft so I can tinker with it a bit more. But I doubt this is going somewhere. It's a pity because the per-file feedback is a lot more user friendly.

### DIFF
--- a/tests/test_language_intents.py
+++ b/tests/test_language_intents.py
@@ -3,19 +3,11 @@ import dataclasses
 import sys
 from typing import Any, Dict, Iterable, Set
 
-import pytest
 from hassil import Intents
 from hassil.expression import Expression, ListReference, RuleReference, Sequence
 from hassil.intents import TextSlotList
 
 from . import USER_SENTENCES_DIR
-
-
-@pytest.fixture(name="common_language_intents", scope="session")
-def common_language_intents_fixture(language, language_sentences_yaml):
-    """Loads the common language intents."""
-    language_sentences_yaml["_common.yaml"].setdefault("intents", {})
-    return Intents.from_dict(language_sentences_yaml["_common.yaml"])
 
 
 def test_language_common(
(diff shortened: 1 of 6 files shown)


balloob:
Caching the parsing of `_common.yaml` solved everything.

```
288 passed in 11.60s

Slowest:
1.49s setup    tests/test_language_intents.py::test_homeassistant_HassTurnOn[de]
1.00s call     tests/test_language_intents.py::test_light_HassLightSet[hu]
0.75s call     tests/test_language_intents.py::test_cover_HassOpenCover[de]
```

There is now no noticeable speed loss per test.

tests/test_language_intents.py
```diff
@@ -33,10 +45,23 @@ def test_language_common(
...
 ):
     """Ensure all language sentences contain valid slots, lists, rules, etc."""
+    parsed_sentences_without_common = Intents.from_dict(
+        language_sentences_yaml[f"{file_name}.yaml"]
+    )
+
+    # Merge common rules with file specific intents.
+    language_sentences = dataclasses.replace(
+        common_language_intents, intents=parsed_sentences_without_common.intents
+    )
```
balloob, line 62: @synesthesiam instead of relying on `Intents.from_dict` I now manually combine the results of 2. The sentence file only contains `intents` while `_common.yaml` contains the other fields. Is that ok?

PR #138, opened 2022-12-25
[DE] Added expansion_rules & changed sentences/tests to match new expansion rules

Scorpoon:
- Increased possible max value of temperature to 300 (maybe there are some smart ovens out there :D)
- Added expansion_rules for increasing/decreasing and possible lamp names
- Changed sentences/tests to match new expansion rules


sentences/de/_common.yaml
```diff
@@ -54,6 +54,9 @@ expansion_rules:
   schliessen: "(schließ | schließe)"
   machen: "(mach | mache)"
   setzen: "(setze | stelle | ändere)"
+  verringere: "(verringere | reduziere)"
+  erhoehe: "(erhöhe | steigere)"
```
mib1185, line 58: I would suggest to merge `verringere` (_decrease_) and `erhoehe` (_increase_) into `setzen` (_set_).
This would also lower the complexity of possible sentenc definitions, without really loosing any noteworthy  grammatical correctness.

sentences/de/_common.yaml
```diff
@@ -64,5 +67,7 @@ expansion_rules:
   alle_luefter: "alle (Ventilatoren | Lüfter)"
   brightness: "{brightness:brightness_pct} [Prozent]"
   temperature: "{temperature} [Grad] [{temperature_unit}]"
+  lampe: "([(die | der)] Lampe | [(das | des)] Licht[s] | [(den | des)] (Spot[s] | Strahler[s]))"
```
mib1185, line 70:
```suggestion
  lampe: "([(der | die | das | des | den)] (Lampe |  Licht[s] | Spot[s] | Strahler[s]))"
```
again lower the complexity of definition, without really loosing any noteworthy grammatical correctness.

sentences/de/climate_HassClimateGetTemperature.yaml
```diff
@@ -7,4 +7,4 @@ intents:
           - "Wie hoch ist [<area>] die Temperatur"
           - "Wie hoch ist die Temperatur (von | vom) <name>"
           - "Wie hoch ist die Temperatur vom Thermostat <area>"
-          - "Wie warm ist es <area>"
\ No newline at end of file
+          - "Wie warm ist es [<area>]"
```
mib1185, line 10: I thought no the area expansion should not be optional, since just "Wir warm ist es" (_How warm is it?_) seems incomplete without knowing the target area/room

sentences/de/cover_HassOpenCover.yaml
```diff
@@ -3,16 +3,16 @@ intents:
   HassOpenCover:
     data:
       - sentences:
-          - "öffne <area> (<abdeckung> | <tor> | <alle_tore> | <garage> | <alle_garagen>)"
-          - "öffne (<abdeckung> | <tor> | <alle_tore> | <garage> | <alle_garagen>) <area>"
-          - "<machen> <area> (<abdeckung> | <tor> | <alle_tore> | <garage> | <alle_garagen>) auf"
-          - "<machen> (<abdeckung> | <tor> | <alle_tore> | <garage> | <alle_garagen>) <area> auf"
-          - "<machen> <area> (<abdeckung> | <garage> | <alle_garagen>) (rauf | hoch)"
-          - "<machen> (<abdeckung> | <garage> | <alle_garagen>) <area> (rauf | hoch)"
+          - "Öffne <area> (<abdeckung> | <tor> | <alle_tore> | <garage> | <alle_garagen>)"
```
mib1185, line 6: uppercase is not necessary here, because we are not sure if we are at the beginning of a sentence or a `skip_word` was there before

mib1185 (changes requested):

sentences/de/_common.yaml
```diff
@@ -37,7 +37,7 @@ lists:
     range:
       type: "temperature"
       from: 0
-      to: 100
+      to: 300
```
mib1185, line 40: If we are thinking about smart oven, we should also consider smart fridges and decrease the lower boundary, too ^^

sentences/de/_common.yaml
Scorpoon: Good point, i changed that.

sentences/de/climate_HassClimateGetTemperature.yaml
Scorpoon: Just thought of the possibility where the device knows in which room you are. I reverted that for now.

sentences/de/cover_HassOpenCover.yaml
Scorpoon: Good point, i forgot about the skip_words ^^

sentences/de/_common.yaml
Scorpoon: Thaha, thanks for that :D For now i changed it to -30, that should be enough i guess.

sentences/de/light_HassLightSet.yaml
```diff
@@ -8,11 +8,13 @@ intents:
         slots:
           name: "all"
       - sentences:
-          - "<setzen> die Helligkeit [<area>] auf <brightness>"
+          - "<setzen> die Helligkeit [([(von [dem] | vom)] <lampe> | <alle_lampen>)] [<area>] auf <brightness>"
```
mib1185, line 11: We should move `(von [dem] | vom)` into an expansion rule

sentences/de/light_HassTurnOff.yaml
```diff
@@ -3,14 +3,7 @@ intents:
...
-          - "Schalte (das Licht | die Lichter) <area> <ab_aus>"
-          - "Schalte <area> (das Licht | die Lichter) <ab_aus>"
-          - "<machen> (das Licht | die Lichter) <area> <ab_aus>"
-          - "<machen> <area> (das Licht | die Lichter) <ab_aus>"
-
-          - "Schalte alle Lichter [<area>] <ab_aus>"
-          - "Schalte [<area>] alle Lichter <ab_aus>"
-          - "<machen> alle Lichter [<area>] <ab_aus>"
-          - "<machen> [<area>] alle Lichter <ab_aus>"
+          - "(Schalte | <machen>) (<lampe> | <alle_lampen>) [<area>] <ab_aus>"
```
mib1185, line 6: `schalt[e] ` can be added to the `<machen>` expansion rule

sentences/de/climate_HassClimateGetTemperature.yaml
mib1185: This possibility/feature is not there, yet... So we should not count on it 😉

sentences/de/cover_HassCloseCover.yaml
```diff
@@ -5,14 +5,14 @@ intents:
       - sentences:
           - "<schliessen> <area> (<abdeckung> | <tor> | <garage>)"
           - "<schliessen> (<abdeckung> | <tor> | <garage>) <area>"
-          - "<machen> <area> (<abdeckung> | <tor> | <garage>) zu"
-          - "<machen> (<abdeckung> | <tor> | <garage>) <area> zu"
-          - "<machen> <area> (<abdeckung> | <garage>) runter"
-          - "<machen> (<abdeckung> | <garage>) <area> runter"
+          - "<machen> <area> (<abdeckung> | <tor> | <garage>) [<komplett>] zu"
```
mib1185, line 8:
```suggestion
          - "<machen> <area> (<abdeckung> | <tor> | <garage>) [<komplett>] (zu | runter)"
```
Same below 🙂

mib1185: or better make it an expansion rule ^^

sentences/de/_common.yaml
```diff
@@ -36,8 +36,8 @@ lists:
   temperature:
     range:
       type: "temperature"
-      from: 0
-      to: 100
+      from: -30
```
balloob, line 39: @synesthesiam how do you prefer to handle this? If we want to avoid matching nonsense values, we could consider having different temperature ranges that we expect to be called for Climate/Oven/Fridge.

balloob: (we can adjust this later btw, this comment doesn't block this PR)

synesthesiam: We should have different temperature ranges. I'm less concerned about matching, and more about when we generate sentences that don't make sense.

balloob:
@Scorpoon can you wrap up this PR?

sentences/de/_common.yaml
```diff
@@ -64,5 +65,7 @@ expansion_rules:
   alle_luefter: "alle (Ventilatoren | Lüfter)"
   brightness: "{brightness:brightness_pct} [Prozent]"
   temperature: "{temperature} [Grad] [{temperature_unit}]"
+  lampe: "([(der | die | das | des | den)] (Lampe |  Licht[s] | Spot[s] | Strahler[s]))"
```
RosemaryOrchard, line 68: While the above genders/cases are gramatically correct, would it be worth including others to allow for children and non-native speakers who make mistakes?

sentences/de/cover_HassCloseCover.yaml
mib1185: these expansion rule will be available after #492

mib1185:
Hi @Scorpoon 
there are many merge conflicts, because there were many changes merged meanwhile, which also partially covers changes from this PR. Also the cover related sentences has been moved and migrated around. Could you please try to solve these conflicts or start over with a new PR containing just the changes from this PR?

balloob:
Let's start over. This one is too old.

PR #148, opened 2022-12-26
[EL] Added EL tests

dsimop:


tests/el/climate_HassClimateGetTemperature.yaml
```diff
@@ -1,6 +1,26 @@
 language: el
 tests:
-- sentences: []
-  intent:
-    name: HassClimateGetTemperature
-    slots: {}
+  - sentences:
+      - "Wie hoch ist die Temperatur im Wohnzimmer"
```
balloob, line 4: This is German.

balloob:
These tests are good but they need to be accompanied with sentence definitions, or else the test will fail.

PR #158, opened 2022-12-27
[uk]Added fan intents to Ukrainian language

skynetua:


tests/uk/_fixtures.yaml
```diff
@@ -2,7 +2,7 @@ language: uk
 areas:
 - name: Кухн(я|і)
   id: kitchen
-- name: Гостинн(а|ій)
+- name: Вітальн(я|і)
```
balloob, line 5: Names should not use matching syntax. They should be as you would name them in Home Assistant.

balloob: Why would you want to match on both?

skynetua: It is hard to explain in English, I’m not a linguist.

Basically in all Slavic languages we change the ending of nouns and adjectives depending on Cases. As you can see in `pl` situation is the same https://github.com/home-assistant/intents/blob/main/tests/pl/_fixtures.yaml

For example:
I want to name the kitchen area - it is ‘Кухня’
But if I want to reference something, like ‘Light at the kitchen’ - it is ‘Light на Кухні’

or: 
Lamp - Лампа
Turn on Lamp - Ввімкни Лампу

We change the last 1-2 letters in such cases or add vowels if the word ends with consonants.

I added it as is for now to not fail tests but there should be more complex solution.

balloob: Your name in Home Assistant is static. It does not contain sentence syntax. It should never have been allowed here. This needs to be solved differently. 

Cc @synesthesiam

nagyrobi: Hungarian similar:
Kitchen - Konyha
Lignt in the kitchen - Lámpa a konyhában
Turn on the light in the kitchen - Kapcsold fel a lámpát a konyhában

nagyrobi: And they also depend on other factors...

balloob: Check. Let's track it in #164. For now we're going to approve this PR as it's not new for UK.

PR #164, opened 2022-12-28
Validate no matching syntax in fixture names

balloob:
This updates the validator to catch matching syntax being used in fixture names.

Although our parser allows fixtures with parsing syntax, it's not something Home Assistant would ever provide.

We need a solution where we can define a format how nouns are transformed.

See [files changed](https://github.com/home-assistant/intents/pull/164/files) for languages that fail validation.

### DIFF
--- a/script/intentfest/validate.py
+++ b/script/intentfest/validate.py
@@ -252,6 +252,7 @@ TESTS_SCHEMA = vol.Schema(
     }
 )
 
+NAME_MATCHER = vol.Match(r"^[\w ]+$", "Names cannot contain matching syntax")
 TESTS_FIXTURES = vol.Schema(
     {
         vol.Required("language"): str,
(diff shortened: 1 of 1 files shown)


HepoH3:
Russian have similar grammar and will fail validation, bacause I used the same hack as mentioned in https://github.com/home-assistant/intents/pull/158#discussion_r1057781441
Additionaly I had to use another hack with color cases:
https://github.com/home-assistant/intents/blob/78613430c7dda465d84bcdd31fdeed0aef6905e6/sentences/ru/_common.yaml#L13
The problem is that it's not only Zone naming problem, but every entity in the house. If it was just zones, we could ask user to specify all word cases for zone, but it's unreallistic to specify all entities cases.
I believe the problem need to be solve with something like [this](http://snowball.tartarus.org/algorithms/russian/stemmer.html). This algorithm will get base part of the word (root of the word).
Moreover those page raise another russian language feature: usage of the literal `ё`. This letter is rarely used and should be mapped to `е`. It's not ambiguous, and will simplify patterns like this:
https://github.com/home-assistant/intents/blob/78613430c7dda465d84bcdd31fdeed0aef6905e6/sentences/ru/_common.yaml#L15

balloob:
(rebased to get a new list of languages that require a solution)

PR #172, opened 2022-12-28
Fix command to add new language

HepoH3:
When trying to run command as in example:

```
python3 -m script.intentfest add-language <lang_code>
```
following error appeared:

> error: argument action: invalid choice: 'add-language' (choose from 'add_language', 'validate', 'parse', 'sample', 'sample_template')


balloob (approved):
Thanks.

PR #175, opened 2022-12-28
[RU] Add russian language support

HepoH3:
It was quite hard to start (espesially because there is six noun cases in russian) and there is no example in english how to handle them. But in the end, I made this to work, which makes me proud :).
Hope this will help other people who have similar language grammar.
As I can see, there is no leader for Russian language, I can fill this vacancy 'till someone who want this appears.

PR #177, opened 2022-12-28
Add Russian language leader

HepoH3:
See #175.


synesthesiam:
Great, welcome aboard!

Issue #184, opened 2022-12-29
add female version of hebrew sentences


regevbr:
Who says that the voice assistant has to be male :-) ?

thecode:
For now he recommendation is to make natural form sentences, there is a discussion about it in https://github.com/home-assistant/intents/issues/33, I think it is not Hebrew specific problem

regevbr:
ok

PR #185, opened 2022-12-29
[NL] Sentences to toggle all entities of domain

TheFes:
Added sentences to toggle all entities of a domain (currently light and fan). 
Can be useful if you leave the house, and want to turn of all lights/fans.

I added a "domain_only" slot combination to `intents.yaml` to support this.

I used matching syntax in the lists, as was done in https://github.com/home-assistant/intents/pull/111 If this is banned later on, I will update it :)

### DIFF
--- a/sentences/nl/homeassistant_HassTurnOff.yaml
+++ b/sentences/nl/homeassistant_HassTurnOff.yaml
@@ -4,3 +4,5 @@ intents:
     data:
       - sentences:
           - "<doe> <name> uit"
+      - sentences:
+          - "<doe> (overal | alle) [de | het] {domain} uit"
--- a/intents.yaml
+++ b/intents.yaml
@@ -23,6 +23,8 @@ HassTurnOn:
       - "name"
     area_only:
       - "area"
+    domain_only:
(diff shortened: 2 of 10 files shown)


TheFes:
Hmm, something went wrong here, as this also includes ass changes from #184 
Will remove these changes from this PR

TheFes:
I see that for German sentences to toggle all domain entities have been added to the domain specific sentence files. If this is the preferred approach let me know :)

sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -4,3 +4,5 @@ intents:
     data:
       - sentences:
           - "<doe> <name> uit"
+      - sentences:
```
balloob, line 7: Ipv domain matcher, voeg een zin toe aan light_HassTurnOff met domain also hardcoded value

TheFes:
I'll close this one, and wait until #189 is merged so I can use the `lamp` expansion rule as suggested in that PR
I'll add the sentences to the domain specific files then

PR #194, opened 2022-12-29
[RO] Added initial intents

gabimarchidan:
Initial work, based on what's in the folder and the es one.


gabimarchidan:
I am unable to sign the CLA. 
I've tried in both Firefox and Chrome, from different machines.
The signing seems broken.

https://prnt.sc/Y9Gl-vL3xAk0

PR #197, opened 2022-12-29
Add empty Bulgarian (bg) templates

synesthesiam:


languages.yaml
```diff
@@ -1,28 +1,30 @@
 ar:
-  nativeName: العربية
+  nativeName: "\u0627\u0644\u0639\u0631\u0628\u064A\u0629"
```
balloob, line 2: Why change this?

balloob: Or did this happen because of the `add_language` output ?

balloob (approved):

PR #220, opened 2023-01-02
Update light_HassTurnOn.yaml

jirimissbach:
adding plural, and singular variants as well as the formal vs spoken language

### DIFF
--- a/sentences/cs/light_HassTurnOn.yaml
+++ b/sentences/cs/light_HassTurnOn.yaml
@@ -2,7 +2,9 @@ language: cs
 intents:
   HassTurnOn:
     data:
-    - sentences: []
+    - sentences:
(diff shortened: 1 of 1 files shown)


balloob:
When adding sentences, you also need to update the corresponding test files. They live in the `tests` folder.

jirimissbach:
Hi Paulus,
 I did update the test files; however, when I saw the failure, I realized that a sentence was missing (at least based on the error  - tests/cs/light_HassTurnOn.yaml: not all sentences have tests). 

I have tried to correct that now.  Sorry it's my 1st time on Github, but I'll do my best to be useful!

TomaszPilch:
@jirimissbach I think you need create new PR from branch because this can't be merged, the tests will never pass without the test files. You should close these PRs.

jirimissbach:
closing due to errors

PR #223, opened 2023-01-02
Update light_HassTurnOff.yaml

jirimissbach:


balloob:
Please note that you also need to add a matching sentence to match the test sentences sentence.

halecivo:
Ahoj @jirimissbach, 

Thanks for starting with the translation!

In this PR, only test for light_HassTurnOff is being changed. Would it be possible to name PRs in more descriptive way, so it is easier for others to know, what is already being worked on?

Also just maybe having a single PR for both test and sentence, having clean commit history (not having first commit with wrong test, then updated with the correction), would make changes more readable.

I don't want to be pedantic or to argue here, just wanted to propose way for easier colaboration at least for next czech PRs.

Thanks for considering. :)

jirimissbach:
Hi @halecivo ,
Sorry - I know I have messed that up. Newbie mistakes :-(

I'll definitively do better, and thanks for the advice. My main motivation is to help HA development

jirimissbach:
btw @halecivo    -is it better if I close the messed-up PRs?

halecivo:
@jirimissbach no worries, I am just same guy as you and what I wrote may not be aligned with NabuCasa.

Personally I would keep just one PR for the domain (lights) and do tests and sentences in one of the PR and decline the others. We also don't need to be super strict here and can improve mostly for later PRs. 

I think there is also missing definition for expansion macros like <turn_on>, I tried to fix this in PR #237 , but @balloob seems to prefer to not have expansion rules until there are more possibilities, so maybe we should directly write zapni instead of <turn_on> in sentences for now, but let's wait for his feedback.

halecivo:
Hello @jirimissbach.

So after review of my PR, I learned that:
 - commits are squashed here before merge, therefore commit history seems to not matter much, although personally I think that having clean history before first review can help reviewers if PR. Incorporating comments from reviewers into separate commits also helps to check the changes since last review.
 - expansion rules, which are not slots, should really be in native language, therefore <zapnout> should be used in czech sentences instead of <turn>. You can check _common.yaml to see them.
 - You may need to update both test and sentences in one PR in order to have test passing.

In order to be able to use czech expansion rules from my PR, you need to update on latest main branch. Do you know how to do that? I don't know if there is a button in web github, but otherwise Github Desktop can be used for advanced work with git.

Do you need any help or are you able to finalize this PR?

jirimissbach:
@halecivo, thanks for the detailed explanation. I will close this PR; probably better to have a clean start. I'm happy to help  with any sort of HA and would appreciate your help/guidance how to do it effectively. 


Cheers,

jirimissbach:
closing due to errors - need to submit differently

PR #229, opened 2023-01-02

joselito11:


balloob:
New matching sentences also need to include tests in the corresponding `tests` file.

frenck:

../Frenck

PR #232, opened 2023-01-03

khymmera:
Added some alternative commands


balloob:
Do not forget to add tests for the newly added sentences.

PR #233, opened 2023-01-03
[ES] Update tests for homeassistant_HassTurnOn.yaml

khymmera:
Added some alternative commands


tests/es/homeassistant_HassTurnOn.yaml
```diff
@@ -7,3 +7,15 @@ tests:
       name: "HassTurnOn"
       slots:
         name: "fan.ventilador_techo"
+  - sentences:
+      - "abre la cortina izquierda"
```
balloob, line 11: Cover specific sentences should be added to `cover_HassTurnOn.yaml`

khymmera: Removed sentence.

tests/es/homeassistant_HassTurnOn.yaml
```diff
@@ -7,3 +7,9 @@ tests:
       name: "HassTurnOn"
       slots:
         name: "fan.ventilador_techo"
+  - sentences:
```
cibernox, line 10: While it's not incorrect,do we ever refer to the switch when turning on/off something? I always ask to turn on/off the light (or whatever) controlled by the switch, but never the switch itself. Perhaps there's an use case I'm not aware of.

davefx: Sometimes you could have a generic switch in a room, where you connect several devices. 
In that case, you would use this sentence.

davefx:
In order to merge the PR, tests must pass

PR #235, opened 2023-01-03
[NL] Sentences to toggle all

TheFes:
I tried to created sentences to toggle all entities of a domain. `turnOff` works fine, but I ran into issues with the `turnOn` sentences because of the `slot_combinations`

Inspired by the [German sentences](https://github.com/home-assistant/intents/blob/main/sentences/de/light_HassTurnOn.yaml) I added an optional `<name>` selector, which I did not use in the tests.

This works (at least the tests are successful, but this looks like a hack to me, and not a real solution.
When `[<name>]` is removed, it gives the following error:
`FAILED tests/test_language_intents.py::test_light_HassTurnOn[nl] - AssertionError: Slot combination not matched: intent='HassTurnOn', slots=set(), sentence='<doe> [overal | alle] <lamp> [<naar>] aan'`

Decided to create the PR anyway so others can take a look and maybe advice what would be the best way to implement this.

Tried another (in my view less hacky) way in PR #185 but that required a `domain_only` option as `slot_combination` and translations per domain in `_common.yaml`

Also made some changes to the existing sentences, and added a separate sentence for `"Schakel [...] in"`
I could have made an expansion rule with `(aan | in)` but that would allow sentences like `Zet de lamp in` so I thought it would be better to create a separate sentence. I did leave `schakel` in the existing expansion rule. Although `"Schakel de lamp aan"` might be technically incorrect, it does seem like something people might say.


TheFes:
Hi @synesthesiam, I already asked balloob to have a look, but I would like to ask if you also maybe have some advice on this?
See the first comment for more info.

balloob:
The reason we don't allow it is because we do not want to allow turning on all lights. It's a dangerous sentence to match and the odds that someone ever wants to turn on everything is very low.

TheFes:
Clear, I'll remove them from `on` then and only leave `off`

Might be good to have a look at the German sentences then.

TheFes:
@balloob 
I tried to add this sentence:
`<doe> [(alle | in)] <area> <lamp> uit`

This should allow something like `Zet de keuken lampen uit`

This now seems to conflict with `Mag de slaapkamerlamp uit?` from the generic `HassTurnOff` intents.
I expect the same applies to the `on` sentences as well (didn't show in the checks though).

I can of course remove the sentence from de domain specific files to resolve the error, but it does seem like something people will say.
However I don't really see a solution here.
- It seems the spaces are ignored, als slaapkamerlamp is one word, but it is still mapped to `<area> <lamp>`. But even if they would not be ignored this will be an issue, as people might just name the lamp `slaapkamer lamp` (not everybody removes spaces in combined words as they should)
- `slaapkamerlamp` is singular, so I could make a separate plural extension rule for the words for lights and use that in the sentence, but that will not solve it, as people will use light groups, like `keukenlampen` which will be a single target.

For now I made `(alle | in)` mandatory instead of optional. That resolves the (potential) conflicts, but it doesn't allow `Zet de keuken verlichting uit`, that needs to be `Zet alle keuken lampen uit` or `Zet in de keuken de verlichting uit` then.

For now I would say this PR is ready to be merged.

balloob (approved):

PR #236, opened 2023-01-03
[EL] Greek initial support

dsimop:


dsimop:
The failing tests occur only because the recognition model cannot handle a specific case for <domains>, <names> etc. - which happens always often in Greek for words whose gender is male. Let <word1 word2> be a <name> and <word3> a <domain>. In any case the logic is the same and the recognition model has to handle these _<entity_names>_ as follows:

If there is **any** word in a <name>/<domain> ending with "ς", that character has to be omitted/skipped when reading a sentence/test. For instance: name: "word1 word2",   _"word1" ends with "ς"_.  But the sentence/test must include that <name> _without_ the aforementioned character "ς".

glogiotatidis (approved):
nice work! :tada: 

Translations are good. The failing tests are surfacing a problem in the parsing engine imho. We will need to have this flexibility with Greek (maybe other langs as well).

Here's a related issue:

An entity name is `ανεμιστήρα οροφής` (fan ceiling)
When we say a full sentence to turn it off we would add the word "the" between "fan" and "ceiling" and remove the trailing `ς` from the word before. I would become 

`κλείσε τον ανεμιστήρα **της** οροφής`

Another example for `φωτιστικό υπνοδωματίου` (lamp bedroom)

`κλείσε το φωτιστικό **του** υπνοδωματίου`

here there's no trailing `ς` and thus none removed.

Thoughts on how to handle those scenarios?

PR #237, opened 2023-01-03
Add turn_on/off(_light) expansion rules for czech

halecivo:
This seems to be needed for PRs #220 and #221 .

### DIFF
--- a/sentences/cs/_common.yaml
+++ b/sentences/cs/_common.yaml
@@ -9,11 +9,11 @@ responses:
     handle_error: "Při zpracování došlo k neočekávané chybě."
 lists: {}
 expansion_rules:
-  name: "{name}"
-  area: "{area}"
-  turn_on: "zapni"
-  turn_off: "vypni"
-  turn_on_light: "rozsviť"
-  turn_off_light: "zhasni"
+  nazev: "{name}"
+  oblast: "{area}"
+  zapnout: "(zapni|zapnout)"
+  vypnout: "(vypni|vypnout)"
+  rozsvitit: "(rozsviť|rozsvítit)"
+  zhasnout: "(zhasni|zhasnout)"
 skip_words:
   - "prosím"
--- a/sentences/cs/_common.yaml
+++ b/sentences/cs/_common.yaml
@@ -9,8 +9,6 @@ responses:
     handle_error: "Při zpracování došlo k neočekávané chybě."
 lists: {}
 expansion_rules:
-  nazev: "{name}"
-  oblast: "{area}"
   zapnout: "(zapni|zapnout)"
   vypnout: "(vypni|vypnout)"
   rozsvitit: "(rozsviť|rozsvítit)"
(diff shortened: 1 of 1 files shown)


sentences/cs/_common.yaml
```diff
@@ -8,5 +8,12 @@ responses:
...
 lists: {}
-expansion_rules: {}
-skip_words: []
+expansion_rules:
+  name: "{name}"
+  area: "{area}"
+  turn_on: "zapni"
+  turn_off: "vypni"
+  turn_on_light: "rozsviť"
+  turn_off_light: "zhasni"
```
balloob, line 17: expansion rules that do not wrap a slot should be in the native language. Also, expansion rules should expand something. These are just 1:1 mappings to words.

halecivo: If you prefer, we can decline this change, just here is my view:

1. If new variant is added, I find it good to have expansion rules already prepared for easier extension. Other region dialects may use variants, which is not common for me (and not part of official language), but they can easily be added if there is a need. 
3. Other languages (like sk) already merged usage of expansion rules in the same way (of course having it wrong elsewhere does not mean it needs to be wrong here as well).
4. Those rules are used in PRs #220 and #221, but not defined, so I wanted to prepare the missing part. I could also rework Jiri's PR if preferred.

balloob: They still will need to be translated in Czech.

halecivo: Ah, you mean that if we use expansion rule, which is not a slot, we should name the rule itself in native language - like using <zapni> instead of <turn_on> in sentences?

sentences/cs/_common.yaml
```diff
@@ -9,11 +9,11 @@ responses:
     handle_error: "Při zpracování došlo k neočekávané chybě."
 lists: {}
 expansion_rules:
-  name: "{name}"
-  area: "{area}"
-  turn_on: "zapni"
-  turn_off: "vypni"
-  turn_on_light: "rozsviť"
-  turn_off_light: "zhasni"
+  nazev: "{name}"
```
halecivo, line 12: @balloob did I get it right?

halecivo:
- [ ] squash fixups before merge

sentences/cs/_common.yaml
balloob: Correct.

sentences/cs/_common.yaml
balloob: Yep!

balloob: well no. The ones that wrap a slot should not be translated.

sentences/cs/_common.yaml
```diff
@@ -8,5 +8,12 @@ responses:
     no_entity: "Žádné zařízení ani entita se nejmenuje {{ entity }}"
     handle_error: "Při zpracování došlo k neočekávané chybě."
 lists: {}
-expansion_rules: {}
-skip_words: []
+expansion_rules:
+  nazev: "{name}"
+  oblast: "{area}"
```
balloob, line 13:
```suggestion
  name: "{name}"
  area: "{area}"
```

PR #239, opened 2023-01-04
[FA] Initial Persian Support

AwesomeGuy000:
Initial support for Persian


sentences/fa/_common.yaml
```diff
@@ -0,0 +1,92 @@
...
+    no_entity: "وجود ندارد {{ entity }} هیچ دستگاه یا موجودی به نام"
+    handle_error: "هنگام اجرای هدف یک خطای غیر منتظره بوجود آمد "
+lists:
+  color:
+    values:
+      # white
+      - in: "سفید"
+        out: "white"
+      - in: "white"
+        out: "white"
```
balloob, line 17: Why are you supporting English too?

Also, if values are the same, you can drop the `in`, `out` keys
```suggestion
      - "white"
```

AwesomeGuy000: Changed.

tests/fa/homeassistant_HassTurnOff.yaml
```diff
@@ -0,0 +1,9 @@
+language: fa
+tests:
+  - sentences:
+      - "bedroom lamp را خاموش کن"
```
balloob, line 4: For a follow-up PR, let's also convert the entity names to Farsi.

balloob (approved):

PR #241, opened 2023-01-04
[LT] Added Lithuanian language

ErnestStaug:


sentences/lt/_common.yaml
```diff
@@ -0,0 +1,19 @@
...
+    no_domain: "{{ area }} neturi {{ domain }}"
+    no_device_class: "{{ area }} neturi {{ device_class }}"
+    no_entity: "Nėra įrenginio ar daikto pavadinimu {{ entity }}"
+    handle_error: "Atsirado nenumatyta klaida bandant įvykdyti užklausą"
+lists: {}
+expansion_rules:
+  name: "{name}"
+  area: "{area}"
+  turn_on: "(įjunk | paleisk | pradėk)"
+  turn_off: "(išjunk | stabdyk | sustabdyk)"
```
balloob, line 15: Expansion rules should be in the native language unless it is wrapping a slot name, in which case it should match the slot name.

ErnestStaug: I look at other language examples and very few of them use native language for the names. I would argue it's better to keep the keys of variables (expansion keys) in english as I am not sure yaml supports keys with special characters (accents/non latin letters/symbols/logograms). As well it's easier to debug for someone not native in that language.

PR #246, opened 2023-01-04
[SV] Add improvements to Swedish language

larsdunemark:
Trying to make the error codes feel more human in the Swedish language.
Adding more variants for climate temperature get controls.
Adding light turn on and turn off commands and include some extra courtesy word "tack" to be ignored


sentences/sv/_common.yaml
```diff
@@ -50,11 +50,16 @@ expansion_rules:
...
-  is_it: "(är det)"
+  is_it: "(är det | är)"
   brightness: "{brightness} [percent]"
   turn_on: "( tänd | sätt på | slå på | starta )"
   turn_off: "( släck | stäng av | stoppa )"
+  in_on: "(i | på)"
   temp: "(temperatur)"
+  all: "(all | alla )"
+  lightning: "(ljuset | belysning[en] | lampan | lampor[na])"
+  hot_cold: "(varmt | varm | kallt | kall | het | svalt )"
```
balloob, line 61: expansion rules should be in Swedish, not English. Unless they wrap a slot name like `{name}`

larsdunemark: ok fixed in additional commit

lellky (approved):
I think these changes look good!

PR #248, opened 2023-01-04
Adding codeowner for Greek

drthanwho:
I know @Chreece from the Greek HA server and he's quite experienced with Rhasppy already and asked to be leader. I know he'll do a good job 😄 

Saving some time by making the PR myself.


balloob:
We first want to see language leaders do a few contributions before we can accept them, as we also need to make sure they understand doing contributions with GitHub. So we'll open this PR again when it's time.

PR #253, opened 2023-01-04
Adding more words to HassOpenCover and HassCloseCover in Polish

Wojciechgc:


synesthesiam:
@Uriziel01 can you have a look at the conflicts here please?

Wojciechgc:
@synesthesiam I added @Uriziel01's changes.

Uriziel01:
Looks good to me.
@Wojciechgc  We should also fix the `- "(Zamknij | Zasłoń | Opuść) <area> [wszystkie] rolety"` as the <area> part does not make sense in that form when in Polish (_Opuść kuchnia rolety?_).
But this will be done separately and is outside the scope of this PR.

Wojciechgc:
> Looks good to me.
> @Wojciechgc  We should also fix the `- "(Zamknij | Zasłoń | Opuść) <area> [wszystkie] rolety"` as the `<area>` part does not make sense in that form (_Opuść kuchnia rolety?_).
> But this will be done separately and is outside the scope of this PR.

I think that the person who added this wanted sentences like _Opuść w kuchni rolety_ to work (the system would have to understand that _kuchni_ means _kuchnia_, _salonie_ means _salon_ etc.)

Uriziel01:
If you check https://github.com/home-assistant/intents/blob/main/sentences/en/cover_HassCloseCover.yaml you'll see that it requires additional square brackets for `[the]` prefix (in our case `[w]`) so I belive it will not work as you've described it currently, but we can adjust it easily.
Regarding the semantics and general word order, it is a valid sentence in Polish but it's quite the uncommon one. Although in other PR authors emphasised that they are putting focus in flexibility for now so we probably can leave it in that form.

PR #260, opened 2023-01-04
[PL] Update cover opening sentences

Uriziel01:
`Rozsłoń` is actually an archaic form and is not used in Polish anymore. Replaced it with `Odsłoń`.

### DIFF
--- a/sentences/pl/cover_HassOpenCover.yaml
+++ b/sentences/pl/cover_HassOpenCover.yaml
@@ -11,9 +11,9 @@ intents:
         slots:
           device_class: "garage_door"
       - sentences:
-          - "Rolety <area> (otwórz | rozsłoń)"
-          - "(Otwórz | Rozsłoń) rolety <area>"
-          - "(Otwórz | Rozsłoń) <area> [wszystkie] rolety"
+          - "Rolety <area> (otwórz | odsłoń)"
+          - "(Otwórz | Odsłoń) rolety <area>"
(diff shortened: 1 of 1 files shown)


sentences/pl/cover_HassOpenCover.yaml
```diff
@@ -11,9 +11,9 @@ intents:
         slots:
           device_class: "garage_door"
       - sentences:
-          - "Rolety <area> (otwórz | rozsłoń)"
-          - "(Otwórz | Rozsłoń) rolety <area>"
-          - "(Otwórz | Rozsłoń) <area> [wszystkie] rolety"
+          - "Rolety <area> (otwórz | odsłoń)"
```
balloob, line 14: archaic forms can still be used, so let's keep matching on it.

Uriziel01: @balloob No problem I can just add the modern form to the list and leave it as is.
Just to make sure - I was not able to find a modern dictionary that references that word, all examples I was able to find are dated back to XVI - XIX centuries.
What's our take for such cases in the future? We want to leave as much flexibility as possible or rather limit the words so we make the voice recognition a bit easier?

synesthesiam: Flexibility is preferred here since we are just matching text. In the future, when generate training sentences we will need to be more precise.

PR #263, opened 2023-01-04

vladacr:
Přidány základní věty pro jas a barvu

### DIFF
--- a/sentences/cs/light_HassLightSet.yaml
+++ b/sentences/cs/light_HassLightSet.yaml
@@ -2,4 +2,23 @@ language: cs
 intents:
   HassLightSet:
     data:
-    - sentences: []
+      #jas | brightness
+      - sentences:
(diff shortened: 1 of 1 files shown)


sentences/cs/light_HassLightSet.yaml
```diff
@@ -2,4 +2,23 @@ language: cs
 intents:
   HassLightSet:
     data:
-    - sentences: []
+      #jas | brightness
```
halecivo, line 5: Maybe I'm having my OCD moment here, but would be happier if there was a space between # and jas in the same way as you did for # barva | color.

halecivo (changes requested):

TomaszPilch:
You are missing tests in `tests` folder

halecivo:
closing due to inactivity - work already done in different PR

PR #265, opened 2023-01-04

OzGav:
Added [the] for grammatical correctness

### DIFF
--- a/sentences/en/homeassistant_HassTurnOn.yaml
+++ b/sentences/en/homeassistant_HassTurnOn.yaml
@@ -3,5 +3,5 @@ intents:
   HassTurnOn:
     data:
       - sentences:
-          - "turn on <name>"
-          - "turn <name> on"
+          - "turn on [the] <name>"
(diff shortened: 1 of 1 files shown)


sentences/en/homeassistant_HassTurnOn.yaml
```diff
@@ -3,5 +3,5 @@ intents:
   HassTurnOn:
     data:
       - sentences:
-          - "turn on <name>"
-          - "turn <name> on"
+          - "turn on [the] <name>"
+          - "turn [the] <name> on"
```
balloob, line 7: `[the]` is already part of [the `<name>` expansion rule](https://github.com/home-assistant/intents/blob/main/sentences/en/_common.yaml#L41-L42)

OzGav: I have a look at some of the other intents then as they had [the] so I thought it was missing here? Maybe the rule didnt apply in that case

PR #267, opened 2023-01-05
Adding Estonian language

LauriLepik:
Working on adding Estonian language support


languages.yaml
```diff
@@ -23,7 +23,10 @@ el:
 en:
   nativeName: English
 es:
+  nativeName: "Espa\xF1ol"
```
balloob, line 26:
```suggestion
```

sentences/et/_common.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - in: "c"
+        out: "celsius"
+      - "fahrenheit"
+      - in: "f"
+        out: "fahrenheit"
+expansion_rules:
+  name: "{name}"
+  area: "{area}"
+  what_is: "(mis | mis on)"
+  how_many: "(mitu)"
```
balloob, line 53: Expansion rules should be in Estonian unless they wrap a slot (ie `{name}`)

tests/et/light_HassTurnOn.yaml
```diff
@@ -0,0 +1,16 @@
...
+      - "lülita sisse kõik toa elutuba tuled"
+      - "lülita sisse toa elutuba tuli"
+      - "lülita sisse toa elutuba tuled"
+      - "lülita kõik tuled toas elutuba sisse"
+      - "lülita tuled toas elutuba sisse"
+      - "lülita tuli toas elutuba sisse"
+    intent:
+      name: "HassTurnOn"
+      slots:
+        area: elutuba
```
balloob, line 16: These tests are for `homeassistant_HassTurnOn`. Tests in the `light_HassTurnOn`. domain should trigger the sentences in `light_HassTurnOn`, which contain the `light` domain.

balloob:
We've changed a bunch of stuff around in the last 2 days as we introduced responses to the repository. This means that this PR is no longer up to date and it's missing responses for all intents + it's format changed slightly. I would suggest to start with a clean `add_language` command and copy your files on top of it.

PR #251, opened 2023-01-04
Update _common.yaml

iddiek:
qual è si scrive senza apostrofo

PR #275, opened 2023-01-05
Revert "Update _common.yaml"

balloob:
Reverts home-assistant/intents#251 as it breaks CI.


xraver:
I have to fix the merge. I'll do in two days.
I'm out with limited internet access

auanasgheps:
@xraver I can do it

PR #276, opened 2023-01-05

rauffinger:
fixed variable names <name> , <brightness> and <area> (which should remain in English). Added some more verb variations.


joaorgoncalves:
Closing this request as it won't pass the test due to variable conflict. 
Although it is a good practice to keep variables in English, these specific variables came from the "expansion_rules" and might be in the translators' language to simplify sentence construction. 
I'll make a new PR to add the suggested sentence additions.

PR #309, opened 2023-01-05
schliessen: "(schließ | schließe | zumachen)"

stijink:


andreasbrett:
There's no german sentence definition yet that uses <schliessen> at the end of a sentence. So "zumachen" doesn't really make sense (for now).

- "<schliessen> <area> (<abdeckung> \| <tor> \| <garage>)"
 - "<schliessen> (<abdeckung> \| <tor> \| <garage>) <area>"

"Zumachen im Hinterhof das Tor" is just weird and incorrect. "Im Hinterhof das Garagentor zumachen" makes much more sense. So I suggest to only introducing "zumachen" into _common.yaml when there's also a sentence where this makes sense. Furthermore there's no added test for this addition.

mib1185 (changes requested):
this is already covered by sentences like

https://github.com/home-assistant/intents/blob/2b3df8a84198b297b450be8b73431e91c3e732e7/sentences/de/cover_HassCloseCover.yaml#L8-L9

https://github.com/home-assistant/intents/blob/2b3df8a84198b297b450be8b73431e91c3e732e7/sentences/de/cover_HassCloseCover.yaml#L15-L16

https://github.com/home-assistant/intents/blob/2b3df8a84198b297b450be8b73431e91c3e732e7/sentences/de/cover_HassOpenCover.yaml#L8-L9

https://github.com/home-assistant/intents/blob/2b3df8a84198b297b450be8b73431e91c3e732e7/sentences/de/cover_HassOpenCover.yaml#L15-L16

mib1185:
> There's no german sentence definition yet that uses `<schliessen>` at the end of a sentence. So "zumachen" doesn't really make sense (for now).
> 
> * `<schliessen> <area> (<abdeckung> \| <tor> \| <garage>)`
> * `<schliessen> (<abdeckung> \| <tor> \| <garage>) <area>`
> 
> "Zumachen im Hinterhof das Tor" is just weird and incorrect. "Im Hinterhof das Garagentor zumachen" makes much more sense. So I suggest to only introducing "zumachen" into _common.yaml when there's also a sentence where this makes sense. Furthermore there's no added test for this addition.

I don'r think adding "zumachen" as own expansion rule would be sufficient - think about "auf machen", "hoch machen" ...

in this case i would recommend to combine the expansion rules in the sentence like:
_`auf` and `zu` will be available with #492_

```
<area> <name> <auf> <machen>
<area> <name> <zu> <machen>
```

RosemaryOrchard (approved):

PR #331, opened 2023-01-05
Update french

baptistebisson:


acesyde:
I think you need to speak with @slegars56 and @Gorouk you are working on the same files both

flexy2dd:
Outaded PR

PR #352, opened 2023-01-05
Update _fixtures.yaml

thefjordkeeper:
Add new rooms.


balloob:
Changes like these should be combined with test sentences that are using them.

PR #354, opened 2023-01-05
Validate non domain files

balloob:
Make sure that files like `homeassistant_HassTurnOff.yaml` do not test for sentences defined in `light_HassTurnOff.yaml`.


sentences/fr/homeassistant_HassTurnOff.yaml
```diff
@@ -2,16 +2,6 @@ language: fr
 intents:
   HassTurnOff:
     data:
-      - sentences:
```
balloob, line 5: Removed these as they are duplicate from the light_HassTurnOff

PR #355, opened 2023-01-05
Update _common.yaml

thefjordkeeper:
Add color temp in kelvins


ValShaped:
I suggest allowing input color temperatures between 2000 and 10000 Kelvin. Color temperatures for light.turn_on are bound to the range the target entity supports, and some devices (say, a computer running a Home Assistant-enabled color-temperature filter) may accept very high values. As far as I know the default range for lights in Home Assistant is 150-500 mired (~6500-2000 K)

I personally have lights that claim to support 2700-6500K.

PR #390, opened 2023-01-05
Slot values no dict

balloob:
We allowed both a value or a dictionary containing a `value` key for expected slot values. In practice, we don't support any other keys but `value` so we could just always use the shortcut.

This updates our validation to enforce shortcuts and converts all test files to shortcuts.

### DIFF
--- a/script/intentfest/validate.py
+++ b/script/intentfest/validate.py
@@ -166,10 +166,10 @@ TESTS_SCHEMA = vol.Schema(
                 vol.Required("intent"): {
                     vol.Required("name"): str,
                     vol.Optional("slots"): {
-                        str: vol.Any(
-                            {vol.Required("value"): match_anything},
-                            match_anything_but_dict,
-                        ),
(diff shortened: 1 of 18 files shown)


script/intentfest/validate.py
```diff
@@ -166,10 +166,10 @@ def validate(value):
                 vol.Required("intent"): {
                     vol.Required("name"): str,
                     vol.Optional("slots"): {
-                        str: vol.Any(
-                            {vol.Required("value"): match_anything},
-                            match_anything_but_dict,
-                        ),
+                        # In the future, if we want to allowa dictionary,
```
HepoH3, line 169: allowa -- missed space?

HepoH3 (commented):
I have not enough expertise to judge changes in python files, but changes in yaml files seems ok

script/intentfest/validate.py
balloob:
```suggestion
                        # In the future, if we want to allow a dictionary,
```

PR #397, opened 2023-01-05
[LV] Add Latvian language

makstech:


sentences/lv/_common.yaml
```diff
@@ -0,0 +1,20 @@
...
+    no_domain: "Zonā {{ area }} nav {{ domain }}"
+    no_device_class: "Zonā {{ area }} nav {{ device_class }}"
+    no_entity: "Nav ierīces vai entītijas ar nosaukumu {{ entity }}"
+    handle_error: "Apstrādājot nolūku, radās neparedzēta kļūda"
+lists: {}
+expansion_rules:
+  name: "{name}"
+  area: "{area}"
+  turn_on: "(palaid | ieslēgt | ieslēdz | ieslēdziet)"
+  turn_off: "(atslēgt | atslēdz | izslēgt | izslēdz | izslēdziet)"
```
balloob, line 15: Please translate the keys `turn_on` and `turn_off` to Latvian.

balloob:
Thanks! 🐬 

Please translate the expansion rule keys in a follow-up pull request

makstech:
@balloob Sure, no problem, though I must say, I agree with [this comment](https://github.com/home-assistant/intents/pull/241#discussion_r1062237266)
> I look at other language examples and very few of them use native language for the names. I would argue it's better to keep the keys of variables (expansion keys) in english as I am not sure yaml supports keys with special characters (accents/non latin letters/symbols/logograms). As well it's easier to debug for someone not native in that language.

If that won't raise any issues in YAML processing, then sure. But having English as the primary language for keys would be a bit more streamlined IMO.

PR #404, opened 2023-01-05
Add Danish cover sentences

fadamsen:


fadamsen:
I'm not sure about the sentences for shutters, as I've never had shutters. But we can always improve them later, especially if we get some input from someone who has experience with shutters.

PR #433, opened 2023-01-05
[SV] Adding Set climate and enhance area handling

larsdunemark:


sentences/sv/_common.yaml
```diff
@@ -49,6 +49,7 @@ lists:
 expansion_rules:
   name: "{name}"
   area: "{area}"
+  område: "<area>[(et | en | s | met)]"
```
balloob, line 52: Instead of adding a new expansion rule, why not add it to `area` one? It exists exactly for the reason that one might want to add optional things around it.

larsdunemark: ok, will redo it with the area instead, did get the feeling that I should translate it to Swedish also for the expansion_rules from previous PR. And then removing the old area if we don't need it anymore.

sentences/sv/_common.yaml
```diff
@@ -48,13 +48,14 @@ lists:
         out: "fahrenheit"
 expansion_rules:
   name: "{name}[s | n | ns | en ]"
-  area: "{area}[e | me][t | ts | n | ns ]"
```
Robin-St, line 51: Add the extra variants from main

sentences/sv/light_HassLightSet.yaml
```diff
@@ -7,6 +7,6 @@ intents:
           - "[sätt | vrid | dra [på] | skruva | dimmra] [upp | ner ] <name>[n]  till <brightness> [<ljusintensitet>]"
           - "<name>[s | n | ns ] [upp | ner ] [till] <brightness> [<ljusintensitet>]"
       - sentences:
-          - "[sätt | vrid | dra [på] | skruva | dimmra] [upp | ner ] <alla> <ljuskällor> [<i_på>] <area>[et] [upp | ner ] till <brightness> [<ljusintensitet>]"
+          - "[sätt | vrid | dra [på] | skruva | dimmra] [upp | ner ] <alla> <ljuskällor> [<i_på>] <area> [upp | ner ] till <brightness> [<ljusintensitet>]"
```
Robin-St, line 10: Maybe skip this change if we are to merge https://github.com/home-assistant/intents/pull/487 as it fixes this and some more of this intent.

sentences/sv/_common.yaml
```diff
@@ -48,13 +48,14 @@ lists:
...
+  area: "{area}[e | me][t | n ][s]"
   vad: "(vad är | vad är det | vad är det för | vilken)"
   är: "(är det | är)"
   brightness: "{brightness} [percent]"
   slå_på: "( tänd | sätt på | slå på | starta )"
   slå_av: "( släck | stäng av | stoppa )"
   i_på: "(i | på | vid | bredvid)"
+  ändra: "(ställ in | sätt | justera | ändra)"
```
Robin-St, line 58: "vrid ( upp | ner)" Might be a good addition here.

Robin-St (commented):
Looks good, personally i would have missed adding tests for farenheight temperature units. Some minor conflicts with main and other merge requests. 

The definition of the temperature range goes only from 0-100, might be good to extend to at least -50 to 150 to handle freezers and saunas.

tests/sv/climate_HassClimateSetTemperature.yaml
```diff
@@ -1,6 +1,49 @@
 language: sv
 tests:
-- sentences: []
-  intent:
-    name: HassClimateSetTemperature
-    slots: {}
+  - sentences:
+      - "sätt temperaturen till 30°"
+      - "ändra temperaturen till 30°"
```
lellky, line 5: Kan man ens säga ° ? ;)

lellky (approved):
Looks good!

tests/sv/climate_HassClimateSetTemperature.yaml
larsdunemark: I'm not sure how that is suppose to be handled if it is just verification that it is ignored in the parsing of sentences. It is based on the English version of the same test.

sentences/sv/_common.yaml
larsdunemark: I think that I have all combinations that is in main, and also adding [s] as single usages.

tests/sv/climate_HassClimateSetTemperature.yaml
lellky: If it's the same in the english file, it's okay with me. And I just realised I wrote my comment in Swedish :)

Robin-St: I think they are just ignored as % seems to be.

PR #441, opened 2023-01-06
Add temperature and its tests in bulgarian

hristo-atanasov:
Also changed some values for better readability in _common.yaml


dobromir-hristov:
Actually...I have the light files done, so maybe that will be helpful.

PR #449, opened 2023-01-06
Initial changes for ms (Malay Language/Bahasa Melayu)

zubir2k:
Updated common yaml and change ms nativeName to Bahasa Melayu for consideration

### DIFF
--- a/sentences/ms/_common.yaml
+++ b/sentences/ms/_common.yaml
@@ -1,12 +1,12 @@
 language: ms
 responses:
   errors:
-    no_intent: Maaf, saya tidak faham arahan anda
-    no_area: Tiada kawasan {{ area }} dijumpai
-    no_domain: {{ domain }} tiada di kawasan {{ area }}
-    no_device_class: {{ device_class }} tiada di kawasan {{ area }}
-    no_entity: Peranti atau entiti {{ entity }} tidak ditemui
-    handle_error: Maaf, berlaku ralat semasa intent sedang dijalankan
+    no_intent: "Maaf, saya tidak faham arahan anda"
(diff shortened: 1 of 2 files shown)


sentences/ms/_common.yaml
```diff
@@ -1,12 +1,60 @@
...
+      - "celsius"
+      - in: "c"
+        out: "celsius"
+      - "fahrenheit"
+      - in: "f"
+        out: "fahrenheit"
+expansion_rules:
+  name: "[the] {name}"
+  area: "[the] {area}"
+  what_is: "(apa itu | apakah | apa)"
```
balloob, line 52: This one does not wrap a slot name like `{name}` and so the key `what_is` should be updated to be in Malay.

sentences/ms/_common.yaml
```diff
@@ -1,12 +1,60 @@
...
+    values:
+      - "celsius"
+      - in: "c"
+        out: "celsius"
+      - "fahrenheit"
+      - in: "f"
+        out: "fahrenheit"
+expansion_rules:
+  name: "[the] {name}"
+  area: "[the] {area}"
```
balloob, line 51: Does Malay use "the" ?

sentences/ms/_common.yaml
balloob: (this applies to all expansion rules that don't wrap slots)

zubir2k: noted.. removing this for time being. thank you

sentences/ms/_common.yaml
zubir2k: ya.. i only noticed that and have removed [the]. thank you

PR #451, opened 2023-01-06
Updates with synonymous and clean code

iddiek:


auanasgheps:
The checks are failing because you have removed the following

`"quanto [(è | c'è | fa)] (caldo | freddo) [(<in>|<the>) <area>]`

sentences/it/climate_HassClimateGetTemperature.yaml
```diff
@@ -3,9 +3,5 @@ intents:
...
     data:
       - sentences:
-          - "<what_is> <temp>"
-          - "[(<what_is> | che)] <temp> [c'è]"
-          - "<what_is> <temp> [(<in> | <of>)] <area>"
-          - "quanto (caldo | freddo) [(è | c'è | fa)] [(<in>|<the>) <area>]"
-          - "quanto [(è | c'è | fa)] (caldo | freddo) [(<in>|<the>) <area>]"
-          - "[(<what_is> | che)] <temp> [c'è] [<in>] <area>"
+          - "[<what_is>] <temp> [c'è] [(<in> | <of> | <the>)] [<area>]"
+          - "[(è | c'è | fa)] (caldo | freddo) [(<in> | <the>)] [<area>]"
```
balloob, line 7: Per comment by @auanasgheps 
```suggestion
          - "[quanto] [(è | c'è | fa)] (caldo | freddo) [(<in> | <the>)] [<area>]"
```

sentences/it/_common.yaml
```diff
@@ -53,7 +53,7 @@ expansion_rules:
   to: "(a | al | allo | alla | agli | alle)"
   name: "{name}"
   area: "{area}"
-  what_is: "(qual è | quale è | quanto è | quant'è | come è | com'è)"
+  what_is: "(qual è | quale è | quanto è | quant'è | come è | com'è | che)"
```
balloob, line 56: Please in a new PR translate the expansion rules to Italian unless they wrap a slot (ie `{name}`). So change `what_is` to `quanto` or something.

iddiek: Of course is possible but imho is better use en words to slots because, as you can see, there are too many words for "what_is" and they are used in different context. "what_is" btw wrap all of them and I don't think that a truly equivalent exists. Also if translated anyway, we (I) have to work with common.yaml opened as a dictionary.

sentences/it/climate_HassClimateSetTemperature.yaml
```diff
@@ -3,8 +3,8 @@ intents:
   HassClimateSetTemperature:
     data:
       - sentences:
-          - "imposta <temp> a <temperature>"
-          - "imposta [<temp>] [<in>] <area> [a] <temperature>"
-          - "imposta [<temp>] [di] <temperature> [<in>] <area>"
-          - "imposta [<in>] <area> [<temp>] [di] <temperature>"
-          - "Imposta <temp> <area> [a] <temperature>"
\ No newline at end of file
+          - "(imposta | cambia | metti | modifica) <temp> a <temperature>"
```
alessandroias, line 6: Should we have `<set>`?

sentences/it/cover_HassCloseCover.yaml
```diff
@@ -12,10 +12,8 @@ intents:
         slots:
           device_class: "garage_door"
       - sentences:
-          - "<close> [la] (tenda|tapparella) [(<in> | <of>)] <area>"
-          - "<close> [le] (tende|tapparelle) [(<in> | <of>)] <area>"
-          - "<close> [<in>] <area> [la] (tenda|tapparella)"
-          - "<close> [<in>] <area> [le] (tende|tapparelle)"
+          - "<close> [(la | le)] (tenda | tende | tapparella | tapparelle) [(<in> | <of>)] <area>"
```
alessandroias, line 15:
```suggestion
          - "<close> [(la | le)] (tend(a | e) | tapparel(a | e) | venezian(a | e) ) [(<in> | <of>)] <area>"
```

alessandroias (changes requested):

sentences/it/cover_HassCloseCover.yaml
xraver: @alessandroias ok for me

sentences/it/climate_HassClimateSetTemperature.yaml
xraver: I think is good to have!

sentences/it/cover_HassCloseCover.yaml
alessandroias: `veneziane` should be added in every file.
Also, we can use the singular/plural expansion in every file...

xraver: > `veneziane` should be added in every file. Also, we can use the singular/plural expansion in every file...

I agree

xraver: This sentence should be kept:
- "<close> [<in>] <area> [(la | le)] (tend(a | e) | tapparel(a | e) | venezian(a | e) )"

alessandroias:
What about having expansion rules for every entity domain?
Something like:
https://github.com/home-assistant/intents/blob/d397128208d20d69a8fe7a2e09d1d3a21b2894de/sentences/it/_common.yaml#L64

xraver:
> What about having expansion rules for every entity domain? Something like:
> 
> https://github.com/home-assistant/intents/blob/d397128208d20d69a8fe7a2e09d1d3a21b2894de/sentences/it/_common.yaml#L64

I think is a good idea!
That was my starting point.

sentences/it/climate_HassClimateSetTemperature.yaml
auanasgheps: I agree!

auanasgheps:
CI test are still not passing

xraver:
> CI test are still not passing

I see all them passed.

auanasgheps:
> > CI test are still not passing
> 
> I see all them passed.

Now looks good, I can merge!

PR #452, opened 2023-01-06
[GL] Add galician language

cibernox:
I'm adding Galician as a new language. Since it's a new language that implicitly means i'm implicitly applying as Language Leader, at least for now.

I used the Spanish definitions as a starting point and adapted them.


cibernox:
I’ll check the failings tests tonight. Consider this a WIP for the time being.

synesthesiam:
Looks like one `bg` file is included in the PR by accident.

cibernox:
@synesthesiam Fixed. Also I can appoint myself as leader if that's OK.

cibernox:
I believe this is ready to be merged. Should I add myself as language leader in languages.yaml?

PR #460, opened 2023-01-06
Cs branch

schizza:
HassTurnOn/off sentences.


sentences/cs/_common.yaml
```diff
@@ -7,11 +7,60 @@ responses:
...
+  temperature_unit:
+    values:
+      - "(celsius | celsia)"
+      - in: "c"
+        out: "Celsia"
+      - "(fahrenheit | fahrenheita)"
+      - in: "f"
+        out: "Fahrenheita"
 expansion_rules:
+  area: "{area}"
```
halecivo, line 50: trivial expansion is not needed

sentences/cs/_common.yaml
```diff
@@ -7,11 +7,60 @@ responses:
...
-  vypnout: "(vypni|vypnout)"
+  vypnout: "(vypni|vypnout|zastavit|zastav)"
   rozsvitit: "(rozsviť|rozsvítit)"
   zhasnout: "(zhasni|zhasnout)"
+  nastav: "(nastav|nastavit)"
+  teplota: "{temperature} stupňů {temperature_unit}"
+  brightness: "{brightness} [procent]"
+  zavri: "(zavři | zavřít)"
+  otevri: "(otevři | otevřít)"
+  garaz: "garáž"
```
halecivo, line 61: trivial expansion is not needed, also please let's not add unused expansions. To explain why - Your PR is about HassTurnOn/Off, if somebody starts working on curtain, he will add his expansion rules, which may be different from the ones you entered. It also helps to keep PR small.

halecivo (changes requested):
Please also use more informative PR name - here it could be for example "czech HassTurnOn/Off". This helps others to see what is being worked on, so they can pick something else. Thanks!

PS: There is already a PR for HassTurnOn/Off, but since not even tests are passing there, it's a question whether author will finish it.

sentences/cs/_common.yaml
schizza: Ok, but might be needed in future:  area: "[(na|v)] {area}"
```suggestion
  area: "[(na|v)] {area}"
```

sentences/cs/_common.yaml
schizza: Point taken, but I have all sentences done and now working on tests. That's why there are here. Will post test Giles soon for other domains.

sentences/cs/_common.yaml
halecivo: Okay, like this it would make sense.

sentences/cs/_common.yaml
halecivo: Okay, but it is difficult to review unused expansion rules. For garáž (garage), I would say it is a cover device or an area. 
Of course I cannot know for sure, if I don't see the usage.

schizza: Ok, cleaned.

sentences/cs/_common.yaml
```diff
@@ -7,11 +7,56 @@ responses:
...
+        out: "Fahrenheita"
 expansion_rules:
+  area: "[(na|v)] {area}"
   zapnout: "(zapni|zapnout)"
-  vypnout: "(vypni|vypnout)"
+  vypnout: "(vypni|vypnout|zastavit|zastav)"
   rozsvitit: "(rozsviť|rozsvítit)"
   zhasnout: "(zhasni|zhasnout)"
+  nastav: "(nastav|nastavit)"
+  teplota: "{temperature} [stupňů] {temperature_unit}"
```
halecivo, line 56: Better would be 
`{temperature} [stupňů] [{temperature_unit}]`
Your version allow to say "15 Celsia", but not "15 stupňů", which is quite common.
Not a critical comment, however.

sentences/cs/homeassistant_HassTurnOff.yaml
```diff
@@ -2,4 +2,6 @@ language: cs
 intents:
   HassTurnOff:
     data:
-    - sentences: []
+      - sentences: 
+          - "<vypnout> <name>"
```
halecivo, line 6: Name is not expansion rule, therefore I think the tests will fail here.
`<vypnout> {name}` should work.

halecivo (changes requested):

sentences/cs/_common.yaml
schizza: Yes, found it when testing climate :)
Corrected.

schizza:
Now I don't know if I need to apply for PR?

halecivo:
The problem is that you pushed to the same branch another change, therefore this PR had been updated and is no more focused on a single domain.

If you start a new work, you need to create a new branch, so you can have separate PR.

In Github desktop, you can create new branch from your current state, push it, then switch back to the old branch (cs-branch), drop/revert commits for different domain than HassTurnOn/Off.

halecivo:
Well, since #279 has been merged right away, this PR can be abandoned.

PR #464, opened 2023-01-06
Hassil 0.2

synesthesiam:


TheFes (approved):
Looks good, grammatically correct this way :)
But it seems to work with and without the space anyway (ran into that when I created a sentence to turn on lights in an area)

synesthesiam:
@TheFes With hassil 0.1 the space was (unintentionally) optional. But not it's more strict and predictable :slightly_smiling_face:

synesthesiam:
@haim-b, @leranp - I've updated the intent matcher, which is now a bit more strict. I had to comment out some of the tests for the merge. Can we work together to figure out what's wrong?

TheFes:
@synesthesiam What will happen when there are two possible sentences that can match?

entity name: kitchen lights
area name: kitchen

sentences:
`Turn <area> lights on`
`Turn <name> on`

Both will match with:
`Turn kitchen lights on`

This is what I ran into when the space was optional, but while `slaapkamerlamp` is grammatically correct, there are a lot of people who might use `slaapkamer lamp` as it is a very commonly made mistake in Dutch

PR #466, opened 2023-01-06
[FR] Add sentences for opening and closing covers

slegars56:


sentences/fr/_common.yaml
```diff
@@ -52,6 +52,13 @@ expansion_rules:
...
+  ouvre: "(ouvre | lève | monte | remonte)"
+  ferme: "(ferme | referme | tire | descend[s] | redescend[s])"
+  artdef: "(le | la | les | l')"
+  artindef: "(dans [le] | de | du | des)"
+  ouverture: "(stores | rideaux | volets | persiennes)"
+  porte: "(porte | portail)"
 skip_words:
   - "s'il te plaît"
   - "merci"
+  - "est ce que tu peux"
```
acesyde, line 64: - peux tu
- pourrais tu

sentences/fr/_common.yaml
```diff
@@ -52,6 +52,13 @@ expansion_rules:
   lumieres: "[les] (lumières | lampes | ampoules)"
   ventilateur: "(le ventilateur | le brasseur d'air)"
   ventilateurs : "(les ventilateurs | les brasseurs d'air)"
+  ouvre: "(ouvre | lève | monte | remonte)"
+  ferme: "(ferme | referme | tire | descend[s] | redescend[s])"
+  artdef: "(le | la | les | l')"
```
acesyde, line 57: Il serait préférable si on utilise le artdef de refacto partout (ventilateur / lumière / ...)

piitaya (changes requested):
I don't think we must use artdef artindef. It add complexity to sentences. Let's keep this PR focus on covers actions with the same logic as we do for lights.

slegars56:
I disagree, I like this way of managing articles, but as mentioned, it should be aligned with other intents.
I give up these expansion rules and close my PR

PR #468, opened 2023-01-06
Add Set Temperature and its tests in bulgarian language

hristo-atanasov:
Also closed and unclosed bracket in _common.yaml


hristo-atanasov:
There was a missing closing bracket in HA repository. This PR also resolves this bug.

PR #487, opened 2023-01-07
[SV] intents for max/min light intensity

Robin-St:
First a minor refactor of moving synonyms to "dim" to commons. 

Then an attempt to catch  intents for maximum and minimum intensity for lights.

It passes all tests but i have not seen it done in any other language so i dont know if it is correct. I used slots to set brightness to 100 or 1 as in the following example:

      - sentences:
          - "dim <name> to max brightness"
        slots:
          brightness: 100


sentences/sv/light_HassLightSet.yaml
```diff
@@ -3,10 +3,44 @@ intents:
...
+          - "<name> [upp | ner ] [till] <brightness> [<ljusintensitet>]"
       - sentences:
-          - "[sätt | vrid | dra [på] | skruva | dimmra] [upp | ner ] <alla> <ljuskällor> [<i_på>] <area>[et] [upp | ner ] till <brightness> [<ljusintensitet>]"
+          - "[<dimra>] [upp | ner] [<alla>] <ljuskällor> [<i_på>] <area> [upp | ner ] till <brightness> [<ljusintensitet>]"
+          - "[<dimra>] [upp | ner] [<alla>] <ljuskällor> till <brightness> i <area>"
         slots:
           name: "all"
+
+      - sentences:
+          - "[<dimra>] [upp] <name> [<ljusintensitet>] till <maximal> [<möjliga>]"
```
larsdunemark, line 16: [] around <möjliga> is redundant since <möjliga> contains everything inside it's own [], we should probably keep one set of [] and review the expansion_rules to:
`
  möjliga: "(möjlig[a] | tillgänglig[a])"
`
So it's up to the usage to decide if it should be optional or not.

larsdunemark (commented):
I think that we should check on the discord for more feedback around the usage for max/min I think it looks good but not sure if there is a bigger plan for handle this in the future or not.

sentences/sv/light_HassLightSet.yaml
Robin-St: Removed the redundant [] in a new commit. 

@synesthesiam is this method to find the intents for maximum and minimum brightness without having to say 100/1 percent correct usage of the intents?

larsdunemark:
Setting a light to maximal seems simple but I'm not sure if 1% is suitable for minimum, feels like it can require more knowledge about the light or a different intent to handle this use case. It might be that a minimum for one light could be 10% to be visible and 20% for an other.

Robin-St:
I think that is another problem than interpreting the intent.

sentences/sv/_common.yaml
```diff
@@ -51,14 +51,19 @@ expansion_rules:
   area: "{area}[e | me][t | ts | n | ns ]"
   vad: "(vad är | vad är det | vad är det för | vilken)"
   är: "(är det | är)"
-  brightness: "{brightness} [percent]"
+  brightness: "{brightness} [percent | procent]"
```
larsdunemark, line 54: should we keep the English version for percent? Seems like we might want to limit the scope to Swedish and as I understand it it would be possible for end users to add their own commands also.

Robin-St: Fixed :)

sentences/sv/_common.yaml
```diff
@@ -51,14 +51,19 @@ expansion_rules:
...
   slå_av: "(slå (av | från | ifrån) | släck | stäng av | stoppa)"
   i_på: "(i | på | vid | bredvid)"
   temp: "(temperatur)"
   alla: "(all | alla | allt)"
-  ljusintensitet: "(ljusstyrka | intensitet | styrka | nivå | effekt)"
-  ljuskällor: "(ljuskällor[na] | lampor[na] | lampa[n]| belysning[en] | ljus[en|et] | glödlampor[na] | lysrör[en|et])"
+  möjliga: "(möjlig[a] | tillgänglig[a])"
+  ljusintensitet: "(ljus[et] | belysning[en] | [ljus](styrka | intensitet | nivå | effekt))"
+  lampa: "(ljuset | belysning[en] | lampan | lampor[na])"
+  ljuskällor: "(ljuskällor[na] | belysning[en] | ljus[en | et] | [glöd|led|diod]lamp[ a | an | or | orna] | lysrör[en| et] | belysning[en])"
```
larsdunemark, line 63: `belysning[en]` is listed twice.

sentences/sv/_common.yaml
```diff
@@ -51,14 +51,19 @@ expansion_rules:
...
+  dimra: "(sätt | vrid | dra [på] | skruva [på] | dimra | dimma | ställ [in] | höj | sänk)"
+  maximal: "(maximal | max[imum] | högsta | starkaste | intensivaste | full)"
```
larsdunemark, line 65: Not sure but maybe "`full"` also should cover `"full[t]"`

larsdunemark (commented):
I think that we might in the future add setting a light to half or something similar. But I think that it should be as a separate MR.

sentences/sv/_common.yaml
Robin-St: Oh, lampa has sneaked back from main while fixing merge conflicts

sentences/sv/_common.yaml
Robin-St: Added.

sentences/sv/_common.yaml
larsdunemark: great, but we still have `belysning[en]` twice so we should probably remove the last one.

Robin-St:
I think it is time to give up on this PR soon, tests fails here and running the branch locally gives a lot of different error for domains not changed by this PR..

Robin-St:
ERROR tests/test_language_sentences.py::test_climate_HassClimateSetTemperature[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_homeassistant_HassTurnOn[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_light_HassTurnOff[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_cover_HassCloseCover[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_fan_HassTurnOff[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_light_HassLightSet[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_light_HassTurnOn[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_climate_HassClimateGetTemperature[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_fan_HassTurnOn[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_homeassistant_HassTurnOff[sv] - ValueError: too many values to unpack (expected 2)
ERROR tests/test_language_sentences.py::test_cover_HassOpenCover[sv] - ValueError: too many values to unpack (expected 2)

tetele:
Run script/setup in your dev container. The PIP dependencies are out of date.

Robin-St:
> Run script/setup in your dev container. The PIP dependencies are out of date.

Thanks, that solved the weird errors, thou still gets some tests that suddenly fails. Probably something missed when keeping the branch up to date with main. Copied some of the edits to a fresh branch of main and submitted a new PR

sentences/sv/light_HassLightSet.yaml
synesthesiam: Yes, here's how I would do it in English.
In `_common.yaml` create a new list:

```yaml
lists:
  brightness_min_max:
    values:
      - in: "(max | maximum)"
        out: 100
      - in: "(min | minimum)"
        out: 1
```

Then you reference the list in a sentence:

```yaml
intents:
  HassLightSet:
    data:
      - sentences:
          - "set <name> brightness to [the] {brightness_min_max:brightness}"
```

Note that `{brightness_min_max:brightness}` means to use the `brightness_min_max` list but put the result into the intent slot named `brightness`.

And now this will work:

```sh
python3 -m script.intentfest parse --language en --sentence 'set bedroom lamp brightness to max'
{
  "text": "set bedroom lamp brightness to max",
  "match": true,
  "intent": "HassLightSet",
  "slots": {
    "name": "light.bedroom_lamp",
    "brightness": 100
  }
}
```

Robin-St: @synesthesiam That looks like a much cleaner way to do it. I tried to add it to swedish but the tests/linter seems to not like the ":" as i get the error below. But testing with  intentfest as you showed does return the expected responese.

```
Running tests for file light_HassLightSet.yaml

Validation failed

Language: sv
 - sentences/sv/light_HassLightSet.yaml: invalid format: Sentences should only contain words and matching syntax. They should not contain punctuation. @ data['intents']['HassLightSet']['data'][2]['sentences'][0]. Got '[<dimra>] [upp | ner] <name> [<ljusintensitet>] till {brightness_level:brightness} [<möjliga>]'
```

PR #491, opened 2023-01-07
Add "släck" as a keyword to turn off lights in Swedish

toffbiffle:


Robin-St:
Hi, @toffbiffle. "Släck" is already a keyword for this. If you look in the file _common.yaml you will see a line as below that is referenced to by the <slå_av> tag in the intent sentence. The "<>" tags  are used to refers to synonyms and simmilar constructions to not have to write every variant of a sentence in full. 

  slå_av: "( släck | stäng av | stoppa )"

But looking at that line and in the adjecent slå_på list there are a some good words to add if you like :) 

https://github.com/home-assistant/intents/blob/2b3df8a84198b297b450be8b73431e91c3e732e7/sentences/sv/_common.yaml#L56

toffbiffle:
> Hi, @toffbiffle. "Släck" is already a keyword for this. If you look in the file _common.yaml you will see a line as below that is referenced to by the <slå_av> tag in the intent sentence. The "<>" tags are used to refers to synonyms and simmilar constructions to not have to write every variant of a sentence in full.
> 
> slå_av: "( släck | stäng av | stoppa )"
> 
> But looking at that line and in the adjecent slå_på list there are a some good words to add if you like :)
> 

Ah thanks for the clarification, I tried using the "tänd" and "släck" keywords but my google assistant didn't respond with the correct action, instead it showed me a search for the term, and therefore I assumed that it wasn't in the specification. 

I didn't see the _common.yaml file at all in my hurry to contribute. :-D

PR #445, opened 2023-01-06
Update climate_HassClimateGetTemperature.yaml

diplix:
added some variations

PR #501, opened 2023-01-07
Update _fixtures.yaml

diplix:
added climate device


mib1185:
this seems to be part of #502 which should go directly in #445.

related changes should always go into the same PR, so they can be reviewed together and in this case they also depends on each other, so each single PR is invalid (_because of failing tests_)

PR #505, opened 2023-01-07
[LV] Add light domain

makstech:


balloob:
Want to be language leader for Latvian ?

PR #508, opened 2023-01-07
Add Danish climate sentences

fadamsen:


MTrab (approved):
You are fast - looks good to me

PR #515, opened 2023-01-08
Neo-branch

BossNeo:
1. edit on error messages, to be more native without using zones
2. merge all collors in proper short format
3. removing word Пусни from <close> on line 57 becouse its same like <turn_on> word on line 54 and change it with proper Свали word. Пусни  word on turn on action and close action will make a conflict. 
4. adding expansion rukes what_is


hristo-atanasov (commented):
Take a look how `[]` and `()` are used in the places where I changed them. Also in places, where You have `[()]` the nested round brakets are not currently required and I removed them. I guess You will appreciate, that I modified Your files, instead of rejecting the PR again. This will not happen on Your other PR. You will have to modify it by Yourself.

PR #529, opened 2023-01-08
[SV] Open/Close cover

Robin-St:
Added open and close cover for swedish


larsdunemark:
Looks good to me, have an outstanding question in discord about how device_classes should be used and if it should be template instead, seems like we get some hard coded just for garage_door and not any other device_class.

tests/sv/cover_HassCloseCover.yaml
```diff
@@ -1,6 +1,40 @@
...
+      slots:
+        name: cover.bedroom_curtain
+
+  - sentences:
+      - "Stäng garagedörren"
+      - "Dra igen garagedörren"
+    intent:
+      name: "HassCloseCover"
+      slots:
+        device_class: "garage_door"
```
larsdunemark, line 22: I think that we can change the device_class to match: "garage" looks like it might be an issue in other languages also.
https://developers.home-assistant.io/docs/core/entity/cover/#device-classes

larsdunemark (approved):
Looks good

PR #534, opened 2023-01-08
[PL] Few changes related with polish name of values, turn_on, turn_off, temp, garage_door and lights sentences

witold-gren:


sentences/pl/_common.yaml
```diff
@@ -10,23 +10,23 @@ responses:
...
+      - in: "pomarańcz | pomarańczow(y|o|e)"
         out: "orange"
-      - in: "żółt(y|o)"
+      - in: "żółt(y|o|e)"
         out: "yellow"
-      - in: "zieleń | zielon(y|o)"
+      - in: "zieleń | zielon(y|o|e)"
         out: "green"
-      - in: "niebiesk(i|o)"
+      - in: "niebiesk(i|o) | niebieskie"
```
Uriziel01, line 25: Can't we just use a `ie` in the original form like `niebiesk(i|o|ie)`?

sentences/pl/light_HassLightSet.yaml
```diff
@@ -12,3 +12,5 @@ intents:
       # color
       - sentences:
           - "(Ustaw | Zmień) [kolor] <name> na {color}"
+      - sentences:
+          - "(Ustaw | Zmień) [wszystkie] [kolory] <name> na {color}"
```
Uriziel01, line 16: You can just allow multiple different phrases in one `Optional word` syntax like `[kolor|kolory]` no need for additional sentence.
If in doubt you can check https://voice.home-assistant.io/docs/intents/matching-syntax/#sentence-templates-syntax for formatting tips.

Uriziel01: And while we are here you've added a new prefix `Przestaw` in other places 😉 

At this point this should probably be added to the https://github.com/home-assistant/intents/blob/main/sentences/pl/_common.yaml as a expansion_rule to not repeat ourselves all over the place.

balloob:
The CI errors are probably due to two sentence templates matching the same sentence. One way this can be avoided is to make sure fixtures do not have generic names ([background why](https://voice.home-assistant.io/docs/intents/test-syntax#fixtures))

sentences/pl/_common.yaml
witold-gren: Yes, later I noticed that I can do that. I will try to improve it today

sentences/pl/light_HassLightSet.yaml
witold-gren: One problem which I see is that in polish language we not say: "Zmień wszystkie kolor lamp na czerwone." The correct one is "Zmień wszystkie kolory lamp na czerwone." That was the reason why I created new sentence 😀 We should some how cover this differences

Uriziel01: The issue here is even more with the entire sentence word order not even this particular word (which you could solve with nesting the optional phrase like so `(Ustaw | Zmień) [kolor | [wszystkie] kolory] <name> na {color}`).
But even then you end up with a sentence like in your example `Zmień wszystkie kolory lamp na czerwone.` and AFAIK it won't work like that because it lacks the device name and you are trying to use something like device_class here. Did you manage to test it? Could you show an example of proper use please?

witold-gren: Yep, I will check it and prepare correct sentence

witold-gren: OK, you have a right, I forgot add area to this sentence. Now I updated tests and added more examples. 

PS. Looking through the tests, I noticed wrong device names for the tests, I also corrected them 😀

witold-gren: BTW. Regarding prefixes I reorganised `expansion_rules` and I update tests 😉 I didn't want to go too crazy because I don't know how deep we want to go into this. Now it's worth adding a bit more tests

Uriziel01: Great, now it looks amazingly good, will check everything today's evening but at the first glance it's all good. Thank you for your contribution Witold 👍

sentences/pl/_common.yaml
```diff
@@ -46,11 +46,17 @@ lists:
...
   area: "[w|na] {area}"
   what_is: "Jak[a|i] jest"
+  all: "(wszystkie | całe)"
   how_many_is: "Ile jest"
-  turn_on: "(włącz | zapal | oświeć | uruchom )"
-  turn_off: "(wyłącz | zgaś | zatrzymaj )"
+  turn_on: "(włącz | zapal | oświeć | uruchom)"
+  turn_off: "(wyłącz | zgaś | zatrzymaj)"
+  open: "(otwórz)"
+  open_blind: "(odsłoń | podnieś)"
```
Misiu, line 57: Just wondering, should we also add `otwórz` to open_blinds, or will it work?
User may say `otwórz rolety w sypialni` (open blinds in the bedroom) or `zamknij rolety w symialni` (close blinds in the bedroom).

Uriziel01: Yes `otwórz` should and in fact is a part of blinds opening extensions 😄 
This is why @witold-gren used the `"(<open> | <open_blind>) rolety <area>"` syntax. I really like this approach because it gives you full flexibility by both covering the sentences like in your example while also keeping a clear separation between general extensions and these for blinds.

sentences/pl/_common.yaml
```diff
@@ -46,13 +46,19 @@ lists:
...
   name: "{name}"
-  area: "[w] {area}"
-  what_is: "Jak[a] jest"
+  set: "(ustaw | zmień | przestaw)"
+  area: "[w|na] {area}"
+  what_is: "Jak[a|i] jest"
+  all: "(wszystkie | całe)"
   how_many_is: "Ile jest"
-  turn_on: "(włącz | zapal | uruchom)"
+  turn_on: "(włącz | zapal | oświeć | uruchom)"
```
Uriziel01, line 54: `turn_on` is a general extension used in all sort of situations and as we do not want to `oświeć` a ceiling fan a toaster, please remove it or move it outside of this extension.

sentences/pl/_common.yaml
Misiu: @Uriziel01 thank you for a quick reply. I need to read how to test the sentences, this will probably answer many questions before posting them here.
What about this part - https://github.com/home-assistant/intents/pull/534/files#diff-8cdabf72346a4b01ae6dafb0a8b29b8d72b1c5078c2726219659fcf57d61043dR15
`"(<open> | <open_blind>) rolety <area>"`
We should be able to say/type `Otwórz rolety salon`, but what about `Otwórz rolety w salonie`?
We will be able to add aliases to entities and areas in HA, but what about `w`?

Uriziel01: No problem. The documentation mentioned in readme file in really good, you can use that to learn the syntax and how to test it properly.
And to answer your question - this one is handled by the area extension here `area: "[w|na] {area}"` so it covers cases like `w kuchni` or `na balkonie`. Of course after you declare all the forms for an area in https://github.com/home-assistant/intents/blob/main/tests/pl/_fixtures.yaml

Misiu: I've run some tests, I've added two covers to living room (in _fixtures), and I have another set of questions:
`otwórz rolety w salonie` works fine, but `otwórz wszystkie rolety w salonie` doesn't return match.
Same with `otwórz wszystkie rolety`.
Maybe let's discuss this here: https://github.com/home-assistant/intents/discussions/388

Uriziel01: This is because this sentence https://github.com/home-assistant/intents/pull/534/files#diff-8cdabf72346a4b01ae6dafb0a8b29b8d72b1c5078c2726219659fcf57d61043dR16 uses another word ordering. And yes, I agree that putting `wszystkie` before the area extensions would be way more natural way of saying it.

Misiu: so we should add another option, like this:
`"(<open> | <open_blind>) [wszystkie] rolety <area>"`
I think a similar approach should be used with other platforms, you can turn on a single light source or all lights in a specific room (area) or even in entire house

Uriziel01: It is added in some places, not all though. To not get @witold-gren PR even more complicated than it is currently I suggest to only fix ordering here to a correct one and leave the rest of cases for now.

I want to avoid having hanging PRs for prolonged amount of time because we figure out new ways to make it better.

sentences/pl/_common.yaml
witold-gren: this problem was fixed

sentences/pl/_common.yaml
witold-gren: hey, I also fixed problem with `(<open> | <open_blind>) [wszystkie] rolety <area>`

Uriziel01:
LGTM and all the tests passed. Merging now.

sentences/pl/_common.yaml
witold-gren: OK, so I also added support for "all devices" without set area. So now we can say `Zamknij wszystkie rolety` , `Wyłącz wszystkie wentylatory` or `Zgaś wszystkie światła` 😉

witold-gren: My proposition is just close this PR and not add more rules, because now it is quite big. If we need add more functionality we should create another PR or list of requested feature..

Uriziel01: It's already merged 48mintues ago @witold-gren 😉

PR #541, opened 2023-01-09

anaglic:


Robin-St:
You need to add some tests for the intents to this file https://github.com/home-assistant/intents/blob/main/tests/sl/homeassistant_HassTurnOn.yaml :)

PR #544, opened 2023-01-09

pmentis:


dsimop:
This PR as well as the #545 cannot be approved in the current phase as we have to follow the official guidelines. That is, any sentence with its respective test have to be committed into a single PR. In that way, we can see if the tests pass.

dsimop (commented):
This PR as well as the https://github.com/home-assistant/intents/pull/545 cannot be approved in the current phase as we have to follow the official guidelines. That is, any sentence with its respective test have to be committed into a single PR. In that way, we can see if the tests pass.

PR #567, opened 2023-01-09
[RO] TurnOn renaming and diacritic support refactoring

tetele:
Replaced "inchide" and "deschide" with "opreste" and "porneste" for better generic meaning.

Also, diacritics should be provided as pipe-separated options in-place, as opposed to different words, for brevity.


gabimarchidan:
Please merge. Love the changes !

PR #434, opened 2023-01-05
Update climate_HassClimateGetTemperature.yaml

Kalma-House:
More ways ta ask about temperature.

PR #573, opened 2023-01-10
Adding extra ES sentences to get temperatures

davefx:


davefx:
Thanks to @cibernox and @Kalma-House for their comments at #434 . Closing now that PR.

PR #591, opened 2023-01-10
Migrate pyproject

synesthesiam:
Use pyproject.toml instead of setup.py
Change to date versioning
Add parameter for json loader in get_intents


package/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools~=62.3", "wheel~=0.37.1"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name        = "home-assistant-intents"
+version     = "2022.1.0.dev0"
```
synesthesiam, line 7: @balloob  Not sure if this is what we want to do or not for versioning.

balloob: yeah that works.

package/home_assistant_intents/__init__.py
```diff
@@ -2,28 +2,24 @@
...
-except (ImportError, AttributeError):
-    # Backport for Python < 3.9
-    import importlib_resources  # type: ignore
-
-    files = importlib_resources.files
+import importlib.resources
 
 _PACKAGE = "home_assistant_intents"
-_DIR = Path(typing.cast(os.PathLike, files(_PACKAGE)))
+_DIR = Path(typing.cast(os.PathLike, importlib.resources.files(_PACKAGE)))
```
balloob, line 10: Why can't we do `Path(__file__).parent` ?

balloob (approved):

package/home_assistant_intents/__init__.py
synesthesiam: It should allow the package to be zip safe, which I would need to mark in the `pyproject.toml`.

PR #596, opened 2023-01-10
[IT] Add Cover Synonyms

alessandroias:
Add `saracinesca` and `cler` as synonyms for `garage door`.
`cler` is used a lot in northern Italy.


xraver:
I don't understand why it does not pass CI....

xraver (commented):
what does it means "cler"?

alessandroias:
It means "shutter"... I'm afraid it's a dialectal form, but it's so used here (in Milan) that it confused me.

auanasgheps:
Looks like another language (bulgarian) is blocking the test.
Please wait until it is resolved before merging.

`Cler` is not Italian, we're not allowing dialects, otherwise the Italian files would be long like an encyclopedia...

xraver:
> Looks like another language (bulgarian) is blocking the test. Please wait until it is resolved before merging.
> 
> `Cler` is not Italian, we're not allowing dialects, otherwise the Italian files would be long like an encyclopedia...

I agree.
thanks

alessandroias:
As soon as I'm at home :)

PR #599, opened 2023-01-11
Add Finnish

huusissa:


synesthesiam:
@huusissa just need at least one test for each sentence in HassTurnOn and HassTurnOff

huusissa:
Now all tested.

PR #600, opened 2023-01-11
Update climate_HassClimateGetTemperature.yaml

chuk-a:


frenck (changes requested):
@chuk-a Can you please add some tests for these sentences?

Thanks already 👍 

../Frenck

Issue #575, opened 2023-01-10
Similar phrases for different device_classes produces error


tetele:
Let's say we have the following entities (in Romanian):
`tests/ro/_fixtures.yaml`
```
areas:
  - name: Garaj
    id: garage
entities:
  - name: intrarea dinspre garaj # meaning "entrance from the garage"
    id: cover.garage_backdoor
    device_class: door
    area: garage
  - name: rulou garaj
    id: cover.garage_door
    device_class: garage
    area: garage
```

And the following sentences to operate them
`sentences/ro/_common.yaml`
```
expansion_rules:
  area: "{area} [(zona | zonă | regiune)]"
  name: "({name}[ul] | {name})"
  din: "(din | (î|i)n | pentru | [de] la)"
  inchide: "((î|i)nchide | coboar(ă|a))"
  usa: "(u(ș|s)a | u(ș|s)ile)"
  poarta: "(poarta | gardul)"
```

`sentences/ro/cover_HassCloseCover.yaml`
```
intents:
  HassCloseCover:
    data:
      - sentences:
          - "<inchide> <usa>"
          - "<inchide> <usa> [<din>] <area>"
        slots:
          device_class: door
      - sentences:
          - "<inchide> <poarta>"
          - "<inchide> <poarta> [<din>] <area>"
        slots:
          device_class: gate
      - sentences:
          - "<inchide> (<poarta> | <usa>) [<din>] garaj"
        slots:
          device_class: garage
```

Say I wanted to close the garage door. Id' say "inchide usa de la garaj"="close the garage door", which should match the last sentence in the set above. Here's the test for this scenario:
`tests/ro/cover_HassCloseCover.yaml`
```
tests:
  - sentences:
      - "inchide usa de la garaj"
    intent:
      name: HassCloseCover
      slots:
        device_class: garage
```

Just because I have an area called "garaj"="garage" and a specific device_class "usa"="door" for the `cover_HassCloseCover` intent, that means my sentences will not get properly matched.

```
E               AssertionError: Slots do not match for: inchide usa de la garaj
E               assert {'area': 'gar...lass': 'door'} == {'device_class': 'garage'}
E                 Differing items:
E                 {'device_class': 'door'} != {'device_class': 'garage'}
E                 Left contains 1 more item:
E                 {'area': 'garage'}
E                 Use -v to get more diff

tests/test_language_sentences.py:75: AssertionError

[...]

PR #608, opened 2023-01-11
Verify slots with multiple possible values in tests

synesthesiam:
Fix for issue: https://github.com/home-assistant/intents/issues/575

Sentences may specify a list of `device_class` options, for example:

```yaml
language: "en"
intents:
  HassCloseCover:
    data:
      - sentences:
          - "close [the] blinds in <area>"
        slots:
          device_class:
            - "blind"
            - "curtain"
            - "shutter"
```

Previously, each test for these sentences had to include *all* of the `device_class` options, so this **didn't** work:

```yaml
language: "en"
tests:
  - sentences:
      - "close the blinds in the bedroom"
    intent:
      name: "HassCloseCover"
      slots:
        area: "bedroom"
        device_class: "blind"
```

Now the following scenarios are supported (for each slot):

1. Template and test have single values for a slot: must match exactly
2. Template has list, test has single value: test value must be in template list
3. Template has list, test has list: test list must be subset of template list
4. Template has single value, test has list: error


tetele:
Issue #575 was referring to having a zone named the same as a particle in the sentence used for triggering a certain `device_class`-related intent.

Case in point, if you want to open a garage door (`device_class="garage"`) by saying `"open the garage door [in [the] {area}]"`, but you have an area called `garage` with an entity with `device_class="door"` and you have a sentence defined for `open_cover, device_type="door"` which sounds something like `"open [the] {area} door"`, then there will be confusion regarding the intent.

Note that this only applies with inferred area context, i.e. saying "open the garage door" in the garage, in case the area is implied. Will that ever be supported?

tests/test_language_sentences.py
```diff
@@ -71,10 +71,36 @@ def do_test_language_sentences_file(
...
+                        ), "Slots do not match for: {sentence}"
+                    else:
+                        assert (
+                            actual_value in match_value
+                        ), f"Slot {match_name} must be one of {match_value} for: {sentence}"
+                else:
+                    # Only one acceptable value
+                    assert (
+                        actual_value == match_value
+                    ), f"Expected {match_value}, got {actual_value} for slot {match_name} for: {sentence}"
```
balloob, line 97: Can you check this firt, and then add `continue` to save on indentation ?

synesthesiam:
> Will that ever be supported?

I think the problem is the template: `"open the garage door [in [the] {area}]"`
Maybe I'm not thinking of all the use cases, but it seems like following a few simple rules will help:

1. If an `area` is optional, then a `name` must be required
2. If an `area` is required, then either a `domain` or `device_class` must also be required

tetele:
@synesthesiam I agree with your rule set from a programmatic standpoint, i.e. matching sentences to intents. However, regarding the actual wording, my point is that the `area` may not always be included in the sentence due to context.

Say you are in the bedroom and want to turn on the lights. If you had a voice satellite that's defined as "in the bedroom", would you really care to say "turn on the lights in the bedroom"? I think it's far more natural to just say "turn on the lights". That's how Google Assistant does it and I think it comes very naturally.

In the example above, the sentence `"turn on [the] light[s] [in [the] {area}]"` can be defined as having a required `area`, but the area may or may not be uttered by the user. If it's not actually spoken, then it should be inferred from context (i.e. the receiving voice satellite is in the area for which you're trying to turn on the lights). Nevertheless, the sentence should not be matched to a `light_HassTurnOn` intent if there's no area provided by neither the context, nor the user.

Does that make more sense?

synesthesiam:
@tetele I agree, let me add one more rule:

3. If no `area` or `name` is present, then `requires_context` must include `area`

Now, instead of `"turn on [the] light[s] [in [the] {area}]"` you would have two templates:

* `"turn on [the] light[s] [in [the]] {area}"`
* `"turn on [the] light[s]"`
    *  `requires_context` with `area`

This lets you say "turn on the lights in the bedroom" from anywhere in the house, and "turn on the lights" when you are in the bedroom.

tetele:
Sounds good.

balloob:
I am not sure if this is something that should be defined on the level of template sentences. It sounds like something the Intent Handler in HA should deal with.

PR #614, opened 2023-01-11
Merging treatment of "gate", "garage_door" and "door" device classes

Kalma-House:
I feel that the device classes "gate" and "garage-door" should be handled the same way, along with device class 'door'. Right now, the the only difference is the addition of 'garaje' or 'cochera' at the end of the sentence for device class 'garage_door'. Frankly, I think 'garaje' or 'cochera' are the names of areas, and they should be treated as such.

### DIFF
--- a/sentences/es/cover_HassOpenCover.yaml
+++ b/sentences/es/cover_HassOpenCover.yaml
@@ -6,7 +6,8 @@ intents:
           - "<abre> <name>"
           - "<abre> <name> <area>"
       - sentences:
-          - "<abre> <puerta> [<area>]"
+          - "<abre> <puerta>"
+          - "<abre> <puerta> <area>"
         slots:
           device_class:
             - "gate"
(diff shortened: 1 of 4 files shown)


davefx:
I'm really not sure about this: the different treatment for garage_door comes from the original code. 
Your reasoning is also valid in English, but the original English intents have the one for the garage door as a totally different intent, not considering "garage" in "close the garage door" an area, but "garage door" as a totally different entity...

That's why we currently have it as a different intent.

I think this is not a discussion just for Spanish, but it would also involve many (if not all) languages, and we should act in a coordinated way.

@synesthesiam Could you give your opinion about this?

sentences/es/cover_HassOpenCover.yaml
```diff
@@ -5,14 +5,13 @@ intents:
...
-          - "<abre> <puerta> [(del|de|de la)] (garaje|cochera)"
-        slots:
-          device_class: "garage_door"
       - sentences:
           - "<abre> <puerta> [<area>]"
         slots:
-          device_class: "gate"
+          device_class:
+            - "gate"
+            - "garage_door"
```
spuljko, line 13: Device class for garage is "garage" (https://www.home-assistant.io/integrations/cover/)

frenck:

../Frenck

PR #620, opened 2023-01-12
[BG] merge and add missing fuctuinality

BossNeo:
1.add a  <set><area><light> functionality to now was only <set><light><area> do that for turn on and off too
2. took all bg meaning for (all| every) under <all>
3. add termostat name to set temperature in room
4.took all light names in common under <lapi>
5. add another native name to turn on lights " запали"
6. add a suggsesion change for new RP in cliamte
7. change all (all and every) word do be under <all>
8. change all lights meaning under <lampi>
9. Fix kitchen grama
10. add test sentences for <action> <area> <name> fuctionality .


BossNeo:
Boys what you thing if we add (е , на , до ) to skip words , we aways put them in [] to can be skip

hristo-atanasov (commented):
Not sure if the `термостат` should be included in `temperature` template. Termostat can be the device or the entity. But over all it is ok. Also currently there is no device types translations so I think this ok for now. In future it could lead to conflicts.

hristo-atanasov:
> Boys what you thing if we add (е , на , до ) to skip words , we aways put them in [] to can be skip

Skip words are to be skipped. The words in brakets are used for matching templates and with these brakets you'te lling the ML, that the user might ommit these words at this part in the sence, but he can also use it. So, no -  we don't move them in skip words.

PR #622, opened 2023-01-12
Improve error message for unknown areas

tetele:
If an entity defined as a fixture references an area that's not
defined as a fixture, name that area as opposed to the entity ID


synesthesiam (approved):

synesthesiam:
Better late than never :grimacing:

PR #636, opened 2023-01-12
[EN] add max/min brightness sentences

Robin-St:
Add sentences that look for max/min commands instead of a percentage number for set light.


balloob (approved):
Nice! 🐬

PR #637, opened 2023-01-12
[SV]Add max min light

Robin-St:
Add ability to set max and min light by words instead of saying 1/100.


edgein666 (commented):
I think it looks good.

larsdunemark (approved):
Looks good!

PR #655, opened 2023-01-15
[PL] New expansion rules and usage in multiple tests, some optimizations

Uriziel01:
Added new expansion rules and applied them in multiple tests, some optimizations


Uriziel01:
@witold-gren @Wojciechgc @rrakso Could some of you please take a look at this, it would be plain unethical for me to just accept my own changes 😀 Thank you!

tests/pl/cover_HassCloseCover.yaml
```diff
@@ -27,6 +27,7 @@ tests:
       - "Zasłoń rolety w sypialni"
       - "Zasłoń wszystkie rolety w sypialni"
       - "Zamknij w sypialni wszystkie rolety"
+      - "Przymknij całe rolety w sypialni"
```
Misiu, line 30: The grammar here is not correct.
One can say "Zasłoń wszystkie rolety w sypialni" to close all the roller shutters in the sleeping room, but "Przymknij całe rolety w sypialni" is confusing (at least to me).
Personally for me "Przymknij rolety w sypialni" means that I'd like to close (lower) roller shutters to 20% or 10% (almost closed), but I think this use case isn't supported (yet)

tests/pl/fan_HassTurnOn.yaml
```diff
@@ -5,6 +5,7 @@ tests:
       - "Uruchom w salonie wszystkie wentylatory"
       - "Włącz wentylatory w salonie"
       - "Uruchom wentylator w salonie"
+      - "Uruchom całe wentylatory w salonie"
```
Misiu, line 8: Same as roller shutters.
One may use "całe" with lights, but not with fans.
But on the other side, this is just a test case, so we can support this, even if someone won't use that case.

tests/pl/cover_HassCloseCover.yaml
Uriziel01: Yes it was also my feeling but it was decided in one the previous PR to let it through because we focus on flexibility right now. Maybe in different regions of Poland it means something different but at my location using "Przymknij" as "Zamknij" is not common (at least I've never heard it).

This way or another, this was not added in this PR, I'm just adding the tests now ;)

tests/pl/homeassistant_HassTurnOn.yaml
```diff
@@ -10,6 +10,7 @@ tests:
   - sentences:
       - "Zapal lampka nocna"
       - "Włącz lampka nocna"
+      - "Zaświeć lampka nocna"
```
Misiu, line 13: Correct forms are missing "Włącz lampkę nocną" and others.

tests/pl/light_HassTurnOff.yaml
```diff
@@ -3,6 +3,7 @@ tests:
   - sentences:
       - "Zgaś światła w kuchni"
       - "Zgaś wszystkie światła w kuchni"
+      - "Wyłącz całe światła w kuchni"
```
Misiu, line 6: The correct form should be "Wyłącz całe światło w kuchni"

tests/pl/light_HassTurnOn.yaml
```diff
@@ -3,6 +3,7 @@ tests:
   - sentences:
       - "Zapal światło w kuchni"
       - "Zapal wszystkie światła w kuchni"
+      - "Zaświeć całe światła w kuchni"
```
Misiu, line 6: The correct form should be "Zaświeć całe światło w kuchni"

tests/pl/homeassistant_HassTurnOn.yaml
Uriziel01: Will improve, thank you 😃

tests/pl/fan_HassTurnOn.yaml
Misiu: As I mentioned earlier, this is just a test case, maybe someone speaks like this.
Let's leave it :)

tests/pl/light_HassTurnOff.yaml
Uriziel01: Hmm, the whole `Całe` was added in another PR and I see the exact same issue as you do, I think this is in fact grammatically incorrect, converting `światła` to `światło` for this specific form will be somewhat problematic because it applies also for other different forms. I'll give it a go later today and try to find suitable solution, thanks for the feedback.

tests/pl/light_HassTurnOn.yaml
Uriziel01: Hmm, the whole `Całe` was added in another PR and I see the exact same issue as you do, I think this is in fact grammatically incorrect, converting `światła` to `światło` for this specific form will be somewhat problematic because it applies also for other different forms. I'll give it a go later today and try to find suitable solution, thanks for the feedback.

tests/pl/cover_HassCloseCover.yaml
witold-gren: Hey, in my region we use sentence "Przymknij wszystkie rolety w sypialni". 😉 The sentence "Przymknij całe rolety w sypialni" is little weird..

tests/pl/cover_HassOpenCover.yaml
```diff
@@ -27,6 +27,7 @@ tests:
       - "Otwórz rolety w kuchni"
       - "Otwórz wszystkie rolety w kuchni"
       - "Rolety w kuchni otwórz"
+      - "Otwórz całe rolety w kuchni"
```
witold-gren, line 30: For me correct the correct sentence is "Otwórz rolety w kuchni" or "Otwórz wszystkie rolety w kuchni". I never heard "całe rolety" 😀

tests/pl/fan_HassTurnOn.yaml
witold-gren, line 8: again here "całe wentylatory" is not good sentence, I only heard "wszystkie wentylatory"

tests/pl/light_HassTurnOff.yaml
witold-gren: hm.. I'm starting to wonder if the "całe" word fits the context of the things being run.. if we not find any correct sentence, maybe we should remove it? What do you think about it?

witold-gren (changes requested):
Overall good changes, the only thing I have a problem with is the word "całe". For now, I do not see the correct form.. do you think we can remove them?

Uriziel01:
@witold-gren @rrakso @Misiu 
I've looked once again into the whole "całe" situation across the board and I've came to conclusion this should be limited to lights and lights only.
It does not sound right with all other types of instances like covers and fans. And even for lights it would require us to convert all sentences using it to different form (as mentioned by @Misiu in the discussion already).
I'll get rid of it completely for now and create an issue on our Polish issue tracker repository to reevaluate it after the project matures enough.

PR #659, opened 2023-01-15
[MS] Added new for Lights and Fans

zubir2k:
[MS] Added Lights, Fans and Updated common yaml.


zubir2k:
hi @synesthesiam 
Latest pull for [MS] for your consideration. Thank you

PR #660, opened 2023-01-15
Only common-yaml 

poltalashka:
Only common-yaml

### DIFF
--- a/sentences/bn/_common.yaml
+++ b/sentences/bn/_common.yaml
@@ -1,12 +1,12 @@
 language: bn
 responses:
   errors:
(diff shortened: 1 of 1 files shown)


poltalashka:
@synesthesiam, a PR for only common.yaml .... for the HassTurnOn and HassTurnOff the `SENTENCE_MATCHER = vol.Match(
    r"^[\w '\|\(\)\[\]\{\}\<\>]+$",...` does not seem to work with simple BN character set.

synesthesiam:
The `SENTENCE_MATCHER` should be fixed now :+1:

Issue #670, opened 2023-01-16
Validate gives error on vaalid syntax


Robin-St:
@synesthesiam Showed a better way to handle set of min/max brightness in https://github.com/home-assistant/intents/pull/487#discussion_r1071547128. The essence of the problem is that the syntax, as below, uses a colon. 
`
"set <name> brightness to [the] {brightness_min_max:brightness}"`

This works as intended when testing directly with intentfest but the validation catches the colon from the syntax and does not allow the sentence. 

I guess the reg-exp here needs to be updated. Probably let it allow the specific sequence of "{" ? ":" ? "}", would be good to not allow punctuations everywhere. But i do not speak regexp and does not know how to go any further.  https://github.com/home-assistant/intents/blob/8dcb1775910d1351308ca63dd8c53ce24852450b/script/intentfest/validate.py#L114-L117.

synesthesiam:
Fixed in https://github.com/home-assistant/intents/pull/673

PR #671, opened 2023-01-16
Allow colon in sentence matcher

Robin-St:
Allows the precense of a colon in sentences to allow better handling of min/max intents. I dont know if this is enougth to allow the colon or a more complex regexp is wantet that only accepts it between brackets.

See https://github.com/home-assistant/intents/issues/670 for more info.


synesthesiam:
Thanks! Didn't see this and already made the change :stuck_out_tongue:

PR #675, opened 2023-01-17
[SV] Use new max/min syntax

Robin-St:
Updates set light to use the cleaner syntax suggested here: https://github.com/home-assistant/intents/pull/487#discussion_r1071547128

### DIFF
--- a/sentences/sv/_common.yaml
+++ b/sentences/sv/_common.yaml
@@ -33,6 +33,12 @@ lists:
       type: "percentage"
       from: 0
       to: 100
+  brightness_level:
+    values:
+      - in: <maximal>
+        out: 100
+      - in: <minimal>
+        out: 1
   temperature:
     range:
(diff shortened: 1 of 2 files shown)


larsdunemark (approved):
Looks good to me, might be some future work if there is a context aware intent for "minimum" level for a light. Or it might event be that HA can have that as configuration for given lights. So brigthness=1 is actually 25% etc.

Robin-St:
> Looks good to me, might be some future work if there is a context aware intent for "minimum" level for a light. Or it might event be that HA can have that as configuration for given lights. So brigthness=1 is actually 25% etc.

Yes, something like local “senes” for each light that can be defined for min, low, medium, high and highest would be nice. :)

Robin-St:
@larsdunemark Where your intention to merge this? As it looks like you approved but not merged :)

PR #676, opened 2023-01-17
[EN] Use cleaner syntax for min/max brightness

Robin-St:
Uses the cleaner syntax for max/min brightness discussed here: https://github.com/home-assistant/intents/pull/487#discussion_r1071547128


synesthesiam:
Thanks, @Robin-St! I ended up having to add this in manually due to some other edits. Closing this PR, but this will be included :+1:

PR #677, opened 2023-01-17
czech HassLightSet - Min/Max sentenses

schizza:
Add Min/Max sentenses to HassLightSet

### DIFF
--- a/sentences/cs/light_HassLightSet.yaml
+++ b/sentences/cs/light_HassLightSet.yaml
@@ -14,30 +14,30 @@ intents:
 
         #Max/Min sentences
       - sentences:
-          - "(<nastavit>|<zvysit>|<zmenit>) [jas] <area> na [co] (maximum | max | nejvíc)"
-          - "(<nastavit>|<zvysit>|<zmenit>) [na] [co] (maximální | nejvyšší | největší) [jas] <area>"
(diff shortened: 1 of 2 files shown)


sentences/cs/light_HassLightSet.yaml
```diff
@@ -11,8 +11,40 @@ intents:
           - "[jas] <area> (<zvysit>|<ztlumit>|<nastavit>|<zmenit>) [jas] <brightness>"
         slots:
           name: "all"
+
+        #Max/Min sentences
+      - sentences:
+          - "(<nastavit>|<zvysit>|<zmenit>) [jas] <area> na [co] (maximum | max | nejvíc)"
+          - "(<nastavit>|<zvysit>|<zmenit>) [na] [co] (maximální | nejvyšší | největší) [jas] <area>"
+          - "[jas] <area> (<zvysit>|<nastavit>|<zmenit>) [jas] [na] [co] [(nejvyšší|maximální|nejvíc)] [jas] [(maximum | max)]"
```
halecivo, line 19: [co] seems to be related only to nejmíň, don't you think?

halecivo (approved):

sentences/cs/light_HassLightSet.yaml
schizza: yep, good point

PR #686, opened 2023-01-19
Icelandic common

snjallingur:
Made some changes to my previous pull request.
Added common error messages and a list of color sets


synesthesiam:
Thanks!

PR #691, opened 2023-01-20
Add response keys during matching

synesthesiam:
Enables "response" next to "sentences" in intents file. For example:

```yaml
language: "en"
intents:
  HassTurnOn:
    data:
      - sentences:
          - "<turn> on [all] [the] (fan | fans) in <area>"
        slots:
          domain: "fan"
          name: "all"
        response: fans_area
```

which is then defined in `responses/<language>/<intent>.yaml`:

```yaml
language: en
responses:
  intents:
    HassTurnOn:
      default: "Turned on {{ slots.name }}"
      lights_area: "Turned on lights in {{ slots.area }}"
      fans_area: "Turned on fans in {{ slots.area }}"
```

Validates response keys (e.g., `fans_area`) are defined in the correct response YAML file and actually used the correct intent YAML file.
Migrates all "success" response keys to "default".

Response templates are rendered during validation with a fake `state` object that has `name` and `state` attributes. A fake `slots` object is also provided with all slots listed in `intents.yaml`.


responses/ro/HassClimateGetTemperature.yaml
```diff
@@ -2,5 +2,6 @@ language: ro
 responses:
   intents:
     HassClimateGetTemperature:
-      success:
-        - "{{ state.state }} {{ 'grad' if state.state | abs == 1 else 'grade' if state.state == 0 or state.state % 1 != 0 or state.state % 100 | abs < 20 else 'de grade' }}"
+      default: '{{ state.state }} {{ ''grad'' if state.state | abs == 1 else ''grade''
```
Uriziel01, line 5: I understand the conditional statement formatting but why exactly did we migarate here to single quotes? It just makes the code less readable and does not bring much of a value in situation where we still have hundreds of places using double quotes all over the repository.
Does this have other benefits I'm maybe not aware of?

MTrab (approved):
Danish looks good

responses/ro/HassClimateGetTemperature.yaml
synesthesiam: No, this seems to be something PyYAML chose to do during the migration of the response key. We can change it back after the merge; I agree that it's less readable.

synesthesiam:
Going to merge this now, as it's just one step in getting response files ready for translation.

dsimop (approved):
[EL] Greek also looks good.

sentences/en/light_HassLightSet.yaml
```diff
@@ -16,20 +21,31 @@ intents:
           - "(set | change) <area> to <brightness>"
         slots:
           name: "all"
+        response: "brightness_area"
+      - sentences:
+          - "(set | change) <area> to <brightness>"
+        requires_context:
+          domain: "light"
```
makstech, line 28: So, how is this supposed to be working? What does it control/change in the logic? Does it just means that this will set all (set by `slots.name: "all"`) `light` domain devices to `<brightness>`?

makstech (commented):
So, are response keys (e.g. `default`) arrays or just single-option replies? Meaning, there's no variability now?

responses/tr/HassClimateGetTemperature.yaml
```diff
@@ -2,5 +2,4 @@ language: tr
 responses:
   intents:
     HassClimateGetTemperature:
-      success:
-        - "{{ state.state }} derece"
+      default: '{{ state.state }} derece'
```
alpdmrel, line 5: Looks good.

xraver:
IT looks good
thanks

tetele:
@synesthesiam how can we test this? Is there any way to add an expected response to a test?

synesthesiam:
Follow-up PR: https://github.com/home-assistant/intents/pull/698

synesthesiam:
@makstech Just single strings for now, but I do think variability would be a nice option in the future. At this point, I'm trying not to burden translators with thinking they need to do extra work.

sentences/en/light_HassLightSet.yaml
synesthesiam: This means that as the intent matcher is running, if these sentences match the user's text then the domain of the target entity/entities must **also** be light. Otherwise, the match is skipped and the process continues.

It was intended to cases where the same sentence would trigger two intents, so extra info was needed to disambiguate.

PR #696, opened 2023-01-20
[RO] Added responses

tetele:
Added respones as per https://github.com/home-assistant/intents/pull/691


tetele:
Too many conflicts. Starting over

PR #697, opened 2023-01-20
[CS ] czech - add responses

schizza:
default responses added.
Cover changes to sentences files as required due to #691

### DIFF
--- a/responses/cs/HassClimateGetTemperature.yaml
+++ b/responses/cs/HassClimateGetTemperature.yaml
@@ -2,4 +2,4 @@ language: cs
 responses:
   intents:
     HassClimateGetTemperature:
-      default: '{{ state.state }} stupňů'
+      default: '{{ state.state }} {{ ''stupeň'' if state.state == 1 else ''stupně'' if state.state in [2,3,4] else ''stupňů'' }}'
--- a/responses/cs/HassLightSet.yaml
+++ b/responses/cs/HassLightSet.yaml
@@ -1,0 +1,8 @@
+language: cs
+responses:
+  intents:
+    HassLightSet:
+      brightness: "Jas {{ slots.name }} nastaven na {{ slots.brightness }}"
+      brightness_area: "Jas {{ slots.area }} nastaven na {{ slots.brightness }}"
+      color: "Barva {{ slots.name }} nastavena na {{ slots.color }}"
+#      color_area: "Barva {{ slots.area }} nastavena na {{ slots.color }}"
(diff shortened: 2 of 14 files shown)


responses/cs/HassLightSet.yaml
```diff
@@ -0,0 +1,8 @@
+language: cs
+responses:
+  intents:
+    HassLightSet:
+      brightness: "Jas {{ slots.name }} nastaven na {{ slots.brightness }}"
+      brightness_area: "Jas {{ slots.area }} nastaven na {{ slots.brightness }}"
+      color: "Barva {{ slots.name }} nastavena na {{ slots.color }}"
+#      color_area: "Barva {{ slots.area }} nastavena na {{ slots.color }}"
```
halecivo, line 8: can we already add
Barva světel v ...

sentences/cs/cover_HassCloseCover.yaml
```diff
@@ -1,13 +1,15 @@
-language: "cs"
+language: cs
```
halecivo, line 1: this is being changed from cs to "cs" and back already multiple times, is there some final guideline which fotmat we should keep?

responses/cs/HassClimateGetTemperature.yaml
```diff
@@ -2,4 +2,4 @@ language: cs
 responses:
   intents:
     HassClimateGetTemperature:
-      default: '{{ state.state }} stupňů'
+      default: '{{ state.state }} {{ ''stupeň'' if state.state == 1 else ''stupně'' if state.state in [2,3,4] else ''stupňů'' }}'
```
halecivo, line 5: Official language would say also "21 stupeň", but for most of people "21 stupňů" sounds more natural.

For negative values we keep "stupňů" everywhere?

responses/cs/HassLightSet.yaml
schizza: Yes, there will be "Barva v <area>" in next commit. I have prepared it. But then migration come in place.

Good point with negative temp values. 
And I think, nobody will ever use "21 supeň" as it is gramaticly correct, but mosty used for ordinal numbers - 21. stupeň

halecivo:
The PR has quite lot of conflicts with recently merged PR for responses. I checked few of them and they seem to be great, but might be good to check again once you resolve them. 👍

schizza:
Yes I am awared of conflicts, since there was a major update @synesthesiam posted PR with English sentences that is covered in this PR. 

It is closed now and I will post new one to reflect merge update.

PR #698, opened 2023-01-20
Migrate cover intents and add response files

synesthesiam:
This is a big one.

First, `HassOpenCover` is merged into `HassTurnOn` (and `HassCloseCover` into `HassTurnOff`). For many languages, turning on and opening a device are the same thing. So we decided to remove the artificial distinction and merge the intents.

Next, response files for every intent were added to the `responses/<language>` directories. These are just copies of the English responses, and **need to be translated**.

The sentence files contains `response` keys next to their `sentences` now. If `response` is missing, it's assumed to be `"default"`. Different responses for single entity vs. an area actions are provided, for example `cover` vs. `cover_area`.

Finally, response templates are rendered with Jinja2 like in Home Assistant. Templates that affect an entity have access to `state` (with `state.name` and `state.state`). All templates have a `slots` object available with the values of the matched slots (e.g., `slots.name`, `slots.area`, `slots.brightness`).


responses/lv/HassLightSet.yaml
```diff
@@ -0,0 +1,8 @@
+language: lv
+responses:
+  intents:
+    HassLightSet:
+      brightness: '{{ slots.name }} brightness set to {{ slots.brightness }}'
+      brightness_area: Brightness in {{ slots.area }} set to {{ slots.brightness }}
+      color: '{{ slots.name }} color set to {{ slots.color }}'
+      color_area: Color in {{ slots.area }} set to {{ slots.color }}
```
makstech, line 8: So, what standard do we use? Do we use single quotes, double quotes, or no quotes at all? 😄

Ahmed-farag36 (approved):

responses/lv/HassLightSet.yaml
synesthesiam: Anyone know how to force PyYAML to always export with double quotes? I'd prefer that, but haven't found a way to do it let :/

responses/cs/HassClimateSetTemperature.yaml
```diff
@@ -0,0 +1,5 @@
+language: cs
+responses:
+  intents:
+    HassClimateSetTemperature:
+      default: Temperature set to {{ slots.temperature }} degrees
```
schizza, line 5: default: 'Teplota nastavena na {{ slots.temperature }} {{ ''stupeň'' if state.state == 1 else ''stupně'' if state.state in [2,3,4] else ''stupňů'' }}'

responses/cs/HassLightSet.yaml
```diff
@@ -0,0 +1,8 @@
+language: cs
+responses:
+  intents:
+    HassLightSet:
+      brightness: '{{ slots.name }} brightness set to {{ slots.brightness }}'
+      brightness_area: Brightness in {{ slots.area }} set to {{ slots.brightness }}
+      color: '{{ slots.name }} color set to {{ slots.color }}'
+      color_area: Color in {{ slots.area }} set to {{ slots.color }}
```
schizza, line 8: brightness: "Jas {{ slots.name }} nastaven na {{ slots.brightness }}"
      brightness_area: "Jas {{ slots.area }} nastaven na {{ slots.brightness }}"
      color: "Barva {{ slots.name }} nastavena na {{ slots.color }}"

responses/cs/HassTurnOff.yaml
```diff
@@ -0,0 +1,9 @@
+language: cs
+responses:
+  intents:
+    HassTurnOff:
+      default: Turned off {{ slots.name }}
+      lights_area: Turned off lights in {{ slots.name }}
+      fans_area: Turned off fans in {{ slots.name }}
```
schizza, line 7: default: "{{ slots.name }} vypnuto"
      lights_area: "Světla v {{ slots.name }} byla zhasnuta"
      fans_area: "Větráky v {{ slots.name }} byly vypnuty"
      fan: "Větrák {{ slots.name }} vypnut"
      light: "Světlo {{ slots.name }} bylo vypnuto"

responses/cs/HassTurnOff.yaml
```diff
@@ -0,0 +1,9 @@
...
+      cover: Closed {{ slots.name }}
+      cover_area: Closed {{ slots.area }}
```
schizza, line 9: cover: "{{ slots.name }} zavřeno"
      cover_area: "{{ slots.area }} zavřeno"
      garage: "{{ slots.name }} zavřena."

tests/test_language_intents.py
```diff
@@ -76,16 +76,17 @@ def do_test_language_sentences(
...
-            # sentences for the light domain.
-            if intent_schemas[file_intent]["domain"] == file_domain:
-                assert (
-                    "domain" not in data.slots
-                ), f"File {file_name} should only have sentences without a domain slot"
-            else:
-                assert (
-                    data.slots.get("domain") == file_domain
-                ), f"File {file_name} should only have sentences with a domain slot set to {file_domain}"
+            if file_domain != "homeassistant":
```
balloob, line 79: I'm surprised this is needed, as that is what we checked more-or-less with 

```
if intent_schemas[file_intent]["domain"] == file_domain:
```

It checks if the integration that defines the intent is equal to the current file's domain. It's not clear to me why this is no longer valid for `homeassistant` ?

balloob: Especially since `homeassistant` is the main driver for this check to begin with. Other intents are not tested in other domains than the one that defines them.

balloob:
@schizza please open a new PR with the changes once this PR has been merged. This PR is too big to accept individual contributions.

PR #710, opened 2023-01-21
Adding latvian (lv) synonyms and inflections

raivisdejus:
@makstech Please see some synonyms and inflections for different sentence variations.


makstech:
@raivisdejus Thanks! Would you consider adding a couple of tests for these variations in another PR?

PR #711, opened 2023-01-22
Some small additions.

styroxfjant:
Test sentence was missing for fan_HassTurnOff, so I had to add that also.


sentences/sv/_common.yaml
```diff
@@ -59,10 +59,10 @@ expansion_rules:
   är: "(är det | är)"
   brightness: "{brightness} [percent]"
   slå_på: "(tänd | sätt (på | igång | fart på) | slå (på | till) | starta)"
-  slå_av: "(slå (av | från | ifrån) | släck | stäng av | stoppa)"
+  slå_av: "(slå (av | från | ifrån) | släck | stäng (av) | stoppa)"
```
larsdunemark, line 62: Should be "stäng [av]" to make "av" optional.

sentences/sv/_common.yaml
```diff
@@ -59,10 +59,10 @@ expansion_rules:
...
   i_på: "(i | på | vid | bredvid)"
   ändra: "(ställ in | sätt | justera | ändra | vrid (upp | ner))"
-  temp: "(temperatur)"
+  temp: "(temperatur[en])"
```
larsdunemark, line 65: I don't think this is used, in that case we should add a test case. It's a good idea to separate different intents/changes into separate PR.

larsdunemark (changes requested):

sentences/sv/_common.yaml
styroxfjant: You are correct.

sentences/sv/_common.yaml
styroxfjant: I'll remove this for now then. If seen needed we must add it + the test cases later.

larsdunemark (approved):

PR #744, opened 2023-01-24
Add french responses

piitaya:


synesthesiam:
Thanks!

benjaminlecouteux (commented):
I see a problem with the "de"; which could turn into "du" or "de la" depending on the piece.

piitaya:
Yes, we have the same issue with device name too because every word in french has a gender 🥲

benjaminlecouteux:
Wouldn't it be cleaner to delete this "de" and replace it with a stop list?

piitaya:
Can you provide some examples? 😀

benjaminlecouteux:
I was thinking of something like :
Luminosité de {{ slots.area }}
Luminosité {{ list.determinant[slots.area] }} {{ slots.area }}
with {{ list.determinant }} which contain {"de la", "du", "dans la"} selected according the gender of slots.area

piitaya:
The problem is that `area` can be every word and there is no gender on Home Assistant entity and area 😔
I tested with Siri and I think they have the gender of every word... 😅

benjaminlecouteux:
I think there is no choice to link gender with different words. Would it be possible to create an external dictionary linking each word to its gender? I don't see any other solution.

benjaminlecouteux:
It should be possible to automate this with a dictionary such as https://fr.wiktionary.org/

PR #750, opened 2023-01-24
Translate existing Swedish responses

MartinHjelmare:


balloob:
Going to merge this as we're focusing on gettin some responses in. We can improve it later.

PR #752, opened 2023-01-24
Add Danish responses

balloob:


balloob:
Going to merge this as we're focusing on gettin some responses in. We can improve it later.

fadamsen:
The responses are fine, actually, so I see no immediate need to improve them.

PR #759, opened 2023-01-24
[RO] Added context and response testing

tetele:
Added context for sentences (e.g. "turn on the light" said in the kitchen).

Added response tests

This PR depends on https://github.com/home-assistant/intents/pull/756 as it fails tests without it

Known issues to be solved even after #756:
- context information is not treated as slots and populated into responses
- state context is unavailable for climate_HassClimateGetTemperature - see https://discord.com/channels/330944238910963714/1054846248374505572/1067557531590856734


tetele:
God, I hate merges

PR #774, opened 2023-01-26
BG add min and max brightnes and simplify code

BossNeo:


hristo-atanasov:
Moving min and max light brightness in `_common.yaml` as `min` and `max` only in the context of light brightness (in bulgarian as feminine gender) might be unappropriate for later reuse of `min` and `max` for other units of measurment.

BossNeo:
@hristo-atanasov Dear Language leader i am tired to argue with you of all my PR for the things that i did right. From my first PR you start to argue that български must be with small letter , even is not in the middle in sentence. Ignoring all my words that is a table and every row must start with capital letter. 
Now for " use dashes after по and най" i know in Bulgarian language  is  must be write най- and по- , but we dont do translation , we try to make things work. For that sentences must be free from all coma dots and etc . `,.- Because that can mess up with machine processing letter.  You can check how speech to text engine work here https://azure.microsoft.com/en-us/products/cognitive-services/speech-to-text/#features . In most optimistic way assist in HA will do the same and will skip the dashes in the sentence.   For that i didnt put dashes , and hope you will correct your action.
For "By moving too many translations as variables in _common.yaml you're making translations and reading them unmanageable" is more unmanageable if have all this words in sentences ,and  is the prefer way to do it  from my chats in discord, is more easy to add words , more easy to edit words. No need to search every line when this words are used or no. Its only your prefer way and you push it her. You cut out the native sentence of "промени осветеността в кухнята на 50% " and now is not working.  I dont see why you must cut things,  just to dont be complicated for you. Letter maybe we will need to clean some words , and all words to be in variables it be a plus.   Context of the use of varibles is more clear for people who dosent know Bulgarian  when you have :
<set> <light_brightness> <area> { brightness}  is oblivious what you try to do.  
then you need only the word:
light_brightness : "[яркост[та] [на светене] | светлина[та]] "
 area: "[в | на| във] {area}"
 set: "направи | сложи | постави | смени | промени | настрой | задай | нагласи | тури" 
 i just cannot image how for you this :
 (направи | сложи | постави | смени | промени | настрой | задай | нагласи | тури) [яркост[та] [на светене] | светлина[та]] [в | на| във] {area}"  [на | до] {brightness} is more simple, more clear, and more easy for some who docent know Bulgarian to know  what you try to do ? Please tell me ?
Have so much more work to be done : there is no min max for covers , there is no responses sets to lights , fans and etc.  There is  no  requires_context  to skip conflicts when user do not  provide all the words and more.

PR #778, opened 2023-01-26
Update name list

horvathgergo:
names for air purifiers are missing


synesthesiam:
@nagyrobi can you take a look at this?

PR #791, opened 2023-01-27
Initial commit to add Swiss German.

c0ffeeca7:
This adds a minimal set of sentences in Swiss German.
The variant in this set does not correspond to a particular dialect, but is a blend of different influences.

- Swiss German has vast variety of dialects that differ significantly from German.
- There is no standardized way to write Swiss German. There is no prescriptive, written grammar, and there are little rules on how to spell Swiss German.
- Swiss German is spoken in many dialects, but written mostly in informal context.
- In most cases, German is used for writing. 

- -------------------------------------
-  Because there are so many different dialects and there is very little standardization in writing, using Swiss German as an input option will be challenging.
- The transcriber engines I know of transcribe Swiss German into German, not into some form of Swiss German spelling.
- It seems Swiss German is not (yet) listed in [commonvoice](https://commonvoice.mozilla.org/en)
- I still think it would be cool to be able to speak to HA in Swiss German.
---------------------------------------------------

More information about the language:
[wikipedia](https://en.wikipedia.org/wiki/Swiss_German)
[Atlas of the World's Languages](https://books.google.ch/books?id=R0xWDwAAQBAJ&pg=PR309&redir_esc=y#v=onepage&q&f=false)
[Idiotikon](https://www.idiotikon.ch/)

Speech to Text Transcription:
https://xn--tggl-5qa.ch/about
https://stt4sg.fhnw.ch/long

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -28,6 +28,8 @@ de:
   leaders:
     - Scorpoon
     - easterapps
+de-CH:
+  nativeName: Swiss-German
 el:
   nativeName: Ελληνικά
   leaders:
--- a/responses/de-CH/HassClimateGetTemperature.yaml
+++ b/responses/de-CH/HassClimateGetTemperature.yaml
@@ -1,0 +1,5 @@
+language: de-CH
+responses:
+  intents:
+    HassClimateGetTemperature:
+      default: '{{ state.state }} degrees'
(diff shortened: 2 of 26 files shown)


synesthesiam:
Awesome, thank you!

PR #793, opened 2023-01-27
Update _common.yaml

yousaf465:
temperature urdu

### DIFF
--- a/sentences/ur/_common.yaml
+++ b/sentences/ur/_common.yaml
@@ -83,7 +83,7 @@ lists:
       to: 100
   temperature:
     range:
-      type: "temperature"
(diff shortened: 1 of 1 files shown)


sentences/ur/_common.yaml
```diff
@@ -83,7 +83,7 @@ lists:
       to: 100
   temperature:
     range:
-      type: "temperature"
+      type: "temperature" "درجہ حرارت"
```
AalianKhan, line 86: the `type` is used by home assistant internally.
 `درجہ حرارت`
is not a valid type

AalianKhan (changes requested):

PR #795, opened 2023-01-27
Add Remote Domain and sentances, responses and tests

tombo12354:


tombo12354:
Replacing with new version

PR #802, opened 2023-01-28
de-CH: add some first sentences to turn on lights

c0ffeeca7:


dontinelli:
@c0ffeeca7 : I made some amendments in #896. Would be great if you could have a look at it, as it is my first contribution.

PR #806, opened 2023-01-28
[de] Extend German Sentences and Tests for the Light Domain

Johnson145:
I focussed on extending the content of the light domain for the German language. A few generic changes done in the `_common.yaml` file may affect other domains, too.


easterapps (approved):
looks good to me 
Thank you for your contribution!

PR #810, opened 2023-01-29
Add Remote Domain and sentances, responses and tests

tombo12354:


sentences/en/_common.yaml
```diff
@@ -48,7 +48,7 @@ expansion_rules:
   area: "[the] {area}"
   what_is: "(what's | whats | what is)"
   brightness: "{brightness} [percent]"
-  turn: "(turn | switch | change)"
+  turn: "(turn | switch | change | shut)"
```
tetele, line 51: "shut" is only used for shutting something off, not on. Maybe it's better to add it in the HassTurnOff intent (e.g. `(<turn> | shut) off ...`).

tombo12354: That's a good point, I agree that makes more sense.

tombo12354: Updated

tombo12354:
Not sure I understand what's failing on the tests. Reading through it, it seems to be failing in that it cannot find remote files for other languages. Is there somewhere else to update?

tetele:
@tombo12354 would you care to update this PR or should we close it? I'm not sure remotes are used only for TVs, though.

PR #812, opened 2023-01-30
[BG] add change color to area and other minior fix

BossNeo:
1. adding response for turn on/off lights
2. remove the "-" in по and най because speech to text engine dosent give ,.-
3. adding response to fans area
4. adding  requires_context: for covers and lights with different response to them
5. merge light_brightness names in other sentences to be more clear
6. adding set color for area functionality
7. .adding sentences for new set color area


hristo-atanasov:
I find Your actions and speaking very offensive. For a person that barely knows his own native language (grammar and punctuation) and just discovered github, I no longer want to continue working together with You. We are all here with a mission to help each other and the entire HA society. As contributors we have to work as a team and we must make the best decisions and consider/respect the opinion of the others. This is no longer possible with You and Your "intents". With the amount of time spending explaining You how github works and about bulgarian grammar I would translate things by myself. Also another bulgarian translators too (they stopped contributing, because of Your way doing things, dissrespecting other people's oppionion). As I still find You as a good contributor, I would suggest you contacting me and suggesting the changes/improvements You want in pm's. As from the way you are doing things no one from the other translators, including me, wants to be involved in this mission. Or may be continue doing PR's, but wheather it will be approoved or not will be decided at the moment. I decided to to leave You as a sideplayer and just rely on Your ideas, instead of leaving You do all the translations in a manner that only You thing is in the right way. If you want You can still make PRs, but I'll decide what part of it is usefull and can be used, and what can not. Also part of your PR's can be used in some for the translations from me and/or any of the other contributors. Thank you for all of Your contributions till now! I hope you continue helping improving the bulgarian language translation and intents further. Best wishes!

PR #822, opened 2023-01-31
Created "conversation" module to simulate conversations.

rrakso:
)) Changed the "parse sentence" function to a class, because now there are several places that use it.
)) Updated the README file.


script/intentfest/conversation.py
```diff
@@ -0,0 +1,145 @@
...
+    return parser.parse_args()
+
+
+class InstructionsI18n:
+    def __init__(self, languageKey="en") -> None:
+        self.fallbackLanguage = "en"
+
+        # Feel free to add texts in your language ;)
+        self.instructionsI18n = {
+            "en": {
```
balloob, line 38: translations should not live inside the python code. create a new folder under intentfest.

script/intentfest/conversation.py
```diff
@@ -0,0 +1,145 @@
...
+                    print("------------------------------")
+
+
+def run() -> int:
+    args = get_arguments()
+    conversation = Conversation(args.language, args.debug_response)
+    conversation.do()
+    # Liter said 'Missing return statement',
+    # but this particular return is unreachable.
+    # That is why it is here.
```
balloob, line 144: If it doesn't come here, you didn't add a mode to quit the conversation :)

script/intentfest/sentenceParser.py
```diff
@@ -0,0 +1,75 @@
...
+                result.response
+            )
+            self.output_dict["response"] = normalize_whitespace(
+                render_response(response_template, result)
+            ).strip()
+
+            return True
+        return False
+
+    def getResponseForHuman(self):
```
balloob, line 71: Seems weird that we would have a non-debug mode. This is a test script afterall

script/intentfest/sentenceParser.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, language) -> None:
+        self.slot_lists = None
+        self.intents = None
+        self.responses = None
+
+        self.output_dict = None
+
+        self.prepareNewIntents(language)
+
+    def prepareNewIntents(self, language):
```
balloob, line 28: Use snake case for functions

script/intentfest/sentenceParser.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Load test areas and entities for language
+        test_names = yaml.safe_load((tests_dir / "_fixtures.yaml").read_text())
+        self.slot_lists = get_slot_lists(test_names)
+
+        # Load intents
+        intents_dict: Dict[str, Any] = {}
+        for intent_path in language_dir.glob("*.yaml"):
+            with open(intent_path, "r", encoding="utf-8") as intent_file:
+                merge_dict(intents_dict, yaml.safe_load(intent_file))
```
balloob, line 40: a) these are sentences not intents
b) we have a util function for this `load_merged_sentences`

script/intentfest/sentenceParser.py
```diff
@@ -0,0 +1,75 @@
...
+        self.slot_lists = None
+        self.intents = None
+        self.responses = None
+
+        self.output_dict = None
+
+        self.prepareNewIntents(language)
+
+    def prepareNewIntents(self, language):
+        language_dir = SENTENCE_DIR / language
```
balloob, line 29: Please stick to the correct terminology. This is sentences

script/intentfest/conversation.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Toggle debug
+        if userSentence.lower() == "d":
+            self.localDebug = not self.localDebug
+            print(
+                f"{self.instructionsI18n.get('DEBUG')}: {self.instructionsI18n.get('DEBUG_ON') if self.localDebug else self.instructionsI18n.get('DEBUG_OFF')}"
+            )
+
+        # Do intents reload
+        elif userSentence.lower() == "r":
```
balloob, line 98: Instead of supporting reload, we should add [`watchdog`](https://pypi.org/project/watchdog/) to automatically reload + reparse last sentence.

script/intentfest/conversation.py
balloob: but also, we can just keep it in English for now? Otherwise we need to add validation for intentfest translations etc …

script/intentfest/conversation.py
```diff
@@ -0,0 +1,145 @@
...
+        return languageSet[key]
+
+
+class Conversation:
+    def __init__(self, language, debug=False) -> None:
+        self.language = language
+        self.localDebug = False if debug is None or debug is False else bool(int(debug))
+
+        self.sentenceParser = SentenceParser(self.language)
+        self.instructionsI18n = InstructionsI18n(self.language)
```
balloob, line 82: snake case

script/intentfest/conversation.py
rrakso: OK! 😄 Thanks

script/intentfest/sentenceParser.py
rrakso: Yes, but it was by design that I didn't want the debug to be visible all the time. I wanted to be able to ask questions and check the answers straight away. And if at some point I need more info, I just type the letter `d` ;)

Do you think debug mode should be on all the time?

script/intentfest/sentenceParser.py
rrakso: Ok, will use the `load_merged_sentences` function, it's written that way because I relied on the code available here [script/intentfest/parse.py](#diff-ab0ef75141f7639218d09a8b8442c9e33ae6063bde7c5f80d63c0e94c4260c98.)

script/intentfest/conversation.py
rrakso: Wow! Good point! I use [nodemon](https://www.npmjs.com/package/nodemon) all the time, but I hadn't thought of that!

README.md
```diff
@@ -117,6 +117,60 @@ You can add lists, ranges, and expansion rules as well:
...
+>>> Turned on lights in living room sir.
+<<<
+```
+
+You can type **`r`** to reload new responses and / or sentences, and **`a`** (or `arrow up` - something like in terminal using commands history).
+
+There is also an option to see all the response data, just type **`d`** to toggle debug mode, or call the `conversation` module with `--sentence-debug 1` to turn it on immediately:
+
+```
+python3 -m script.intentfest conversation --language en --debug-response 1
```
tetele, line 149: Why not just `--debug-response`? If the parameter included in the args, show debug info. If its' omitted, don't.

script/intentfest/sentenceParser.py
schizza: @rrakso debug mode might be set to true by default and there should be an option to turn in off?

balloob: We should just remove the option and always enable it. This is a debug script after all.

rrakso: I like @schizza's suggestion 😄 but @balloob is also right. I can change any way you want guys. What's the decision @balloob @schizza?

schizza: @rrakso as we need a debug it should be turned on always. If I don't want to debug I will use an Assist in HA :D

README.md
tetele: My suggestion is, regardless of whether you want to add/remove debug functionality at runtime, drop the "1". Just check whether the `--debug-response` parameter is present, not if it has a certain value.

balloob:
Please do not force-push. It makes it impossible for me to see what has changed compared to my previous review.

balloob:
Ah that's fine, I thought you had included fixes with it 👍

README.md
rrakso: Yeah, funny thing 😅 That was my first thought. But I got this error  `__main__.py: error: argument --debug-response: expected one argument`, so that was the reason why I check the value of this parameter 😆. BUT NOW IT IS AS IT SHOULD BE! 😁

script/intentfest/conversation.py
rrakso: OK!

script/intentfest/conversation.py
rrakso: The script now listens for file changes! Thanks!

script/intentfest/conversation.py
rrakso: Now we can end the conversation by typing "q" or pressing CTRL+C 😉

script/intentfest/sentenceParser.py
rrakso: OK, thanks!

script/intentfest/sentenceParser.py
rrakso: Now I'm using `load_merged_sentences`, thanks!

script/intentfest/sentenceParser.py
rrakso: Ok

rrakso:
@tetele @balloob @schizza I have added your suggestions to the code. Please review them 😃

PS
This script is in debug mode by default, but you can disable it for yourself, [here](https://github.com/home-assistant/intents/pull/822/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R171) is very briefly described how to do it. @balloob can we do it like that? Or should I remove this functionality completely?

script/intentfest/conversation.py
```diff
@@ -0,0 +1,159 @@
...
+            self.running = True
+            while time() - self.last_modification_time <= 0.3:
+                sleep(0.1)
+            self.running = False
+
+            self.callback_files_finished_changing()
+
+    def on_any_event(self, event):
+        self.last_modification_time = time()
+        Thread(target=self.was_change).start()
```
balloob, line 58: This will spawn a Thread on any event, why?

script/intentfest/conversation.py
```diff
@@ -0,0 +1,159 @@
...
+        self.last_modification_time = None
+        self.running = False
+        self.callback_files_finished_changing = callback_files_finished_changing
+
+    def was_change(self):
+        if not self.running:
+            self.running = True
+            while time() - self.last_modification_time <= 0.3:
+                sleep(0.1)
+            self.running = False
```
balloob, line 52: Use [`Timer`](https://docs.python.org/3/library/threading.html#threading.Timer). Cancel and create a new timer on every incoming event.

script/intentfest/conversation.py
```diff
@@ -0,0 +1,159 @@
...
+        print(
+            "Type your query here, or type 'a' (or 'arrow up' on the keyboard) to repeat it. Type 'q' (or CTRL+C) to end the conversation.\n"
+        )
+
+        while self.have_a_conversation:
+            try:
+                user_sentence = self.ask_user()
+            except KeyboardInterrupt:
+                print()
+                break
```
balloob, line 108: For completion.

```suggestion
                self.have_a_conversation = False
                break
```

frenck:

../Frenck

PR #830, opened 2023-02-01
Upgrade to Hassil 1.0 with new parser

synesthesiam:
This is a big change!

There is a new template parser that is much faster and a bit more strict. I've copied the release notes from [hassil](https://github.com/home-assistant/hassil/releases/tag/v1.0.0) below.

The most import change is that `(a|b)` and `( a | b )` are **no longer the same**.  Whitespace around `|` is now used by the parser. I've done my best to fix existing YAML files, but I've only done enough to ensure the tests pass!

Another difference is that characters like `%` and `°` need to be put inside templates now. They were previously skipped by the parser, but now only punctuation is skipped (see below for details).

---

# HassIL 1.0.0 Release Notes

## Breaking Changes

- Whitespace is now significant around the `|` character. Previously, `light(s | ing)` would match both "lights" and "lighting", but now it will match "lights" and **"light ing"**
- Only punctuation is skipped automatically, instead of all "non-word" characters (regex `\W`). Templates must now include `%` and `°` explicitly. The current set of punctuation characters is `.。,，?¿？!！;；:：`
- Template parsing is more strict, so unmatched `(` or `]` characters will cause parsing errors

## New Features

Intents YAML has a new settings block:

```yaml
language: "..."
settings:
  ignore_whitespace: false
```

The only option is `ignore_whitespace`, which is intended for language families like `zh` (Chinese) that may be written without spaces. The default value is `False`.


halecivo (commented):
@synesthesiam In lang cs, sentences for fan domain also has spaces around | , which should be removed. I can do that in CET timezone evening, if you need help.

HepoH3 (commented):
what about `[]` or `in/out` syntax? Should spaces removed there too?
Moreover, if I understand correctly, at least in [ru/_common.yaml](https://github.com/home-assistant/intents/blob/hassil-1.0/sentences/ru/_common.yaml) there were no new changes, I can do it in twelwe hours.

TheFes (approved):
I approve the changes to the NL files

LubosKadasi:
SK files spaces and units fixed, 
Thank You

LubosKadasi (approved):

TheFes:
> what about `[]` or `in/out` syntax? Should spaces around `|` to be removed there too? Moreover, if I understand correctly, at least in [ru/_common.yaml](https://github.com/home-assistant/intents/blob/hassil-1.0/sentences/ru/_common.yaml) there were no fixes to new syntax, I can do it in twelwe hours.

I wonder about this as well, I see they are left in in the changed files.

Uriziel01 (approved):

flexy2dd (approved):

zubir2k (approved):
Reviewed and approved for [ms]

alpdmrel (approved):

tetele:
Did anyone else notice the lack of spaces in sample sentences?
```

**Review by MTrab — APPROVED** [review]

**Review by benjaminlecouteux — APPROVED** [review]

**Comment by schizza** [comment]
Pushed fixtures and made PR to comply your changes in this branch. Otherwise it is approved.

**Review by schizza — APPROVED** [review]

**Review by auanasgheps — APPROVED** [review]

**Review by hertzg — APPROVED** [review]
:shipit:

**Review by skynetua — APPROVED** [review]

File: sentences/sv/_common.yaml [hunk 1]
```diff
@@ -53,28 +53,28 @@ lists:
...
+  brightness: "{brightness}[%| percent]"
   slå_på: "(tänd | sätt (på | igång | fart på) | slå (på | till) | starta)"
   slå_av: "(slå (av | från | ifrån) | släck | stäng [av] | stoppa)"
   i_på: "(i | på | vid | bredvid)"
   ändra: "(ställ in | sätt | justera | ändra | vrid (upp | ner))"
   temp: "(temperatur)"
   alla: "(all | alla | allt)"
   ljusintensitet: "(ljusstyrka | intensitet | styrka | nivå | effekt)"
-  ljuskällor: "(ljuskällor[na] | ljus[en | et] | [glöd|led|diod]lamp[ a | an | or | orna] | lysrör[en| et] | belysning[en])"
+  ljuskällor: "(ljuskällor[na] | ljus[en | et] | [glöd|led|diod]lamp[a|an|or|orna] | lysrör[en|et] | belysning[en])"
```
larsdunemark, line 68: ljus[en|et]

sentences/sv/_common.yaml
```diff
@@ -53,28 +53,28 @@ lists:
...
   dimra: "(ändra | sätt | vrid | dra [på] | skruva [på] | dimra | dimma | ställ [in] | höj | sänk)"
   möjliga: "(möjlig[a] | tillgänglig[a])"
   maximal: "(maximal | max[imum] | högsta | starkaste | intensivaste | full[t])"
   minimal: "(minimal | minimum | minsta | lägsta | svagaste)"
-  gardiner: "([rull | motor]gardin[er | erna | en] | persienn[er | erna | en] | spjälgardin[er | erna | en] | draperi[er | erna | en]) | förhänge[n | na]"
+  gardiner: "([rull | motor]gardin[er|erna|en] | persienn[er|erna|en] | spjälgardin[er|erna|en] | draperi[er|erna|en]) | förhänge[n|na]"
```
larsdunemark, line 73: I guess that `[rull |` should be changed to `[rull|` to match rullgardin

larsdunemark (commented):
I guess that we are missing some test cases for Swedish covering some cases. I think that we can fix this in upcoming PR.

script/intentfest/validate.py
```diff
@@ -145,6 +145,7 @@ def validate(value):
 SENTENCE_COMMON_SCHEMA = vol.Schema(
     {
         vol.Required("language"): str,
+        vol.Optional("settings"): {str: match_anything},
```
balloob, line 148: We should match the strings that we support

balloob (approved):

synesthesiam:
> what about `[]` or `in/out` syntax? Should spaces around `|` to be removed there too? Moreover, if I understand correctly, at least in [ru/_common.yaml](https://github.com/home-assistant/intents/blob/hassil-1.0/sentences/ru/_common.yaml) there were no fixes to new syntax, I can do it in twelwe hours.

Yes, these should technically be removed as well. When it's around whole words, it won't cause a problem because it just means extra whitespace around a word. But it's good practice to keep it clean, I think :slightly_smiling_face:

PR #836, opened 2023-02-01
Update fan_HassTurnOff.yaml

OliverSkivdal:


schizza (changes requested):

LaStrada:
This is not needed anymore, except that this PR includes `vifta`, but this should be added to `_common.yaml` (`expansion_rules: vifte`).

Closing this now, since the change is needed in a different file. Either update this PR with the change in `_common.yaml` or create a new PR.

PR #838, opened 2023-02-01

chuk-a:


schizza (changes requested):
I need to translate sentences and then make appropriate test for the file in `test/mn/homeassistant_HassTurnOn.yaml` otherwise CL tests will fail.

And another suggestion: run local tests before PR. While local test fill fail it is non sense to require PR before all is clean.

frenck:

../Frenck

PR #843, opened 2023-02-01
[NL] Add fixed domain after name

TheFes:
Added fixed domain expansion rules in the domain specific files to make sure the right intent is used in case the domain word is added to the intent, but not added in the name of the device.


TheFes:
I initially added tests without a space.
So `<name> <light>` was tested both with eg `werkbanklicht` and `werkbank licht`
But I expect this to fail after https://github.com/home-assistant/intents/pull/830 is merged, so I will try to implement that after that one is merged.

synesthesiam:
#830 has been merged :+1:

sentences/nl/fan_HassTurnOff.yaml
```diff
@@ -5,13 +5,17 @@ intents:
...
-          - "<doe> (alle | in) <area> <ventilator> [<naar>] uit"
+          - "<doe> (alle|in) <area> <ventilator> [<naar>] uit"
         response: fans_area
         slots:
           domain: "fan"
           name: "all"
       - sentences:
-          - "<doe> (overal | alle) <ventilator> uit"
+          - "<doe> <name>[' ']<ventilator> [<naar>] uit"
+        slots:
```
balloob, line 15: Gebruik requires_context zodat je alleen names gelinkt aan fan domain matched.

```suggestion
        requires_context:
          domain: "fan"
        slots:
```

balloob (approved):

PR #856, opened 2023-02-02
Add Brazilian(?) variations for current sentences

igorsantos07:
Not exactly sure if those are strictly Brazilian sentences, but they're definitely stuff I would hear around here :)

### DIFF
--- a/sentences/pt/_common.yaml
+++ b/sentences/pt/_common.yaml
@@ -24,6 +24,8 @@ lists:
         out: "red"
       - in: "laranja"
         out: "orange"
+      - in: "abóbora"
+        out: "orange"
       - in: "amarelo"
         out: "yellow"
       - in: "amarela"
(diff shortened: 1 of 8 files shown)


leozanotto:
Probably you commitd on wrong files, the correct ones should be under pt-br folders not pt

igorsantos07:
I actually committed when pt-br didn't even exist and I wasn't sure it
would 🤷🏼‍♂️

Igor Santos
-- Desenvolvedor Web
[enviado do meu celular]

leozanotto:
> I actually committed when pt-br didn't even exist and I wasn't sure it would 🤷🏼‍♂️ but I guess I wasn't as fast/well connected as those who created the folder and didn't notice other Brazilian PRs.

I have created a PR right now, probably some of the changes you made are already there: #1146

dclobato:
> I have created a PR right now, probably some of the changes you made are already there: #1146

Same regarding the PR #1111...

PR #858, opened 2023-02-02
Add definite form of "temperatur"

ehn:


larsdunemark:
Looks good, could you maybe add a test sentence for this change so we don't miss it in the future.
Guess it could be like: `var är temperaturen i köket?`

ehn:
> Looks good, could you maybe add a test sentence for this change so we don't miss it in the future. Guess it could be like: `var är temperaturen i köket?`

Sure, is there documentation for how to do this correctly? Should I create a new PR for this?

Robin-St:
This is the relevant file to add tests to:
https://github.com/home-assistant/intents/blob/main/tests/sv/climate_HassClimateGetTemperature.yaml
And this is the documentation for tests:
https://developers.home-assistant.io/docs/voice/intent-recognition/test-syntax/

No need for a new PR, just make the changes to the branch and push them and they will be visible here :)

ehn:
Tests added.

tests/sv/climate_HassClimateGetTemperature.yaml
```diff
@@ -3,11 +3,13 @@ tests:
   - sentences:
       - "vad är det för temperatur?"
       - "vilken temperatur är det?"
+      - "vad är temperaturen?"
     intent:
       name: HassClimateGetTemperature
   - sentences:
       - "vad är det för temperatur i vardagsrummet?"
       - "vilken temperatur är det i vardagsrummet?"
+      - "vad är temperaturen i köket?"
```
larsdunemark, line 12: So this doesn't work, the test case is expecting to trigger:
```
   intent:
      name: HassClimateGetTemperature
      slots:
        area: "livingroom"
```
As stated below. But the sentence will trigger `area: "kitchen"` so the testcase fails right now.
So ether change `Köket` to `vardagsrummet` or create a new test case expecting the area of `kitchen`

larsdunemark (changes requested):
Need to pass the testcase

tests/sv/climate_HassClimateGetTemperature.yaml
ehn: Mea culpa. 🤦‍♂️

larsdunemark (approved):
Looks good to me

PR #861, opened 2023-02-02
[EN] Locate person

peter-dolkens:
Add a query to locate individuals


responses/en/HassPersonGetLocation.yaml
```diff
@@ -0,0 +1,5 @@
+language: en
+responses:
+  intents:
+    HassPersonGetLocation:
+      location: "{slots.name} is in the {slots.state}"
```
teleksterling, line 5: This response needs more thought. It's works grammatically for your room examples, but not for zones which I would assume is the more common use case. 

- is in the away
- is in the home
- is in the Dads house
- is in the Sydney
- is in the dancing

Perhaps attributes of the state object could be used in a more complex template to distinguish between cases? 
- is away
- is at Dad's house
- is in the kitchen
- is driving in (suburb) ??

Or a generic response
- (person)s location is state.state

sentences/en/person_HassPersonGetLocation.yaml
```diff
@@ -0,0 +1,11 @@
+language: "en"
+intents:
+  HassPersonGetLocation:
+    data:
+      - sentences:
+          - "<where_is> {person}"
+          - "have you seen {person}"
+          - "<where_did> {person} go"
```
teleksterling, line 8: go [to]

teleksterling (commented):
This is a great start. 👍🏼

balloob:
Thanks for your contribution. However, this intent does not exist. Before we start adding sentences, we need to first have a discussion on what it should be named, what inputs and outputs it should contain etc. Once approved, we can start gathering sentences and adding it to Home Assistant. Going to close this now as we should start at the start, not the end.

teleksterling:
> we need to first have a discussion on what it should be named, what inputs and outputs it should contain etc. 

Thanks for the detailed response. What you say makes sense! 

Is there discussion or documentation on available/ planned intents anywhere (public)?
I've been looking around for a couple weeks and all I seem to come across is this detailed level stuff about the sentences and Hassil.

Would the discussion section of this repository be the right place? I ask because I went through the history of posts there and didn't find the sort of topics you describe. And the release notes have only pointed to sentences too.

I'd love to contribute (effectively), but need to know where to look! 

Thanks @balloob!

balloob:
Yeah, discussions would be the right place. As we're still building things out, we started with a small set of intents and are still iterating on how those all work (we rewrote the parser just this week!). We plan on adding more intents. If you start a discussion, 1 topic per intent please (unless they are coupled, like TurnOn/TurnOff)

responses/en/HassPersonGetLocation.yaml
peter-dolkens: Sure - this was intended as an MVP to try feel out intents and how they work.

It seems that this isn't quite the right place/time for that yet though.

peter-dolkens:
Thanks @balloob - I was (trying) to add the intent with this PR, guess I misunderstood how this was intended to work 😅

peter-dolkens:
Following up on this @balloob - where would I be proposing intents? I see "Supported Sentences" - do we need a "Supported Intents" category, or do I just throw it in "Ideas"?

balloob:
well to be really honest, @synesthesiam and me already had a meeting about this and we're going to look into a `HassGetState` intent that is generic and should work with a wide range of domains.

PR #867, opened 2023-02-02
Added graag to skip words and make doe optional

TyzzyT:
It's polite to say 'graag' when interacting with an assistant. I believe this should be added to the skip words.

Also: it's currently not possible to say something like: plafondlamp woonkamer aan without saying 'doe' first. Sso I added that to the turnOn and turnOff files. I'm not sure if I did this correct, so please decline if it's not how it should be.

### DIFF
--- a/sentences/nl/homeassistant_HassTurnOff.yaml
+++ b/sentences/nl/homeassistant_HassTurnOff.yaml
@@ -3,7 +3,7 @@ intents:
   HassTurnOff:
     data:
       - sentences:
-          - <doe> <name> [<naar>] uit
+          - [<doe>] <name> [<naar>] uit
           - <name> uit
       - sentences:
           - sluit <name>
--- a/sentences/nl/homeassistant_HassTurnOff.yaml
+++ b/sentences/nl/homeassistant_HassTurnOff.yaml
@@ -4,7 +4,6 @@ intents:
     data:
       - sentences:
           - [<doe>] <name> [<naar>] uit
-          - <name> uit
       - sentences:
           - sluit <name>
           - "<doe> <name> dicht"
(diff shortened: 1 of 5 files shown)


sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -4,6 +4,7 @@ intents:
     data:
       - sentences:
           - <doe> <name> [<naar>] uit
```
TheFes, line 6: you can make `<doe>` optional instead of adding another sentence.
So remove the added sentences and change `<doe>` in the existing sentence to `[<doe>]`

sentences/nl/homeassistant_HassTurnOn.yaml
```diff
@@ -4,6 +4,7 @@ intents:
     data:
       - sentences:
           - <doe> <name> [<naar>] aan
```
TheFes, line 6: you can make `<doe>` optional instead of adding another sentence.
So remove the added sentences and change `<doe>` in the existing sentence to `[<doe>]`

TyzzyT:
Good one, thanks! I updated my branch with your suggestions @TheFes

TheFes:
@TyzzyT you only changed if for `on` now, not for `off`

TyzzyT:
@TheFes I changed it for both, see the two lines 'Make doe optional' above my previous comment

TheFes:
@TyzzyT you are right, I got confused because you didn't remove the line you added for `off`. Can you do that?

TyzzyT:
Good one! I totally forgot about the tests, but I updated it now. Thanks for the help and all your great work on HA @TheFes!

TheFes:
@TyzzyT 
de error is caused because the new sentences need quotes around them
`"[<doe>] <name> [<naar>] uit"`
`"[<doe>] <name> [<naar>] aan"`

Can you add these?

TheFes:
Without them, the `[]` are read as a json list.

TheFes (approved):

TheFes:
@TyzzyT Thanks for your contribution!

PR #868, opened 2023-02-02
Update _common.yaml

andydir87:
Small Change in the German _common.yaml to make tha Assist understanding more German sentences:
- adding the possibility to say "mach" for key "machen"
- changing "änderen" to "ändern" in key "setzen"


sentences/de/_common.yaml
```diff
@@ -52,8 +52,8 @@ expansion_rules:
   auf: "(auf | hoch | rauf | nach oben)"
   zu: "(zu | [her]runter | nach unten)"
   schliessen: "(schließ | schließe | zumachen)"
-  machen: "(mach[e|en])"
-  setzen: "(setz[e|en] | stell[e|en] | einstellen | änder[e|en]| veränder[e|n])"
+  machen: "( mach | mach[e|en])"
```
mib1185, line 55:
```suggestion
  machen: "(mach([e|en]))"
```
this does the same

mib1185: i need to revise my statement, because `mach[e|en]` does already match `mach` , `mache` und `machen`

mib1185 (changes requested):
This PR can be closed, because the wanted behavior is already covered by current code.

PR #875, opened 2023-02-02
Create issue templates

TheFes:
Attempt to create issue templates :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/generic_issue.yaml
+++ b/.github/ISSUE_TEMPLATE/generic_issue.yaml
@@ -1,0 +1,126 @@
+name: Report a generic issue
+description: Report a generic issue related to Home Assistant Intents.
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This issue form is for reporting issues related to Home Assistant Intents
+
+        If you have a feature or enhancement request, please use the [feature request][fr] section of our [Community Forum][fr].
+
+        [fr]: https://community.home-assistant.io/c/feature-requests
+  - type: dropdown
+    validations:
+      required: false
+    attributes:
+      label: Is the issue related to a specific language?
+      description: >
+        Select the language to which this issue is related.
+      options:
+        - ar
+        - bg
+        - bn
+        - ca
+        - cs
+        - da
+        - de
+        - de-CH
+        - el
+        - en
+        - es
+        - fa
+        - fi
+        - gl
+        - he
+        - hi
+        - hr
+        - hu
+        - id
+        - is
+        - it
+        - ka
+        - kn
+        - lb
+        - lt
+        - lv
+        - ml
+        - mn
+        - ms
+        - nb
+        - nl
+        - pl
+        - ro
+        - ru
+        - sk
+        - sl
+        - sr
+        - sv
+        - sw
+        - te
+        - tr
+        - uk
+        - ur
+        - vi
+        - zh-cn
+        - zh-hk
+        - zh-tw
+  - type: textarea
+    validations:
+      required: true
+    attributes:
+      label: The problem
+      description: >-
+        Describe the issue you are experiencing here, to communicate to the
+        maintainers. Tell us what you were trying to do and what happened.
+
+        Provide a clear and concise description of what the problem is.
+  - type: markdown
+    attributes:
+      value: |
+        ## Environment
+  - type: input
+    id: version
+    validations:
+      required: true
+    attributes:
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/generic_issue.yaml
```diff
@@ -0,0 +1,126 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This issue form is for reporting issues related to Home Assistant Intents
+
+        If you have a feature or enhancement request, please use the [feature request][fr] section of our [Community Forum][fr].
+
+        [fr]: https://community.home-assistant.io/c/feature-requests
+  - type: dropdown
```
balloob, line 12: We shouldn't do it like this, instead users should use labels. Because labels will trigger code owners being mentioned.

balloob: Although, maybe we can have the bot pick this up? Not sure what's the best user experience. Obviously labels means we don't have to keep this file up to date, but UX might be worse.

.github/ISSUE_TEMPLATE/generic_issue.yaml
```diff
@@ -0,0 +1,126 @@
...
+    validations:
+      required: true
+    attributes:
+      label: The problem
+      description: >-
+        Describe the issue you are experiencing here, to communicate to the
+        maintainers. Tell us what you were trying to do and what happened.
+
+        Provide a clear and concise description of what the problem is.
+  - type: markdown
```
balloob, line 77: what do you expect people to fill in here ?

.github/ISSUE_TEMPLATE/generic_issue.yaml
```diff
@@ -0,0 +1,126 @@
...
+    attributes:
+      label: What was the last working version of Home Assistant Core?
+      placeholder: core-
+      description: >
+        If known, otherwise leave blank.
+  - type: dropdown
+    validations:
+      required: true
+    attributes:
+      label: What type of installation are you running?
```
balloob, line 102: This is not necessary.

Also, we need to be careful, the more we ask, the more likely people don't finish the form.

.github/ISSUE_TEMPLATE/generic_issue.yaml
```diff
@@ -0,0 +1,126 @@
+name: Report a generic issue
```
balloob, line 1: So actual bugs with conversation in HA should be reported in core. Maybe this should just be for sentences? 

We could still have a template picker, where feature requests link to the forums, like this: https://github.com/home-assistant/frontend/blob/dev/.github/ISSUE_TEMPLATE/config.yml

.github/ISSUE_TEMPLATE/sentence_issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+    attributes:
+      label: What was the last working version of Home Assistant Core?
+      placeholder: core-
+      description: >
+        If known, otherwise leave blank.
+  - type: dropdown
+    validations:
+      required: true
+    attributes:
+      label: What type of installation are you running?
```
balloob, line 106: leave this out.

.github/ISSUE_TEMPLATE/sentence_issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+      options:
+        - Home Assistant OS
+        - Home Assistant Container
+        - Home Assistant Supervised
+        - Home Assistant Core
+
+  - type: markdown
+    attributes:
+      value: |
+        # Details
```
balloob, line 120: let's leave this out maybe? it's sentence that's not working, and we already have additional info

.github/ISSUE_TEMPLATE/sentence_issue.yaml
```diff
@@ -0,0 +1,130 @@
...
+    validations:
+      required: true
+    attributes:
+      label: Expected result
+      description: >-
+        What was your expected result of the above sentence(s)?
+  - type: markdown
+    attributes:
+      value: |
+        ## Environment
```
balloob, line 84: let's drop this.

.github/ISSUE_TEMPLATE/generic_issue.yaml
balloob: We should not support multiple languages. It makes everything more complicated. Just let users mention it in their "additional info".

Agree on the test.

PR #876, opened 2023-02-02
[NL] Review existing intents and update

TheFes:
Main changes:
* Make `<doe>` optional everywhere
* Make `in` optional and change to `[in|op]` for `op zolder` or `op terras`
* Double quotes everywhere (with the risk they get removed again)
* Capitalized all tests because I think it looks better (and for consitancy)


balloob (approved):

responses/nl/HassTurnOn.yaml
```diff
@@ -3,8 +3,8 @@ responses:
   intents:
     HassTurnOn:
       default: "{{ slots.name }} aangezet"
-      lights_area: De lampen in {{ slots.area }} zijn aangezet
-      fans_area: De ventilatoren in {{ slots.area }} zijn aangezet
+      lights_area: "De {{ slots.area | trim }}verlichting is aangezet"
```
balloob, line 6: why do we need to trim? That is implying that "area" contains a space?

TheFes: Without `|  trim` I always had a space between area name and verlichting

PR #879, opened 2023-02-02

MathiasThinsz:
In Swedish dimming is called DIMMA not DIMRA


larsdunemark (changes requested):
Hi, words inside of `<>` is variables that will expand, in the case of `<dimra>` it will expand to:
`(ändra | sätt | vrid | dra [på] | skruva [på] | dimra | dimma | ställ [in] | höj | sänk)`
Since the variable isn't changed I suspect that this doesn't pass the tests.

MathiasThinsz:
I missed the _common.yaml file.

PR #881, opened 2023-02-02
[NL] Add a an extra sentences for light_HassTurn(On | Off)

berendhaan:
I've seen this sentence come across the community (lamp [entity] aan), so I've added it to the list.

### DIFF
--- a/sentences/nl/light_HassTurnOff.yaml
+++ b/sentences/nl/light_HassTurnOff.yaml
@@ -3,10 +3,9 @@ intents:
   HassTurnOff:
     data:
       - sentences:
-          - "<doe> [alle] <lamp> [<naar>] uit in <area>"
+          - "[<doe>] [alle] <lamp> [<naar>] uit in <area>"
           - "<doe> [alle] <lamp> in <area> [<naar>] uit"
           - "<doe> (alle | in) <area> <lamp> [<naar>] uit"
-          - "<lamp> uit in <area>"
         response: lights_area
         slots:
           domain: light
--- a/sentences/nl/light_HassTurnOff.yaml
+++ b/sentences/nl/light_HassTurnOff.yaml
@@ -6,6 +6,7 @@ intents:
           - "<doe> [alle] <lamp> [<naar>] uit in <area>"
           - "<doe> [alle] <lamp> in <area> [<naar>] uit"
           - "<doe> (alle | in) <area> <lamp> [<naar>] uit"
+          - "<lamp> uit in <area>"
         response: lights_area
         slots:
           domain: light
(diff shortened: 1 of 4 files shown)


sentences/nl/light_HassTurnOn.yaml
```diff
@@ -9,6 +9,8 @@ intents:
           - "Schakel [alle] <lamp> [in] <area> in"
           - "<doe> (alle | in) <area> <lamp> aan"
           - "Schakel (alle | in) <area> <lamp> in"
+          - "<lamp> aan"
```
TheFes, line 12: this sentence is missing an <area> definition
It is not allowed to add sentences to turn on all lights, regardless of an area

sentences/nl/light_HassTurnOn.yaml
```diff
@@ -9,6 +9,8 @@ intents:
...
+          - "<lamp> aan in <area>"
```
TheFes, line 13: You can make `<doe>` optional in `"<doe> [alle] <lamp> [<naar>] aan in <area>"` to make this option possible without adding an additional sentence.

sentences/nl/light_HassTurnOff.yaml
```diff
@@ -6,13 +6,15 @@ intents:
...
           - "<doe> (alle | in) <area> <lamp> [<naar>] uit"
+          - "<lamp> uit in <area>"
         response: lights_area
         slots:
           domain: light
           name: all
       - sentences:
           - "<doe> (overal | alle) <lamp> uit"
           - "<doe> <lamp> overal uit"
+          - "<lamp> uit"
```
TheFes, line 17: this sentence is missing a test

tests/nl/light_HassTurnOn.yaml
```diff
@@ -15,6 +15,8 @@ tests:
       - "Doe in de keuken de lampen aan"
       - "Mogen in de keuken de lampen aan?"
       - "Zet in de keuken het licht aan"
+      - "Lamp keukenlamp aan"
```
TheFes, line 18: This sentence doesn't match with any of the sentences.
Note that spaces are taken into account as of Hassil 1.0

This sentence would match with `<lamp> <area><lamp> aan` or with `<lamp> <name> aan`
Both do not match with any of the intents.

TheFes:
@berendhaan I resolved the conflicts which were caused because a PR of mine was merged

sentences/nl/light_HassTurnOn.yaml
berendhaan: Makes sense, I've changed that.

TheFes:
Looks fine now, thanks for your contribution @berendhaan

PR #900, opened 2023-02-04
Added min/max brightness setting

dziadeck47:


sentences/pl/_common.yaml
```diff
@@ -33,6 +33,12 @@ lists:
       type: "percentage"
       from: 0
       to: 100
+  brightness_level:
+    values:
+      - in: (max[a|imum] | maks[a|imum])
```
Uriziel01, line 38: What about forms like `maksymalną` and `minimalną`? Wouldn't that more natural to ask 'Ustaw jakość lampy na maksymalną' instead of 'Ustaw jasność lampy sufitowej na maksimum'?

Uriziel01:
I've also allowed myself to fix a typo in your PullRequest title @dziadeck47 😄 Hope you do not mind.

Uriziel01:
LGTM, let me just run the build just to be sure and we are good to go.

Uriziel01:
These changes where already introduced in another PR - duplicate.

PR #919, opened 2023-02-05
[RU] Add prep "[в|на]" to <area> parts

saintman23:


HepoH3:
You should put all this similar changes in one pull-request. Moreover, I examined sentences in more details. This prep are already placed in _common.yaml, so I should to remove them from other files entirely.

[  area: "[в|на] {area}"](https://github.com/home-assistant/intents/blob/2f310edce50d9d6fd64412e2e312831c06c2f0d5/sentences/ru/_common.yaml#L49)

saintman23:
understood, missed info regarding _common.yaml file.

PR #933, opened 2023-02-06
[SV] Possibility to turn on light in area.

edgein666:


larsdunemark:
I'm not sure how this is suppose to be handle in assist, is it even possible to have same entity twice in home assistant?
Example light.bedroom_readlight in 2 differnet rooms would probably be
light.child_bedroom_readlight and light.masterbedroom_readlight.
Combining this with alias for devices. Or would it be possible to have 2 devices with same alias and drill down with area? Like suggested in this PR to add support for.

edgein666:
It is possible to have the same alias on more than one device so I'm hoping it should work with alias+area to differentiate between them.

sentences/sv/light_HassTurnOn.yaml
```diff
@@ -8,9 +8,14 @@ intents:
...
 
       - sentences:
           - "<slå_på> <ljuskällor> <name>"
         slots:
           domain: light
+
+      - sentences:
+          - "<slå_på> <name> <i_på> <area>"
+        slots:
+          domain: light
```
larsdunemark, line 21: I don' think that this should be under the `domain: light` since there isn't anything defining that this is just for lights in the sentence. 
This could also trigger `starta diskmaskinen i köket` that wouldn't be a light but the dishwasher.

larsdunemark (changes requested):

sentences/sv/light_HassTurnOn.yaml
edgein666: makes sense, can I just remove domain altogether or how would you solve it? if you know...

larsdunemark: yes it is for light but the sentence does not limit to light it would also match a sentence like:
`turn on the tv in the livingroom` = "starta tv i vardagsrummet` but in that case it would include the domain light. I'm not sure how that should be handled. 
I think that this sentence should be moved from light to `homeassistant_HassTurnOn.yaml` and `homeassistant_HassTurnOff.yaml`

edgein666: maybe we should add to `homeassistant_HassTurnOn.yaml` and even other files like `fan_HassTurnOn.yaml` but that doesn't mean we should remove it from `light_HassTurnOn.yaml`, right?

sentences/sv/light_HassTurnOff.yaml
```diff
@@ -7,10 +7,15 @@ intents:
...
           - "<slå_av> [<alla>] <area>[s]<ljuskällor>"
-
         slots:
           domain: light
+
       - sentences:
           - "<slå_av> <ljuskällor> <name>"
         slots:
           domain: light
+
```
schizza, line 17: add 
```
requires_context:
 domain: light
```

to manipulate only light domain

edgein666: nice, thanks! what do you think @larsdunemark?

larsdunemark: Sure that should probably be added for the sentences for turning on light.

sentences/sv/light_HassTurnOn.yaml
larsdunemark: I don't know how same sentences would work with different Intents, adding the same to homeassistant_HassTurnOn would trigger 2 intents?

edgein666:
will close this and create a new one in homeassistant_*

mrdanielk1:
> will close this and create a new one in homeassistant_*

Was this created?

PR #940, opened 2023-02-07
[NB] Fix sentences (updated parser rule)

LaStrada:
Fixing two sentences to match the updated parser introduced in #830


sentences/nb/homeassistant_HassTurnOn.yaml
```diff
@@ -4,4 +4,4 @@ intents:
     data:
     - sentences:
       - "<skru_på> <navn>"
-      - "(skru | slå) <navn> på"
+      - "(skru|slå) <navn> på"
```
schizza, line 7: you missed to change `language: 'nb'`

PR #943, opened 2023-02-07
Improve development documentation

Arkkimaagi:
I had few problems when starting to develop intents, so I wanted to clarify the documentation a bit.

1. The README.md did not say that I should rebuild the code with `script/setup` again after initial setup. I had problems with how things were working and it was fixed only after @balloob guided me to rebuild to get latest dependencies.
2. The `source venv/bin/activate` bit was not explained at all and I had an uneasy feeling that my computer was left in some "activated" state after running it. There was no documentation why I had to run it or how I could "un-activate" it. After reading some documentation, I decided to ease possible similar frustration by others.

Please check that my documentation changes do not lead people to wrong direction, I'm not a native English or Python speaker. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,9 @@ See the [documentation](docs/README.md) for more information.
 
 # Development
 
-Checkout the repository and get a development environment with `script/setup`. This will create a new virtual environment in the `venv` directory of the repository, and install all necessary requirements.
+Checkout the repository and get a development environment with `script/setup`. This will create a new virtual environment in the `venv` directory of the repository, and install all necessary requirements. It will also build your code space to install latest dependencies, so you should run it again occasionally.
 
-Before developing, always activate your virtual environment with `source venv/bin/activate`.
+Before developing, always activate your virtual environment with `source venv/bin/activate`. This handles python dependencies so that they do not conflict with other python projects. You can deactivate the virtual environment by running `deactivate`.
 
 ## Run tests
 


README.md
```diff
@@ -28,9 +28,9 @@ See the [documentation](docs/README.md) for more information.
 
 # Development
 
-Checkout the repository and get a development environment with `script/setup`. This will create a new virtual environment in the `venv` directory of the repository, and install all necessary requirements.
+Checkout the repository and get a development environment with `script/setup`. This will create a new virtual environment in the `venv` directory of the repository, and install all necessary requirements. It will also build your code space to install latest dependencies, so you should run it again occasionally.
```
balloob, line 31: this is not correct. You either use a codespace, which does `script/setup` automatically.

We better remove this whole section and point at https://github.com/home-assistant/intents/blob/main/docs/codespace/README.md

README.md
```diff
@@ -28,9 +28,9 @@ See the [documentation](docs/README.md) for more information.
...
 
-Before developing, always activate your virtual environment with `source venv/bin/activate`.
+Before developing, always activate your virtual environment with `source venv/bin/activate`. This handles python dependencies so that they do not conflict with other python projects. You can deactivate the virtual environment by running `deactivate`.
```
balloob, line 33: Explaining what a virtual environment is seems out of scope for the README. We better just link the term in the previous paragraph to the official documentation.

README.md
Arkkimaagi: So one is not supposed to develop on their own machine, but on github codespaces instead? Interesting 🤔 

Is the local developing something that's frowned upon?

README.md
Arkkimaagi: If one is not running this on some codespaces or cloud, you're practically asking people to just start a virtual environment on their own computer and recommending them to run it always, but not explaining even in two sentences what it is or how they can "undo" the activation to get to back to the old state.

Adding two sentences to ease users mind did not seem such a big thing, but I'm fine either way. I'm just trying to help others. 😃 

Shall I close this PR?

README.md
tetele: @balloob even if you use codespaces, they only run `script/setup` upon provisioning. If you reuse them, you still need to run it manually when dependencies change.

I've personally suggested this a few times on Discord to people struggling with their setup and I think it's not a bad idea to have it in the README.

schizza: Yes, running script/setup should be mentioned, because beginners will never re-run setup a thus will never update dependencies.
It might be good to know, that when some is starting new branch and there were some merges in the main brach, it is recommended to re-run setup to use latest updates to dependencies.

balloob: People need to just rebuild their code space to get the latest dependencies. That's what it is for. They should not have to learn development commands about setting up environments.

balloob: You are totally fine to run this on your own machine. However, 99% of the people contributing to this repository is not a developer and doesn't know any of this. Adding instructions will only scare them away. Instead, we should push codespaces as that's doing everything for people out of the box.

tetele: Then can we at least mention that you need to rebuild your codespace for each contribution?

balloob: For a code space we could add a task to install requirements and mark it as [`runOn: folderOpen`](https://code.visualstudio.com/Docs/editor/tasks#_run-behavior) to keep it up to date.

frenck:

../Frenck

Issue #863, opened 2023-02-02
Turn on/off all lights (no area specified)


helldog136:
Hi, 

I just tested the new intents and Assist doesn't seem to understand the intent of turning on (or off) all the lights without specifying an area. I think that this should be a nice addition.

Thanks again for all the nice work :)
Kind regards

TheFes:
Please specify the language you are using

synesthesiam:
We're avoiding adding sentences like this because it may have unintended consequences for some users. Without specifying an area, this would turn on or off every light entity on every device in your house. This is definitely possible to do with a custom sentence + automation, though.

piitaya:
IMO, it can be good to have thing for some domain. I often use "close all shutters" with Siri or Google Assistant. Both support it. Of course, the word "all" must be present in the sentence.

piitaya:
IMO, it can be good to have thing for some domain. I often use "close all shutters" with Siri or Google Assistant. Both support it. Of course, the word "all" must be present in the sentence.

TheFes:
> IMO, it can be good to have this for some domains/device classes. I often use "close all shutters" with Siri or Google Assistant. Both support it. Of course, the word "all" must be present in the sentence.

I think if you say 'Turn of the light` in a room without a light to Google Assistant, it will turn off all lights.
I've accidentally turned on all lights more than once (In the evening, while the kids were sleeping.. That's not so good for the HAF)

piitaya:
I can confirm, if the speaker is in a room without light, it will turn on/off all the lights (pretty bad experience)

TheFes:
@synesthesiam should I remove the turn all off sentences from Dutch? Or should they work.
I see I actually didn't make `all` or `everywhere` mandatory, so I will either add that, or remove the sentences completely

synesthesiam:
Sorry for the late reply! I see the PR for making them mandatory has already been merged, which is what I would have recommended :stuck_out_tongue:

TheFes:
@synesthesiam Yes, without it a simple `light off` (in Dutch) would have turned everything off..
But does it also work in the nightly now? Unfortunately my nightly VM broke, and I didn't have time to fix it yet, so I can't test it :)

synesthesiam:
I'll have to check on nightly.

PR #961, opened 2023-02-09
[NL] alle/overal mandatory when turning all off

TheFes:
Avoid you can simply say `lampen uit` to turn off all lights.


TheFes:
Also related to https://github.com/home-assistant/intents/issues/863

TyzzyT:
Looking good!

balloob (approved):

PR #969, opened 2023-02-09
[NB] Min / max light brightness

svarthvitt:
Adding min/max brightness with tests. A few clean-ups along the way.

### DIFF
--- a/sentences/nb/light_HassLightSet.yaml
+++ b/sentences/nb/light_HassLightSet.yaml
@@ -1,25 +1,41 @@
-language: nb
+language: "nb"
 intents:
   HassLightSet:
     data:
       - sentences:
-        - "sett <navn> [til] [lysstyrke] [til] <lysstyrke>"
-        - "sett lysstyrke[n] <i_på> <navn> til <lysstyrke>"
+        - "<endre> <navn> [til] [lysstyrke] [til] <lysstyrke>"
+        - "<endre> lysstyrke[n] <i_på> <navn> til <lysstyrke>"
         response: brightness
       - sentences:
-          - "sett lysstyrke[n] <i_på> <område> til <lysstyrke>"
-          - "sett <område> [til] [lysstyrke[n]] [til] <lysstyrke>"
+          - "<endre> lysstyrke[n] <i_på> <område> til <lysstyrke>"
+          - "<endre> <område> [til] [lysstyrke[n]] [til] <lysstyrke>"
         slots:
           name: all
         response: brightness
+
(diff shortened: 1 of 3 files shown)


sentences/nb/light_HassLightSet.yaml
```diff
@@ -1,25 +1,41 @@
...
           name: all
         response: brightness
+
+      # Max/Min brightness
       - sentences:
-          - "sett <navn> farge til <farge>"
-          - "sett [farge[n] <i_på>] <navn> til <farge>"
-          - "sett <område>[s][ ]lys"
+          - "<endre> <navn> [til] {brightness_level:brightness}"
+          - "<endre> lysstyrke[n] [til] <navn> til {brightness_level:brightness}"
```
LaStrada, line 19: Should `til` be optional here? Example:
Would you ever say "Endre/sett kjøkkenlys 100%"? Or would you always include `til`?

On the sentence here, the first `til` is optional, but the second one is required.

sentences/nb/light_HassLightSet.yaml
```diff
@@ -1,25 +1,41 @@
...
+        requires_context:
+          domain: light
+        response: "brightness"
+
+      - sentences:
+          - "<endre> lysstyrke[n] <i_på> <område> til {brightness_level:brightness}"
+          - "<endre> <område> [til] [lysstyrke[n]] [til] {brightness_level:brightness}"
```
LaStrada, line 26: Same here. For the first sentence, `til` is required. On the second sentence both `til` are required.

svarthvitt (commented):
Updated based on comments from LaStrada

LaStrada (approved):
Great work!

sentences/nb/light_HassLightSet.yaml
LaStrada: I was mainly thinking of the already optional `til`, maybe a bit poorly written question/comment.

First sentence:
* Sett kjøkkenlys 50%

And for the second sentence:
* Sett lysstyrke kjøkkenlys til 50%

I am not sure how "strict" we should be with that. It is grammatically correct to have the `til` (to) there, like this:
* Sett kjøkkenlys **_til_** 50%
* Sett lysstyrken **_til_** kjøkkenlys **_til_** 50%

The second `til` in the second sentence can be skipped if we want "lazy" people (me included) to skip.

I'm good with this for now and we can revisit sentence like this in the future (including "lazy" people, me included, when writing chatbot texts).

PR #973, opened 2023-02-10
Make je, jij and u optional

TyzzyT:
@TheFes discovered that you can currently not say `Zou de eettafellamp aan mogen?`. This is because `je, jij or u` is mandatory right now. This PR makes it optional.
I also added two tests for the new expansion rule.


TheFes (approved):
Thanks!

PR #979, opened 2023-02-10
Add HassGetState sentences for English

synesthesiam:
This adds some initial sentences for English queries (`HassGetState` intent). Empty templates and responses and added for other languages.

Responses now have access to `state` and two lists of states: `query.matched` and `query.unmatched`. The `state` variable is just the first state that matches the intent. The other two lists are for `HassGetState` specifically.

Like commands, queries can have `name`/`area`/`domain`/`device_class` slots that filter entities. They also have an optional `state` slot with values like "on" or "closed". When the `state` slot is present, `query.matched` will contain entities whose state matches the `state` slot. And `query.unmatched` will contain the entities that passed the filter (name, area, etc.) but whose state does not match the `state` slot.

By checking `query.matched` and `query.unmatched` in your response templates, you can answer questions like "are all windows open in the kitchen?". If `query.unmatched` has anything in it, then the answer is "no". The English responses print out the names of the entities in cases like this.

In the future, we will probably want to create "short" and "long" responses for different clients. For now, though, we're just wanting to get something out to the HA community :slightly_smiling_face:


responses/en/HassGetState.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      which: |
+        {% if not query.matched: %}
+        Not any
+        {% else: %}
+        {{ query.matched | map(attribute="name") | sort | join(", ") }}
+        {% endif %}
+
+      how_many: |
+        {% if not query.matched: %}
```
balloob, line 34: I guess this is not necessary?

responses/en/HassGetState.yaml
```diff
@@ -0,0 +1,38 @@
...
+        Yes: {{ query.matched | map(attribute="name") | sort | join(", ") }}
+        {% else: %}
+        No
+        {% endif %}
+
+      all: |
+        {% if not query.unmatched: %}
+        Yes
+        {% else: %}
+        No: {{ query.unmatched | map(attribute="name") | sort | join(", ") }}
```
balloob, line 23: : is not good for speech to text. Be explicit of what you want to be said

responses/en/HassGetState.yaml
```diff
@@ -0,0 +1,38 @@
...
+      one: "{{ slots.name }} is {{ state.state_with_unit }}"
+      one_yesno: |
+        {% if query.matched: %}
+        Yes
+        {% else: %}
+        No, {{ slots.name }} is {{ state.state_with_unit }}
+        {% endif %}
+      any: |
+        {% if query.matched: %}
+        Yes: {{ query.matched | map(attribute="name") | sort | join(", ") }}
```
balloob, line 14: Same

responses/en/HassGetState.yaml
synesthesiam: Without this, it's an empty box in Assist :/

PR #984, opened 2023-02-11
[NB] HassGetState - first pass

svarthvitt:
This is the first pass of a translation to NB. Note that the states still return "on/off", and that some of the tests still says "soveromslampe is off". Covers is not part of this PR (except an update of _common to prepare for different kind of covers.


LaStrada (approved):
Great work 🤩

PR #988, opened 2023-02-12
Hebrew response fixes.

yosilevy:
Fixed to past tense like English. Also fixed Grammar.

Thanks for your work!


yosilevy:
I have no idea why YAML validation fails. @leranp ?

haim-b:
@yosilevy looks like there are still failures

leranp:
> I have no idea why YAML validation fails. @leranp ?

I am not sure if this is the Grammar that we need to use , google assistant response with " turning the lights off" , it is an action response , not a action that already happened

PR #994, opened 2023-02-12
[EN] More explicit respones for GetState

tetele:
Confirm requested state in response

### DIFF
--- a/responses/en/HassGetState.yaml
+++ b/responses/en/HassGetState.yaml
@@ -11,7 +11,7 @@ responses:
         {% endif %}
       any: |
         {% if query.matched: %}
-        Yes, {{ query.matched | map(attribute="name") | sort | join(", ") }}
+        Yes, {{ query.matched | map(attribute="name") | sort | join(", ") }} {{ 'is' if (query.matched|length) == 1 else 'are' }} {{ slots.state }}
         {% else: %}
         No
         {% endif %}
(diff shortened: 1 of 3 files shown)


TheFes:
You can do add `and` to the last item instead of `, ` in a more readable way like this:
```jinja
        {% if not query.matched %}
          Not any
        {% else: %}
          {% set match = query.matched | map(attribute="name") | sort | list %}
          {% if match | length > 4 %}
            {{ match[:3] | join(", ") }} and {{ (match | length - 3) }} others
          {% else %}
            {%- for name in match -%}
              {% if not loop.first and not loop.last %}, {% elif loop.last and not loop.first %} and {% endif -%}
              {{ name }}
            {%- endfor -%}
          {% endif %}
        {% endif %}
```
This also limits it to maximium 4 names, if there are more than 4, it will list 3 names and adds a count of the remaining items (eg `Bedroom light, Kitchen light, Bathroom light and 5 others`)
With for or list, it will join the first part with comma, and the last one with `and` (eg `Bedroom light, Kitchen light and Bathroom light`)

tetele:
> This also limits it to maximium 4 names, if there are more than 4, it will list 3 names and adds a count of the remaining items (eg `Bedroom light, Kitchen light, Bathroom light and 5 others`)

I think that that 4 is a bit arbitrary. Maybe some people want 7 items, others want 2. Your proposal is perfectly suited for custom responses implemented in peoples' own HA instance, possibly even shareable with a blueprint, as promised by the HA team.

But if I ask "which lights are on?", I'd like to know which lights are on, not that 4 specific lights are on and also 5 others which I don't know about.

Also, code readability was not my end goal here, brevity and explicitness were. TBH, I find my version more readable, but we do all tend to fall in love with our own work. Maybe some other people can share their opinion.

TheFes:
I had another approach as well first, but @balloob suggested this in the review.
Also the limitation on the number of lights listed came from him.

I just mentioned it here to avoid that all the languages use different approaches.

balloob:
I think that what @thefes proposes is better. It's more readable, especially for people inexperienced with the template engine which are looking at English to see what responses need to look at.

About the 4 -> having a voice assistant list 10 items is super annoying. If you want specific results, ask for an area or something. We should aim to not have our responses grow out of control. This is definitely going to be a rule. People that want to list them all 10 can always override it in their HA instance. We need to make a system with sensible defaults.

tetele:
OK, then, thanks for the feedback!

@TheFes I see no point in me rewriting what was wrong with my approach (everything, basically) with your work. Please take the credit.

PR #1009, opened 2023-02-14
[zh-cn] Add HassGetState for Simplified Chinese

al-one:


al-one:
This PR implements `HassGetState` support for Simplified Chinese, and passed verification and testing, you can review and merge it if it's ok.

PR #1024, opened 2023-02-15
Use prettier instead of yamllint

synesthesiam:
Take 2: instead of linting the YAML, we'll just use [prettier](https://prettier.io/) to make the computer do the work for us! This does require having `node` installed, but we'll ensure it's available in the code space.


TheFes (approved):
Guess NL was already pretty enough, no changes there

PR #1027, opened 2023-02-15
Migrate cover sentences

synesthesiam:
Another attempt to migrate the `cover` sentences out of `homeassistant` files. Tests were migrated where possible, but I had to generate some tests in order to keep things passing.

**Please replace the auto-generated cover tests with real ones!**

## Context in Tests

Domain-specific tests like `cover_HassTurnOn` need to declare themselves as belonging to the correct domain in one of two ways: by settings `slots.domain` or by setting `context.domain`. I'll explain this a bit further.

When a sentence template matches, everything in its `slots` goes into the intent's slots. For example:

```yaml
intents:
  HassTurnOn:
    data:
      - sentences:
          - "turn on all [the] lights in {area}"
        slots:
          domain: light
```

During matching, `{area}` will automatically set the `area` slot and the `domain` slot will come from the sentence template.
In the test, we check both:

```yaml
tests:
  - sentences:
      - turn on all lights in kitchen
    slots:
      area: kitchen
      domain: light
```

The **second** way to restrict sentences is through context. Entities carry a context `dict` with them in the `{name}` list, which currently just has their `domain`. This lets us do:

```yaml
intents:
  HassTurnOn:
    data:
      - sentences:
          - "open {name}"
        requires_context:
          domain: cover
```

Without `requires_context`, we could "open" a light (not typical for English).

In the corresponding test, however, we don't have a `domain` slot to check because items from context do not automatically set slots (maybe they should, but that's a future design discussion). So instead, we now have:

```yaml
tests:
  - sentences:
      - open office door
    context:
      domain: cover
```

The `context` here will be compared to the context from the match to ensure they're equal. And now we have a domain-specific sentence test without `slots`.


mojikosu (approved):
Looking good so far

sentences/lb/homeassistant_HassTurnOn.yaml
```diff
@@ -8,16 +8,6 @@ intents:
...
-          - <maach> <all_window> <op>
-        response: cover
-        slots:
-          domain: cover
-          device_class:
-            - window
-            - blind
-            - curtain
-            - shutter
       - sentences:
```
mojikosu, line 11: Lines 11-13: Can remain here if there a) is not going to be a fileset for doors, b) doors (e.g. garage door) not considered as covers. Else, move to new fileset.

sentences/lb/homeassistant_HassTurnOff.yaml
```diff
@@ -8,16 +8,6 @@ intents:
...
-          - <maach> <all_window> <zou>
-        response: cover
-        slots:
-          domain: cover
-          device_class:
-            - window
-            - blind
-            - curtain
-            - shutter
       - sentences:
```
mojikosu, line 11: Lines 11-13: Can remain here if there a) is not going to be a fileset for doors, b) doors (e.g. garage door) not considered as covers. Else, move to new fileset.

PR #1032, opened 2023-02-16
[NB] Correct translation of autogenerated cover tests

svarthvitt:
Note that device class "garage" has not yet been translated by the system and will show as "garage ble lukket". I assume the live environment will use the translated version.


LaStrada (approved):
Noticed the same issues. Thanks for quickly fixing it!

PR #1033, opened 2023-02-16
[NL] Change split covers

TheFes:
change the way the intents are split to the separate file

### DIFF
--- a/sentences/nl/homeassistant_HassTurnOff.yaml
+++ b/sentences/nl/homeassistant_HassTurnOff.yaml
@@ -12,8 +12,9 @@ intents:
         response: "cover"
 
       - sentences:
-          - sluit <name> [in|op] <area>
-          - "[<doe>|<zou>] <name> (<dicht> [in|op] <area>|[in|op] <area> <dicht>) [willen|kunnen]\
-            \ [<doe>]"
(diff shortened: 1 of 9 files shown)


sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -12,8 +12,9 @@ intents:
...
-          - "[<doe>|<zou>] <name> (<dicht> [in|op] <area>|[in|op] <area> <dicht>) [willen|kunnen]\
-            \ [<doe>]"
-          - "<zou> <name> [in|op] <area> (willen|kunnen) [<dicht>]"
-        response: "cover_area"
+          - "sluit [de] garage[ ][deur]"
+          - "[<doe>|<zou>] [de] garage[ ][deur] dicht [willen|kunnen] [<doe>]"
+        response: "cover_device_class"
+        slots:
+          device_class: "garage"
+          domain: "cover"
```
balloob, line 20: I'm confused, why is this cover sentence in `homeassistant` domain file ?

TheFes: Okay, with that in mind I will make some changes. But then there will be only one intent left :)

balloob (approved):

PR #1052, opened 2023-02-18
[pt-br] initial version for sentences, responses and tests

dclobato:
After setting up the correct environment, these are the first version of sentences, responses and tests for brazilian portuguese


synesthesiam:
Hi @dclobato please add some test sentences for the new templates :+1:

dclobato:
@synesthesiam new PR #1111 with tests, responses and sentences. More on the way...

PR #1057, opened 2023-02-19
[hr] Covers rewrite

spuljko:


sentences/hr/cover_HassTurnOff.yaml
```diff
@@ -2,52 +2,91 @@ language: hr
...
+          - <zatvori> [sva] vrata [u|na] {area}
+          - <zatvori> {area} vrata
         slots:
-          device_class:
-            - blind
+          device_class: door
           domain: cover
+        response: door
 
+      #error: Expected window, got door for slot device_class for: zatvori ulazna vrata
```
spuljko, line 39: @synesthesiam can you please help me with this error. Test sentence is "Close the main door", but it ignores device class.

synesthesiam: It's getting caught by this template:

```yaml
      - sentences:
          - <zatvori> <name>
        requires_context:
          domain: cover
        slots:
          device_class: window
        response: window
```

because the name of the entity is "ulazna vrata"

spuljko: I am aware of that. But, in fixtures "ulazna vrata" has device class "door", not "window". I am trying to write different responses for windows and doors.

spuljko: Never mind, I will rewrite it with device name instead of class.

PR #1058, opened 2023-02-19
[En] Add room phrase support to En Expansion Rules

tylerj117:
Allow user to optionally include the words "room", "bedroom", or "bed room"


sentences/en/_common.yaml
```diff
@@ -89,7 +89,7 @@ lists:
         out: window
 expansion_rules:
   name: "[the] {name}"
-  area: "[the] {area}"
+  area: "[the] {area} [bed][ ][room]"
```
tetele, line 92: I'm not sure you understood how this works. If you have an area called "bedroom" or "bed room", it will automatically be matched by the {area} placeholder. Not only is there no need for this addition here, but it will also mess with things, as you could say "turn on the lights in the kitched bed" and it would turn on the kitchen lights.

See the docs https://developers.home-assistant.io/docs/voice/intent-recognition/template-sentence-syntax

tylerj117: Maybe this would be better resolved in naming of my areas then, but this was intended in cases where an area name does not include room/bedroom. For example, areas named "Primary" or "John", that refer to bedrooms would benefit from these options. 

Would it be more appropriate to add a new expansion rule: `bedroom: "['s] [(bed[ ]|)room]"` and then optionally include this in the sentence itself, i.e. `"<turn> on [all] [the] (light | lights) in <area>[<bedroom>]"`? 

You still would have awkward phrases such as "the kitchen bedroom" but it seems more likely to me that the variation in natural language to say "room" vs "bedroom" vs "bed room" is more common than someone saying the Kitchen bedroom.

tetele: IMHO, area names should probably include the functionality of a room (i.e. "John's bedroom") instead of just the main "beneficiary" of that area. You can have the primary area name "John" and an alias for the area called "John's bedroom" if the full name is too long for your UI or for whatever other reason.

Maybe your proposal is acceptable due to the above naming convention being used very frequently, I can't say. In all honesty, it seemed so far fetched the first time I read the change that I thought you didn't understand how expansion rules work. I now realize what it is that you're trying to achieve.

However, `bedroom: "['s] [(bed[ ]|)room]"` makes no sense, especially if you're planning on making it optional in a sentence such as `"<turn> on [all] [the] (light | lights) in <area>[<bedroom>]"`. 1- It matches a void string and 2 - you're using the brackets... inefficiently, to say the least :)

Maybe you meant something like `bedroom: "['s] [bed[ ]]room"`.

synesthesiam: I think this is something that should be done in aliases on the HA side instead of here in the sentence templates.

PR #1060, opened 2023-02-19
[NL] Split curtain from other cover device classes

TheFes:
I've split of the curtain device class from the other supported cover device classes.

This can be convenient on a sunny day, when you want to open the curtains in eg your bedroom, but want to keep the shades in the same state to avoid the room getting too hot.

For now I only taken out curtain, and I only match with the other device classes left in (in which I added `shade`). Optionally every device class could get it's own sentences now.


balloob (approved):
Nice one.

PR #1065, opened 2023-02-21
[CS] Cover responses for one entity

schizza:
One entity response for covers

TODO: pass entity device_class not genric response device_cass


schizza:
@synesthesiam 
Hi, I am making one entity responses for covers. But there is an issue:
We should pass to responses entity device_class that we are currently handling not device_class that is passed from sentences. 

In my point of view (as it is in this commit):
when we are passing device_class from sentences result:
```
{
  "text": "zavři vchodové dveře",
  "match": true,
  "intent": "HassTurnOff",
  "slots": {
    "name": "Vchodové dveře",
    "device_class": [
      "blind",
      "door"
    ],
    "domain": "cover"
  },
  "context": {
    "domain": "cover"
  },
  "response_key": "new_cover",
  "response": "zatahuji vchodové dveře zavírám vchodové dveře"
```
so there are multiple responses as `zatahuji vchodové dveře` and `zavírám vchodové dveře` because in responses in `slots.device_class` a list of `device classes` are passed from sentences not from defined entity in _fixtures

Also, even when I set `requires_context` it parser will throw the first sentence no matter of what context it requires.

schizza:
@synesthesiam 
Please check this commits. I tried to make possible to match entity device_class to pass it to response template.

PR #1081, opened 2023-02-21
SV: Adding GetState and response

larsdunemark:


tests/sv/_fixtures.yaml
```diff
@@ -14,18 +14,66 @@ entities:
   - name: garagelampa
     id: light.garage_light
     area: garage
+    state:
+      in: "av"
+      out: "off"
```
larsdunemark, line 19: I'm not sure this is the correct way to fix the testcases for not responding with "off" instead of "av" when a light is off. I don't see how this will work in the end product where a light will have a state. Or will is just use the translated state from HA?

synesthesiam: We'll be using the translated state(s) from HA.

synesthesiam:
Thanks!

PR #1084, opened 2023-02-22
[UR] adding Hass Get state

AalianKhan:


AalianKhan:
Hello @synesthesiam
I am getting an Assertion error but I am stumped on this one.
image
I tried to condense the recognition as much as possible but it still doesn't go through
`"{name} کیا ہے"`

synesthesiam:
@AalianKhan You need to add a new entity to your fixtures, like:

```yaml
entities:
  - name: Outside Temperature
    id: sensor.outside_temperature
    area: garage
    state: "42"
    attributes:
      unit_of_measurement: "°F"
```

AalianKhan:
Thanks @synesthesiam! that worked
Since our scripture is right to left, our question mark is reversed `؟`, testing fails if I use this. However, it passes if I use `?`

synesthesiam:
I will add "؟" to the list of punctuation for the next version.

PR #1090, opened 2023-02-22
[NB] Shorter response according to style guide

svarthvitt:
Ref https://developers.home-assistant.io/docs/voice/intent-recognition/style-guide/ and [EN] example and proposal on Discord.

Please consider the following carefully in reviewing this PR. I considered a few different responses to meet the shorter and more accurate feedback to actions. 

We are advised not to repeat slots.name/slots.area, and to give a concise answer in the present tense and active voice. By removing slots.name / slots area, I am not certain whether it is correct to answer "It's on" or "It's on". The [ENG] example not not fully meet the style guide, but is my preferred example to follow. Here we have several options (this PR reflects the last one):

    HassTurnOn: (keep slots.name/area to ease grammar challenges)
      default: {{ slots.name }} er slått på
      lights_area: Lys i {{ slots.area }} er slått på
      fans_area: Vifte i {{ slots.area }} er slått på
      cover: {{ slots.name }} er åpnet
      cover_area: Åpnet {{ slots.area }}
      cover_device_class: "{{ slots.device_class }} er åpnet"

    HassTurnOn: (most similar to style guide)
      default: Den er på
      lights_area: Lys er på
      fans_area: Vifte er på
      cover: Den er åpnet
      cover_area: Åpnet {{ slots.area }}
      cover_device_class: "{{ slots.device_class }} er åpnet"

    HassTurnOn: (example from EN)
      default: Slo på {{ state.domain }}
      lights_area: Slo på lys
      fans_area: Slo på vifte
      cover: Åpnet
      cover_area: Åpnet {{ slots.area }}
      cover_device_class: Åpnet {{ slots.device_class }}

PR #1092, opened 2023-02-22
[NB] Introduction of locks, scripts and scenes

svarthvitt:
Reference is made to #1083 for coverage of lock, scene and script domains. 

I have also moved covers to the right files.

There are a few commits following from #1090/#1048. I am not sure how to handle these separately


svarthvitt:
Closed. Will be replaced by new PR.

Issue #1079, opened 2023-02-21
Plural word support.


ernst77:
As we are moving on with responses would be nice to have plural word support.

It's easy for english language as you only have two forms (e.g. `device` and `devices`) some languages have 3 forms for different amounts (e.g. Lithuanian 1 - `įrenginys`, 2-`įrenginiai`, 3-`įrenginių`) some even 5(or 6?).

Would there be a possibility for example to specify `variable name` and `count` and correct form would be given. 

I've had experience with https://www.i18next.com/translation-function/plurals their system seems to work fine for plurals on different languages. Maybe we could get an inspiration from them? :D

AalianKhan:
Hey
I am a bit confused about why. 
if we implement this, what will it enable us to do?

tetele:
I would suggest an option similar to `requires_context`, maybe called `requires_cardinality`, which only recognizes the intent if the matched set of entities has the exact required cardinality. The reasoning for my suggestion is that it can be applied in multiple circumstances, but the main one I can think of is for when asking about the state of "all" devices in an area, where "all"=1.

Specifically: "is the kitchen light on?" is basically the same as "are the kitchen lights on?" when the kitchen has exactly one light. However, when there are 2 lights, "are the kitchen lights on?" has a deterministic answer, whereas "is the kitchen light on?" begs the question "which light?" instead of "sorry, i didn't understand".

Here's a demo snippet of my proposal:
`responses/HassGetState.yaml`
```
responses:
  intents:
    HassGetState:

      all: |
        {% if not query.unmatched: %}
          Yes
        {% else %}
          {% set no_match = query.unmatched | map(attribute="name") | sort | list %}
          {% if no_match | length > 4 %}
            No, {{ no_match[:3] | join(", ") }} and {{ (no_match | length - 3) }} more not
          {%- else -%}
            No,
            {% for name in no_match -%}
              {% if not loop.first and not loop.last %}, {% elif loop.last and not loop.first %} and {% endif -%}
              {{ name }}
            {%- endfor %} not
          {% endif %}
        {% endif %}

      all_one: |
        There's just one {{ slots.domain }} and it's {{ state.state_with_unit }}
```

`sentences/HassGetState.yaml`
```
intents:
  HassGetState:
    data:
      - sentences:
          - are all [the] {on_off_domains:domain} {on_off_states:state} [in <area>]
        response: all
        requires_cardinality:
          above: 1
      - sentences:
[...]

PR #1097, opened 2023-02-22
[EN] Updates to the basic sentences

TheFes:
* Updates to the basic sentences to allow more brief sentences
* Add intents to turn all lights/fans off
* Added activate/deactivate to sentences
* Added optional light/switch to homeasstant turn_on/off to allow to add the type when it's not in the name


sentences/en/fan_HassTurnOff.yaml
```diff
@@ -3,10 +3,22 @@ intents:
   HassTurnOff:
     data:
       - sentences:
-          - "<turn> off [all] [the] (fan | fans) in <area>"
-          - "<turn> off <area> (fan | fans)"
-          - "<turn> [all] <area> (fan | fans) off"
+          - "<turn> off [all] [the] fan[s] in <area>"
```
balloob, line 6: `turn off fan in living room` will now turn off all fans in the living room? Not sure if I would expect that, as it's a singular word "Fan"

sentences/en/_common.yaml
```diff
@@ -96,13 +96,14 @@ lists:
 expansion_rules:
   name: "[the] {name}"
   area: "[the] {area}"
-  what_is: "(what's | whats | what is)"
-  brightness: "{brightness}[%| percent]"
-  turn: "(turn | switch | change)"
-  temp: "(temp | temperature)"
-  temperature: "{temperature}[°| degrees] [{temperature_unit}]"
+  what_is: "(what's|whats|what is)"
+  brightness: "{brightness}[%|percent]"
```
tetele, line 100: this would not match `20 percent` (i.e. with a space), but only `20percent`. I'd change it to `"{brightness}[ ][%|percent]"`, maybe?

sentences/en/_common.yaml
```diff
@@ -96,13 +96,14 @@ lists:
...
+  light: "[the] (light|lights|lighting)"
+  turn: "(turn|switch|change)"
+  temp: "(temp|temperature)"
+  temperature: "{temperature}[°|degrees] [{temperature_unit}]"
```
tetele, line 104: Same thing here without the space. `20degrees` is maybe a bit unnatural.

TheFes: Done, also added an optional space before the unit

sentences/en/fan_HassTurnOff.yaml
TheFes: Well, I gave it some thought when making the changes (not much I must admit, but that's also because in practice I only changed `(fan | fans)` to `fan[s]` here.
But in a case where you would have multiple fans in the living room (eg `Ceiling fan` and `Standing fan`) I don't expect anyone will use the singular form. Maybe if only one is on, but in that case turning both off will also have the wanted effect.
In case there is only one fan in the living room, it does make sense to aks for turning off the fan in the living room.

If Assist could ask questions as a response, we could check if multiple fans in the area are on, and if the singular form is used, let it ask which fan to turn off. That might be something for the future.

For now I would suggest to leave the singular form in, also because it could be that in the speech to text process, `fans` could be converted to `fan`

tetele: > For now I would suggest to leave the singular form in, also because it could be that in the speech to text process, `fans` could be converted to `fan`

+1

Not only that, but it's probably most likely that the user has only one fan per room, in which case one fan is all fans :)

sentences/en/_common.yaml
TheFes: I did add the optional space now, but as the tests did still pass, it doesn't seem required though. I'll leave it in though, to make clear both options will work.

sentences/en/fan_HassTurnOff.yaml
tetele: Here's a more on-point suggestion to cover this usecase https://github.com/home-assistant/intents/issues/1079#issuecomment-1442009567

cc @synesthesiam

PR #1102, opened 2023-02-23
[NL] Update and add missing intents compared to EN

TheFes:


sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -6,4 +6,8 @@ intents:
           - "[<doe>|<zou>] <name>[ ][<type>] [<naar>] uit [willen|kunnen] [<doe>]"
           - "[<zou>] <name>[ ][<type>] [willen|kunnen] (uit[ ]zetten|uitschakelen)"
         excludes_context:
-          domain: cover
+          domain:
+            - cover
+            - scene
```
balloob, line 11: you can't turn off a scene.

sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -6,4 +6,8 @@ intents:
...
+            - sensor
+            - script
```
balloob, line 13: i don't think a user expects long running scripts to be turned off.

sentences/nl/homeassistant_HassTurnOff.yaml
```diff
@@ -6,4 +6,8 @@ intents:
           - "[<doe>|<zou>] <name>[ ][<type>] [<naar>] uit [willen|kunnen] [<doe>]"
           - "[<zou>] <name>[ ][<type>] [willen|kunnen] (uit[ ]zetten|uitschakelen)"
         excludes_context:
-          domain: cover
+          domain:
+            - cover
+            - scene
+            - sensor
```
balloob, line 12: you can't turn off a sensor.

sentences/nl/homeassistant_HassTurnOn.yaml
```diff
@@ -7,4 +7,8 @@ intents:
           - "schakel <name>[ ][<type>] [<naar>] in"
           - "[<zou>] <name>[ ][<type>] [willen|kunnen] (aan[ ]zetten|inschakelen)"
         excludes_context:
-          domain: cover
+          domain:
+            - cover
+            - scene
+            - sensor
+            - script
```
balloob, line 14: Remove this. You wouldn't want to turn on all scripts tied to a room!

TheFes: Unless I misunderstood what `excludes_context` does, I would say that adding `script` and `scene` here excludes them from being targeted by this sentence. I would say they need to be in here, because `script` and `scene` now have their own files.
I have removed sensor, as they can't be turned on.

sentences/nl/homeassistant_HassTurnOff.yaml
TheFes: removed

sentences/nl/homeassistant_HassTurnOff.yaml
TheFes: removed

sentences/nl/homeassistant_HassTurnOff.yaml
balloob: oh snap, you're right. I misread this.

balloob (approved):

PR #1107, opened 2023-02-23
ca: Add GetState

duhow:


duhow:
@synesthesiam I think there's an issue while `matching` stuff. Can you help me clarify what's going on?
- "Is the light on? No, it's on" (expecting yes)
- Counting "off" lights expect 1 but always returns 0 (either asking on or off),

duhow:
@synesthesiam FYI, **all** words `on` and `off` MUST be "quoted". Otherwise YAML parser transforms them into "boolean" (True / False), which was causing mismatch issues.

synesthesiam:
Good catch. I suspect someday 1000 years in the future, a starship will dive into a black hole because a YAML parser incorrect interpreted "on" as a boolean :laughing:

PR #1115, opened 2023-02-24
[WIP][RU] Initial commit on GetState

HepoH3:
Tests with responses for 'all' and 'any' states are failed for now (They are commented out)


HepoH3:
@synesthesiam, can you have a glance at this error? When I run tests from codespaces everything fine.

It says that on sentence "Задернута ли левая штора?" should be responsed "Нет, off". First of all, there is no "Off" state for cover. Moreover, to make things work, I translated states, so there is no even `closed` state in `ru` domain.
`in/out filter` will cast first word of sentence (`задернута`) into `закрыта`:

```yaml
      - in: "закрыт[а|о|ы|ой|ая|ые|ую]|задернут[а|о|ы|ой|ая|ые|ую]|поднят[а|о|ы|ой|ая|ые|ую]"
        out: "закрыта"
```

so there is sentense `задернута ли левая штора`, which parsed by: 

```yaml
  "{cover_states:state} [ли] <name> [<area>]"
```

while entity `Левая штора` has following attributes:

```yaml
  - name: "Левая штора"
    id: "cover.curtain_left"
    area: "living_room"
    state: "открыта"
    attributes:
      device_class: curtain
```

Apparantly it's state not `закрыта`, but `открыта` instead. So response should be `Нет, открыта` (no, closed).

UPD. okay, so I've merged main branch from upstream and get this error again, so problem appeared in some commit after 8310e374. The asserted state still confuse me, maybe it's a bug?

skynetua:
This is probably because this duplicate entity in fixtures:
```
  - name: "Лев(ая|ую|ой) (занавеск|штор)(а|у|ы)"
     id: "cover.curtain_left"
     area: "living_room"
```

It has no device_class and state so off by default.

tests/ru/_fixtures.yaml
```diff
@@ -2,22 +2,49 @@ language: "ru"
...
   - name: "Потолочн(ый|ого) вентилятор[а]"
     id: "fan.ceiling"
     area: "living_room"
+
   - name: "Лев(ая|ую|ой) (занавеск|штор)(а|у|ы)"
     id: "cover.curtain_left"
     area: "living_room"
+
+  - name: "Левая штора"
+    id: "cover.curtain_left"
```
skynetua, line 32: Merge this with entity above

skynetua (changes requested):

tests/ru/_fixtures.yaml
HepoH3: > This is probably because this duplicate entity in fixtures:

That's it, thank you!

sentences/ru/_common.yaml
```diff
@@ -44,6 +44,36 @@ lists:
         out: "celsius"
       - in: "f|фаренгейта|по фаренгейту"
         out: "fahrenheit"
+  cover_states:
+    values:
+      - in: "открыт[а|о|ы|ой|ая|ые|ую]|не[ ]задернут[а|о|ы|ой|ая|ые|ую]|поднят[а|о|ы|ой|ая|ые|ую]"
+        out: "открыта"
+      - in: "закрыт[а|о|ы|ой|ая|ые|ую]|задернут[а|о|ы|ой|ая|ые|ую]|поднят[а|о|ы|ой|ая|ые|ую]"
```
skynetua, line 51: Поднята is wrong here

HepoH3: But it's a cover device states. If we talk about shutter, or blinds, we could have a sentence "подняты ли жалюзи" for example.

skynetua: Yes, but it's added for both out states "открыта" and "закрыта", so change it to опущена

HepoH3: Oh, you're right! Thanks again!

skynetua: Np

skynetua (approved):

PR #1117, opened 2023-02-24
[NL] Binary sensor support

TheFes:


sentences/nl/_common.yaml
```diff
@@ -9,6 +9,7 @@ responses:
     handle_error: "Er is iets misgegaan"
 
 lists:
+  # brightness
```
balloob, line 12: Stating the obvious 😅

TheFes: Yeah, the list (well actually a mapping) of lists became so long I wanted to bring some structure in it.
But I guess I overdid it :)

balloob (approved):

balloob (approved):

Issue #1128, opened 2023-02-24
Get area_name of matched results


TheFes:
I was thinking about creating intents like `in which room are there still lights on`

In HA you could do something like `{{ states.light | selectattr('state', 'eq', 'on') | map(attribute='entity_id') | map('area_name') | select() | unique | list | sort }}` however `area_name` is not working in the test environment, so I can not get responses using something like `{{ query.matched | map(attribute='entity_id') | map('area_name') | select() | unique | list | sort }}` to pass the tests. I would probably work in HA :)

Would it be possible to have `area_name` working in the test environment?

TheFes:
This would also be a nice addition for the binary sensors, so you can ask in which room smoke/co/sound whatever is detected

TheFes:
@synesthesiam would this be something which can be added?

synesthesiam:
Sure, that seems pretty straightforward. I've added it to my TODO list :+1:

Issue #1127, opened 2023-02-24
Check on multiple states in one intent


TheFes:
Now that we have binary_sensor support, I wanted to create intents to check if all doors/windows are closed
However, I ran into the problem that there are also covers which are doors/windows

So to check if all doors are open, we need to check for covers which are `open` (and technically also which are `opening` and `closing`) but also for binary_sensors which are `on` (all with device_class: door)

Is there a way to check on multiple states?

TheFes:
@synesthesiam Could this somehow be added, also see #1200 how this can help in resolving the current conflicts

synesthesiam:
I made  suggestion in https://github.com/home-assistant/intents/discussions/1200 but I haven't had time to think it through very much :slightly_smiling_face:

PR #1136, opened 2023-02-25
[EN] binary_sensor any, all, which, how_many

tetele:


tetele:
As discussed on Discord and as https://github.com/home-assistant/intents/pull/1136/commits/d601cb3fd7d8a7ef06355fb2bd09de747dfa96ac shows, there's a conflict between cover/binary_sensor for the door/window device_classes, as well as between lock/binary_sensor for the lock device_class.

Not sure how that should be fixed, but for the moment we can't ask "which doors are open?"

TheFes:
CC: @synesthesiam 
I've been thinking a bit about the device_classes which could not be added. The `update` device class is added now, but updates nowadays mostly use and `update` entity, and not a binary sensor. So it would make sense to add the update domain as well.

That will give similar issues as for `door`, `window` and `lock`. The issue is similar, but needs to be solved in different ways as far as I can see:

* `door`/ `window`. Both the binary sensor and the cover entity use same device class. So you could match on device class only (using requires_context). The problem then is that the states don't match (`open` vs `on`). So the intent needs to match on 2 different states, which is not possible now.
* `update`. The state is the same (`on` for both) but there is no way to match them right now. For that I guess we need to have the possibility add `or` to the `requires_context`. Something like:
```yaml
requires_context:
  or:
    - domain: update
    - and:
        domain: binary_sensor
        device_class: update
```
* `lock`. This one has both issues described above. `lock` has no device_class to match the similar binary sensors on, and the states differ (`open` vs `on`)
Also see: https://github.com/home-assistant/intents/issues/1127


Another thing is area support. Especially for the binary_sensors which detect something (`co`, `smoke`, `sound` etc) it would make sense to have an intent which returns the area, so you can ask in which area/room `smoke` is detected. Like described here. See: https://github.com/home-assistant/intents/issues/1128

TheFes:
Oh, and does it maybe make sense to create a folder `binary_sensor_HassGetState` and create files per device_class in that folder? So eg `cold_HassGetState.yaml`

synesthesiam:
Thanks @tetele! Not sure yet how we can resolve the conflict. @TheFes gave the solution I would have thought of initially: grouping context constraints in a more complex fashion with `or`. We'll have to give it some more thought for the next release cycle.

For splitting device classes into their own files, we'll need to keep in mind that the current tooling assumes all the files are `{domain}_{intent}.yaml`. One idea is to allow for a directory named `{domain}_{intent}` and have it filled with `{device_class}.yaml` files.

tetele:
For the moment, I'd like to make sure that we cover the most use cases by simply limiting the matching to the most used of the domains.

@balloob or @synesthesiam do you have any telemetry data to tell us which of each pair is most used out in the wild?
- binary_sensor/door vs cover/door
- binary_sensor/window vs cover/window
- binary_sensor/garage vs cover/garage
- binary_sensor/lock vs lock
- binary_sensor/update vs update

tetele:
> Not sure yet how we can resolve the conflict. @TheFes gave the solution I would have thought of initially: grouping context constraints in a more complex fashion with `or`. We'll have to give it some more thought for the next release cycle.
> 
> For splitting device classes into their own files, we'll need to keep in mind that the current tooling assumes all the files are `{domain}_{intent}.yaml`. One idea is to allow for a directory named `{domain}_{intent}` and have it filled with `{device_class}.yaml` files.

One issue with that approach, though: you can't have dedicated files for domains and combine domains in the same sentence/response unless you relax the validation rules for file naming.

PR #1148, opened 2023-02-27
[zh-cn] Ignore whitespace for Simplified Chinese

al-one:


mmlingyu:
with "关闭工作室的灯" or “关闭所有的灯” is not working,

 "plain": {
                "speech": "抱歉，找不到名为 所有的 的区域",
                "extra_data": null
            }

PR #1158, opened 2023-02-28
Small modifications on the existing translations

peyanski:


hristo-atanasov:
Thanks a lot for the correction and the help! :)

PR #1160, opened 2023-02-28
[EN] Support for locking/unlocking locks

TheFes:
Support for locking and unlocking locks


hristo-atanasov:
Pleace, include "in" and "out" values for translation purposases in English version of files. Lets use English as a guide to translate in other languages. Thanks!

TheFes:
@hristo-atanasov @synesthesiam 
I don't see why this intent should be different compared to any of the other `hassTurnOn`/`hassTurnOff` intents?
There are no tanslation lists used in any of these intents.

PR #1171, opened 2023-03-01
[FR] Add one sentence to get temperature by entity name and optional …

slegars56:
…area

response also added to the related test


slegars56:
Closed because of issue in test file

PR #1174, opened 2023-03-02
[NL] Expansion rules <name_area> and <met>

TheFes:
Introduced a new expansion_rule `<name_are>`

It has 3 frequently used options for the combination of the name and the area.
It has 2 advantages
* it simplifies the sentences (as all expansion rules do)
* it avoids grammatically incorrect versions or double usage of area around the name, as the 3 options are now separated, and not in one line which would allow all to be used.

### DIFF
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -311,13 +311,19 @@ expansion_rules:
   # generic expansion rules for sentences
   name: "[de|het] {name}"
   area: "[de|het] {area}"
+  in: "[in|op|van|bij]"
+  name_area: >
+    (
+      [[door|met|bij] [de|het] {area}][ ]{name}
+      |[<in> [de|het] {area}] [door|met|bij] <name>
+      |[door|met|bij] [de|het] {name} [[in|op|van|bij] <area>]
+    )
   doe: "(zet[ten]|mag|mogen|doe[n]|verander[en]|maak|maken|schakel[en])"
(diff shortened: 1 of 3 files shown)


TheFes:
I tried to do this, but then got errors because of recursive rules, so it seems I can only use another expansion rule once in a new expansion rule
```yaml
  name_area: >
    (
      [[door|met|bij] <area>][ ]{name}
      |[<in> <area>] [door|met|bij] <name>
      |[door|met|bij] [de|het] {name} [<in> <area>]
    )
```
So to avoid that, I only uses `<name>`, `<area>` and `<in>` once in the new expansion rule.

sentences/nl/_common.yaml
```diff
@@ -311,13 +311,20 @@ expansion_rules:
   # generic expansion rules for sentences
   name: "[de|het] {name}"
   area: "[de|het] {area}"
+  in: "[in|op|van|bij]"
+  met: "(door|met|bij)"
+  name_area: >
+    (
+      [[door|met|bij] [de|het] {area}][ ]{name}
+      |[<in> [de|het] {area}] [door|met|bij] <name>
```
balloob, line 319: doesn't it match now on the whitespace around `|` ?

balloob (approved):

sentences/nl/_common.yaml
TheFes: there is no `|` before the first one, and also no `|` after the third option. `<name>` is mandatory in all 3 options. So as far as I can see it doesn't match on whitespace only

PR #1194, opened 2023-03-04
[hr] Added fan intents

spuljko:
Added fan intents


momoz:
can we get fan set levels (speed, high, medium,low  and percent)?

spuljko:
@momoz as far as I know fan speed is not currently supported by intents. Please create issue on main repository,

PR #1207, opened 2023-03-07
Add total_results property to query

tetele:
In responses, we should have a handy `query.total_results`

See this Discord discussion https://discord.com/channels/330944238910963714/1054846248374505572/1078695288463704126


synesthesiam:
Looks good, thanks!

PR #1220, opened 2023-03-11
[de & en] Add sensor support

Mattstir:
Add (temperature, humidity, power) sensor support to german

Add basic sensor support to english to get the german part working

### DIFF
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -5,6 +5,15 @@ responses:
       einzeln: |
         {{ slots.name | capitalize }} ist {{ state.state_with_unit }}
 
+      hat: |
+        {{ slots.name | capitalize }} hat {{ state.state_with_unit }}
+
+      verbraucht: |
+        {{ slots.name | capitalize }} verbraucht {{ state.state_with_unit }}
+
+      luftfeuchtigkeit: |
+        {{ slots.name | capitalize }} hat {{ state.state }}%
+
       einzeln_janein: |
         {% if query.matched %}
           Ja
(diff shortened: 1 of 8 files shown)


Mattstir:
Here is my first contribution :) 

Want to note: The fact that tests don't run until there is an englisch test for the same domain cost me quite a bit of headache and looking around how the test stuff actually works. Not really contribution friendly if you just want to improve one language

mib1185 (changes requested):
Hi @Mattstir 

These intents are already covered by the `homeassistant_HassGetState` sentences.
If there is something missing or could be improved, then you very welcome to contribute to them 🙂 

thx 👍

Mattstir:
Hi @mib1185 

Thanks for the feedback.
Sadly I'm not entirely sure what you mean... 
The german part? The english part?
Shouldn't things get split by domain? Especially the sensor part can get quite big if all of the device classes get added...
Also when I look at the questions in `homeassistant_HassGetState` they don't fit at all the to power or humidity.. 🤔 

Thx :)

mib1185:
Hi @Mattstir 
before we add the support of a new platform, we have to start a discussion about it.
Further we should not combine changes of different languages in one PR (_keep a PR as small as possible and do not mix up different topics or scopes_).
Overall, at the moment, the `homeassistant_HassGetState` should cover the requirements for getting the state of some entities.

Mattstir:
Hi @mib1185 
where/what discussion do you want to start? 😅 I think a native HAS sensor should definitely be supported...
I based this contribution on https://github.com/home-assistant/intents/discussions/1211#discussioncomment-5267013 

If wanted I can separate this PR into two stages, TBH as mentioned before the EN part is only to enable the tests for the DE part...

mib1185:
Hi @Mattstir 
Sorry, was not aware of the preliminary discussion.
Nevertheless, please split both languages and create an own PR for English first.
When this get merged, than rebase the PR for German, so only the German sentences are left.
thx 👍

frenck:

../Frenck

PR #1221, opened 2023-03-11
[DE] Add scene and script

steffenrapp:


tests/de/scene_HassTurnOn.yaml
```diff
@@ -0,0 +1,13 @@
+language: de
+tests:
+  - sentences:
```
mib1185, line 3: Please add test sets for each possible variant

tests/de/script_HassTurnOn.yaml
```diff
@@ -0,0 +1,13 @@
+language: de
+tests:
+  - sentences:
```
mib1185, line 3: Please add test sets for each possible variant

mib1185 (changes requested):
Hi @steffenrapp 
thanks again for contributing 🙂 
Sure, there are some comments from my side 🙈

sentences/de/scene_HassTurnOn.yaml
```diff
@@ -0,0 +1,12 @@
+language: de
+intents:
+  HassTurnOn:
+    data:
+      - sentences:
+          - "<aktivieren> [<szene>|die] <name> [<szene>]"
+          - "[<szene>|die] <name> [<szene>] (<aktivieren>|<an>)"
```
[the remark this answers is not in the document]
steffenrapp: Alright, I'll split it up but the "die" was there to make sentences like "aktiviere die Filmabend Szene" possible, so I have to add it sometimes.

steffenrapp:
> Hi @steffenrapp 
> 
> thanks again for contributing 🙂 
> 
> Sure, there are some comments from my side 🙈 

Hi, thank you. Hope I got them all 😄

tests/de/scene_HassTurnOn.yaml
mib1185: i count 15 possible variants, but only 7 test sentences 😁

tests/de/script_HassTurnOn.yaml
mib1185: i count 12 possible variants, but only 7 test sentences 😁

mib1185 (changes requested):
we still need more test sentences 🙂

tests/de/scene_HassTurnOn.yaml
steffenrapp: You mean for every combination which each of the expansion rules? Shouldn't it be fine to have tests for each line of the defined sentences?

mib1185: We should have tests for every possible variant, so we can ensure that later changes do not brake already working sentences (_avoid regressions_)

steffenrapp: Alright, makes sense. Found some more 😄

mib1185 (approved):
Thx @steffenrapp 👍

PR #1223, opened 2023-03-12
[NB] Correct hvilke vs hva

svarthvitt:
I have, as per prior dialog, replaces "hva" with "hvilke". "Hva" is used in some sentences without the use of expansion rules.


LaStrada (approved):
As always, great job!

PR #1241, opened 2023-03-14
[EN] Added increase and decrease to set

AalianKhan:
In regards to setting the brightness of a lamp, we can say `increase the brightness to 75%`


TheFes:
This would allow `increase the desk light on`
I would suggest a separate `brightness_set` expansion rule to be used in the relevant sentences.
Maybe even something more general like `value_set` so it can also be used for volume after support for media players is added

AalianKhan:
@TheFes 
good now?

PR #1249, opened 2023-03-18
[DE] exclude cover and lock from homeassistant_HassTurnOff and homeassistant_HassTurnOn

steffenrapp:


mib1185 (changes requested):
Hi @steffenrapp 
there are (_many_) unrelated commits. Please rebase your branch to current `dev` so only the wanted changes are included.
Further I would recommend to have a look into the [docs](https://github.com/home-assistant/intents/blob/main/docs/README.md) 😉

PR #1257, opened 2023-03-22
[en] Add sensor support for power sensor

Mattstir:
Separated languages based on the suggestions in https://github.com/home-assistant/intents/pull/1220


OzGav:
What about also adding support for:
`<what_is> the power consumption [in <area>]`

tetele:
Fixed in #1451

PR #1258, opened 2023-03-23
[NL] Round 3 binary_sensor any/all/which/how many

TheFes:
* presence
* power


TheFes:
Closed for now, will take this on again if I find some time.

PR #1265, opened 2023-04-01
[BG] Initial commit of binary sensors

hristo-atanasov:
Only half of the sentences and tests are translated


hristo-atanasov:
I will make a new PR

PR #1266, opened 2023-04-02
[ar] add set brightness and color

Natfaji:
you can now do the following:

- Brightness value individual light
- Brightness value specific area
- Brightness value all lights

- Min/Max brightness individual light
- Min/Max brightness specific area
- Min/Max brightness all lights

- color individual light
- color specific area
- color all lights


Ahmed-farag36 (approved):
Amazing job, thank you for your time and effort

PR #1286, opened 2023-04-09
[DE] Update to scene_HassTurnOn

steffenrapp:
Makes commands like "Good night" possible.

### DIFF
--- a/sentences/de/scene_HassTurnOn.yaml
+++ b/sentences/de/scene_HassTurnOn.yaml
@@ -3,7 +3,7 @@ intents:
   HassTurnOn:
     data:
       - sentences:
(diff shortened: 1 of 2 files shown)


mib1185 (changes requested):
I'm not sure if it is a good idea, to assume a single word should be enough for a command.
We should have at least an action and object to perform this action on in each possible command

steffenrapp:
@mib1185 In English it also seems to be possible to activate a scene just by using its name. Siri does that with HomeKit scenes, too. To me it seems to be natural.
And as scenes are excluded from general turn on I don't think there would be a conflict, right? 😃

mib1185 (approved):
After some tests I'm convenient that there are no conflicts.
thx @steffenrapp 👍

PR #1287, opened 2023-04-09
[DE] Add all to (un)locking

steffenrapp:


sentences/de/lock_HassTurnOff.yaml
dontinelli, line 1: With the expansion rule of <tuer> and making "alle" optional, one could say "öffne die Tür im Schlafzimmer" and this would match, giving the response "alle entsperrt". For this option the lock-response would be more appropriate. not sure if it make sense to complicate things in order to catch this or if one should live with it for the sake of simplicity...

steffenrapp: I totally agree 😃 It's a lot simpler like that and also the logic is the same now as it is in English. There "all" is also optional and the response doesn't include the name of the door or "all".

easterapps (approved):
Thank you for your contribution.

responses/de/HassTurnOn.yaml
```diff
@@ -8,6 +8,6 @@ responses:
       cover: "{{ slots.name }} geöffnet"
       cover_area: "{{ slots.area }} geöffnet"
       open_all: "alle geöffnet"
-      lock: "{{ slots.name }} abgeschlossen"
+      lock: "abgeschlossen"
```
mib1185, line 11: we should have different responses for targeting a single entity or all
```suggestion
      lock: "{{ slots.name }} abgeschlossen"
      lock_all: "alle abgeschlossen"
```

mib1185 (changes requested):

responses/de/HassTurnOn.yaml
mib1185: hi @steffenrapp please address this comment in a new PR, thx 👍

steffenrapp: Hey @mib1185 do you have a good recommendation? According to the [response style guide](https://developers.home-assistant.io/docs/voice/intent-recognition/style-guide) names of the entities or area should not be repeated so I don't have a good idea right now 🤔

mib1185: we should at least differ between "abgeschlossen" and "alle abgeschlossen". same counts for the other responses

steffenrapp: Have you seen dontinellis review about that topic? It makes a lot of sense to me. It can be more confusing than beneficial to have separate answers in this case. 😕 
Furthermore in English there's also no difference between single and all – probably due to the same reason ...

mib1185: do you have a reference to this review?

steffenrapp: It's inside of this PR. Don't know how to share a link.

mib1185: ah, found it ... ok, let's orient on the English approach - maybe I wanted to be to "German precise" in this case 🙈

steffenrapp: Haha alright 😄

PR #1301, opened 2023-04-13
[cs] Add intents to open/close one cover by its name

kukulich:


responses/cs/HassTurnOff.yaml
```diff
@@ -7,6 +7,6 @@ responses:
       fans_area: "Větráky v {{ slots.area }} byly vypnuty"
       fan: "Větrák {{ slots.name }} vypnut"
       light: "Světlo {{ slots.name }} bylo zhasnuto"
-      cover: "{{ slots.name }} zavřeno"
+      cover: "Zavřeno"
```
halecivo, line 10: Is there a reason for this change? Concept is that assistant will tell which device's status he changed.

sentences/cs/cover_HassTurnOff.yaml
```diff
@@ -11,6 +11,22 @@ intents:
...
         response: garage
 
+      - sentences:
+          - "<zavrit> {name} [<area>]"
+        requires_context:
+          domain: cover
+        response: cover
+
+      - sentences:
+          - "<zatahnout> {name} [<area>]"
```
halecivo, line 21: is there reason to not allow both (zavrit|zatahnout)?

tests/cs/cover_HassTurnOff.yaml
```diff
@@ -1,5 +1,52 @@
 language: cs
 tests:
+  - sentences:
+      - "zavři žaluzie ložnice"
```
halecivo, line 4: I hate here the name of fixture (which was of course not done by you), which makes here confusion between area command and entity name command. I will check on discord.

responses/cs/HassTurnOff.yaml
kukulich: I've changed it because:
- The same is in English
- As a user you see the response just after the request so you know what you asked
- And also it was much easier to make tests work...

sentences/cs/cover_HassTurnOff.yaml
kukulich: `<zavrit>` is allowed above. I've allowed `<zatahnout>` only for blinds etc.

responses/cs/HassTurnOff.yaml
kukulich: You have to write a lot more tests when you want to write tests with inflection for entities names :(

tests/cs/cover_HassTurnOff.yaml
kukulich: Yes, it makes confusion however I have blind named just like that. I've prepared the PR because I was not able to close my blinds in the way I expected it should work.

responses/cs/HassTurnOff.yaml
halecivo: I see, still that can bring troubles when assistant does not understand you correctly. If you ask to turn on a light in bedroom and it turns on light in bathroom, you know what happened. Otherwise it just says "done" and if nothing happened you need to check whole house on what is wrong.

sentences/cs/cover_HassTurnOff.yaml
halecivo: I was wondering why there are two separate sentences.

What device in domain cover, but not with class blind, curtain or shutter should use zavrit and not zatahnout?

responses/cs/HassTurnOff.yaml
kukulich: Force pushed.

sentences/cs/cover_HassTurnOff.yaml
kukulich: window, door, gate ? However we should probably add others like shade or awning. I've just copy the list already used: https://github.com/home-assistant/intents/blob/main/sentences/cs/cover_HassTurnOff.yaml#L17-L21

We can add more device_classes for both `<zatahnout>` in separate PR.

tests/cs/cover_HassTurnOff.yaml
kukulich: Not sure if it's right - now it requires to write sentences etc in the way that both (names and areas) are supported.

sentences/cs/cover_HassTurnOff.yaml
halecivo: Thanks for explanation. Personally I would probably not put them in domain cover, but I see we anyway already have garage_door as cover. So let's go with this now and we can change a bit our fixtures later.

halecivo (approved):

PR #1321, opened 2023-04-24
[ar] add more fixtures to test with

Ahmed-farag36:
* adjust tests with new fixtures
* test with different domains and areas


Natfaji (approved):
👍

PR #1329, opened 2023-04-28
[PL] Added some missing sentences with tests and removed redundant default response keys

Uriziel01:
[PL] Added some missing sentences with tests and removed redundant default response keys


Uriziel01:
@witold-gren Please have a quick look, thanks.

PR #1330, opened 2023-04-30
[fr] add a new expansion rule <sontils>

slegars56:
Might be discussed:

There are multiple interrogative forms of "être" according to singular/plural and gender.

For the time being, I added it only in binary_sensors_HassGetState.

Note:  I also split a few sentences in binary_sensors_HassGetState for more clarity

### DIFF
--- a/sentences/fr/_common.yaml
+++ b/sentences/fr/_common.yaml
@@ -308,9 +308,11 @@ expansion_rules:
   dans: "(dans|du|de)"
   de: "(du|de)"
   yatil: "(y a-t-il|il y a|y a t il|y a t'il|y a-t'il)"
+  sontils: "(a t|est|sont)[-][ ][(il[s]|elle[s])]"
   tous: "(tous|toutes)"
   eclaire: (éclaire|éclairer|illumine|illuminer)
 skip_words:
   - "s'il te plaît"
   - "merci"
   - "est ce que"
+  - "est-ce que"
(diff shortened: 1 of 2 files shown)


flexy2dd (approved):
I think is à good approach

sentences/fr/_common.yaml
```diff
@@ -308,9 +308,11 @@ expansion_rules:
   dans: "(dans|du|de)"
   de: "(du|de)"
   yatil: "(y a-t-il|il y a|y a t il|y a t'il|y a-t'il)"
+  sontils: "(a t|est|sont)[-][ ][(il[s]|elle[s])]"
```
piitaya, line 311: `a-t'il` is different than `est-t'il`. Why do we have the same expansion rule for that ?
Also, can't we use `estil` to have singular form. It's weird to name the expansion rules with plural name as it supports singular and plural.

slegars56: I had in mind "auxiliaires" verbs handling. That's maybe a lazy path, your remark makes sense.
I can split in:
<estil>: "(est|sont)[-][ ][(il[s]|elle[s])]"
and
<atil>: "(a([-| )t]|ont)[-][ ]((il[s]|elle[s])]"

According to Larousse, we should always write "a-t-il" and "est-il" with hyphens but I don't know how speech-to-text solutions deal with hyphens in French. 

That's also why I created <yatil> moreover, we can have "Y a-t-il une fenêtre ouverte ?" or "il y a une fenêtre ouverte ?"  in a more familiar way. 

What do you think?

PR #1331, opened 2023-04-30
[DE] Fix responses for cover

steffenrapp:
Fix responses for cover so that there are no more unused responses and that they all have response checks. Therefore some sentences for covers had to be added as well. Also I removed the repetition of the names in the cover responses.


mib1185 (approved):
LGTM, thx @steffenrapp 👍

PR #1333, opened 2023-04-30
Update light_HassTurnOn.yaml

Pittermaennchen:
Mein Befehl "Mach die Lichterkette an" hat nicht funktioniert. Also habe ich "mach" und die Artikel hinzugefügt. Bitte um Übernahme.

### DIFF
--- a/sentences/de/light_HassTurnOn.yaml
+++ b/sentences/de/light_HassTurnOn.yaml
@@ -3,21 +3,21 @@ intents:
   HassTurnOn:
     data:
       - sentences:
-          - "Schalte (<licht>|<lichter>) <area> <an>"
-          - "Schalte <area> (<licht>|<lichter>) <an>"
+          - "(Schalte|mach) [den|die|das] (<licht>|<lichter>) <area> <an>"
+          - "(Schalte|mach) <area> (<licht>|<lichter>) <an>"
           - "(<licht>|<lichter>) <area> <an>schalten"
           - "<area> (<licht>|<lichter>) <an>schalten"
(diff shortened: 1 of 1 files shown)


steffenrapp (changes requested):
Hi @Pittermaennchen 😃 Unfortunately your changes won't lead to the intended result as you're addressing a `<name>`. So you have to change homeassistant_HassTurnOn and homeassistant_HassTurnOff instead. Also all of your changes are already covered by the `expansion_rules` as they include the relevant articles and "mach".

Pittermaennchen:
> Hi @Pittermaennchen 😃 Unfortunately your changes won't lead to the intended result as you're addressing a `<name>`. So you have to change homeassistant_HassTurnOn and homeassistant_HassTurnOff instead. Also all of your changes are already covered by the `expansion_rules` as they include the relevant articles and "mach".

Thank you for your response!

But why is my Assistant not responding to any of my questions then? The only answer I get is "Entschuldigung, das habe ich nicht verstanden". Shouldn't there be an adequate response and action, if the imperative and articles are already included in the rules?

steffenrapp:
@Pittermaennchen Because you have to extend homeassistant_HassTurnOn and homeassistant_HassTurnOff by "mach[e]" instead 😃 The light files don't recognize the name of your light as they just react to words like "Licht", "Lampe" and so on ... Hope it's clear now.

steffenrapp:
@Pittermaennchen I think you can close this PR as it should now be working with #1373 😃

mib1185:
superseded by #1373

PR #1344, opened 2023-05-03
[EN] Add (switched|turned|set) to `all` question on `on_off_domains`

JamesSwift:
And optional switched|turned|set language

### DIFF
--- a/sentences/en/homeassistant_HassGetState.yaml
+++ b/sentences/en/homeassistant_HassGetState.yaml
@@ -23,7 +23,7 @@ intents:
 
       - sentences:
           - are all [the] {on_off_domains:domain} [(switched|turned|set[ to])] {on_off_states:state} [in <area>]
-          - are all [the] {on_off_domains:domain} [in <area>] [(switched|turned|set[ to])] {on_off_states:state}
+          - are all [the] {on_off_domains:domain} in <area> [(switched|turned|set[ to])] {on_off_states:state}
         response: all
 
       - sentences:
--- a/sentences/en/homeassistant_HassGetState.yaml
+++ b/sentences/en/homeassistant_HassGetState.yaml
@@ -22,7 +22,8 @@ intents:
         response: any
 
       - sentences:
-          - are all [the] {on_off_domains:domain} {on_off_states:state} [in <area>]
+          - are all [the] {on_off_domains:domain} [(switched|turned|set[ to])] {on_off_states:state} [in <area>]
+          - are all [the] {on_off_domains:domain} [in <area>] [(switched|turned|set[ to])] {on_off_states:state}
         response: all
 
       - sentences:
--- a/tests/en/homeassistant_HassGetState.yaml
+++ b/tests/en/homeassistant_HassGetState.yaml
@@ -62,3 +62,15 @@ tests:
         domain: "light"
         state: "on"
     response: "6"
+
+  - sentences:
+      - "are all the lights on in the kitchen"
+      - "are all lights turned on in the kitchen?"
+      - "are all lights in the kitchen on?"
+    intent:
+      name: HassGetState
+      slots:
+        domain: "light"
+        state: "on"
+        area: "Kitchen"
+    response: "Yes"


sentences/en/homeassistant_HassGetState.yaml
```diff
@@ -22,7 +22,8 @@ intents:
         response: any
 
       - sentences:
-          - are all [the] {on_off_domains:domain} {on_off_states:state} [in <area>]
+          - are all [the] {on_off_domains:domain} [(switched|turned|set[ to])] {on_off_states:state} [in <area>]
```
tetele, line 25: Leaving all the optional parts out, the two sentences are the same. I recommend you make at least the `[(switched|turned|set[ to])]` part in one of the sentences mandatory.

tetele (changes requested):
Please also add some tests for these new sentences

sentences/en/homeassistant_HassGetState.yaml
tetele: They both match `are the lights on?`

JamesSwift: Yes, is that a problem? We need two variants to match these two forms:
Are all the lights on in the bedroom? 
Are all the lights in the bedroom on?

tetele: Correct. And my initial suggestion does just that - it matches one of the defined sentence templates to each of your test questions. All the while avoiding ambiguity among which sentence template was matched.

In this specific case, when both sentence templates belong to the same intent, it might not be such big of a problem to have them both matching a sample sentence like the one I mentioned - `are all lights on?`. But if you have many such templates or decide to break them up into multiple intents at some point, you will have a problem as templates from several intents match the same test sentence, thus leaving the intent matcher unable to pick one single intent.

It's good practice to make sure that a test sentence can't match more than one sentence template.

JamesSwift: I'm away from my computer for a few days, and can't test this from my phone. But would the following work to satisfy both in one sentence:
`- are all [the] {on_off_domains:domain} [in <area>] [(switched|turned|set[ to])] {on_off_states:state} [in <area>]`

I'm unsure whether the parser allows multiple `<area>` definitions. I can't forsee any sentence where neither would be omitted.

tetele: I'm pretty sure it would work, but it would match `are all lights in the bedroom on in the bedroom?`. Longer term, the plan is to take all (valid) sentences and train an AI model with them to generate other possible user inputs. As such, it would be best if all sentences made sense.

Is there any particular reason you are reluctant to implement my original proposed change?

JamesSwift: Ah that's interesting to know. Unless I'm misunderstanding you, you want to make it mandatory to say "switched/turned/set" on one sentence. But then if that is the case, if the phrase is omitted, one of the orders of "in area" will fail without it.

tetele: Not if you still leave it optional, as opposed to the "switched/turned/set" part.

JamesSwift: Maybe I've just got a blind spot, but if I follow your suggestion as below:

`- are all [the] {on_off_domains:domain} (switched|turned|set[ to])]{on_off_states:state} [in <area>]
- are all [the] {on_off_domains:domain} [in <area>] [(switched|turned|set[ to])] {on_off_states:state}`
That would fail to match "are all the lights on in the bedroom", wouldn't it?

tetele: You're right, sorry about the confusion.

I've encountered similar situations in the past and I created a misconception in my mind that it doesn't matter which optional part you can make mandatory in one of the sentences.

Would making the `[in <area>]` fix it? I believe so.

JamesSwift: But that would fail to match "are all the lights switched on". I haven't checked, but I assumed that was supported by handler.

JamesSwift:
````
    - are all [the] {on_off_domains:domain} [(switched|turned|set[ to])] {on_off_states:state} [in <area>]
    - are all [the] {on_off_domains:domain} in <area> [(switched|turned|set[ to])] {on_off_states:state}
 ````
Actually you're right, I think the above would work. Am I missing anything?

tetele: Nope, that's OK. Please patch the PR so we can move forward.

JamesSwift: Over I'm back at my computer in a few days I will do. I can't find any documentation on your testing policy. What are you looking for? Coverage seems to be light at the moment.

tetele: > I can't find any documentation on your testing policy

There isn't one. Just add a test sentence with "switched/turned/set", the simplest "are all lights off?" and another one with area. That should be ok.

tetele (changes requested):
Please add a test for the sentence "are all lights off?" (no area).

JamesSwift:
> Please add a test for the sentence "are all lights off?" (no area).

There is already a test for this case:

https://github.com/home-assistant/intents/blob/5669685f733f1aff0a410faa2ff00a8ea07634e4/tests/en/homeassistant_HassGetState.yaml#L41

tetele (approved):

tetele:
Thanks @JamesSwift!

If you feel like it, consider opening another PR for improving the other questions for the same intent. I think that `which lights are turned on in the bedroom` sounds just as good.

PR #1346, opened 2023-05-04
[PL] Simplified scene/script intent sentences, added missing activation variants and tests

Uriziel01:
[PL] Simplified scene/script intent sentences, added missing activation variants and tests


Uriziel01:
@witold-gren Have a look please.

PR #1358, opened 2023-05-08
[fr]split exisitng binary sensors intents and add two expansion rules

slegars56:
Binary sensors:

- Split intent sentences for more clarity and less weird matches
- Add two expansion rules: <quel> and <capteur>

### DIFF
--- a/sentences/fr/_common.yaml
+++ b/sentences/fr/_common.yaml
@@ -319,7 +319,9 @@ expansion_rules:
   yatil: "(y a[-][ ]t[-][']il|il y a)"
   estil: "(est|sont)[-][ ][(il[s]|elle[s])]"
   atil: "(ont|a)[-][ ][t][ ][-][(il[s]|elle[s])]"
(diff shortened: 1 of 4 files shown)


sentences/fr/_common.yaml
```diff
@@ -319,7 +319,9 @@ expansion_rules:
   yatil: "(y a[-][ ]t[-][']il|il y a)"
   estil: "(est|sont)[-][ ][(il[s]|elle[s])]"
   atil: "(ont|a)[-][ ][t][ ][-][(il[s]|elle[s])]"
-  tous: "(tous|toutes)"
+  quel: "quel[le][s]"
+  capteur: "(capteur|appareil|machine|chose|sonde)[s]"
```
bemble, line 323: - `chose` is too generic
- `appareil` is the translation of `device`, not `sensor`
- `machine` is more a synonym of `device`, but I can't find an appropriate sentence where it can be used

PR #1378, opened 2023-05-14
[DE] small improvements to lights

steffenrapp:
Use more expansion rules for the lights and add some more commands including tests.


mib1185 (approved):
LGTM, thx @steffenrapp 👍

PR #1379, opened 2023-05-14
[DE] small improvements to covers

steffenrapp:
Use more expansion rules for the covers and add some more commands including tests.


mib1185 (approved):
LGT, thx @steffenrapp 👍

PR #1381, opened 2023-05-15
[el] Add sentences/test for homeassistant_HassGetState

chatziko:
This PR adds the missing el sentences and tests for homeassistant_HassGetState.

Btw I'm generally trying to be consistent with the current translations of state values. But I noticed that the on/off states for light entities are currently translated as:
- on: Ενεργό
- off: Ανενεργό

I don't think that anyone really says "το φως είναι ενεργό", the usual phrases are:
- το φως είναι ανοικτό / κλειστό
- το φως είναι αναμμένο / σβηστό

Maybe we can try to fix this? Should this change be done in lokalise?

### DIFF
--- a/sentences/el/homeassistant_HassGetState.yaml
+++ b/sentences/el/homeassistant_HassGetState.yaml
@@ -1,4 +1,39 @@
 language: el
 intents:
   HassGetState:
-    data: []
+    data:
+      - sentences:
+          # - (do you know|tell me|<what_is>) [the state of] <name> [in <area>]
+          - (ποι(ός|α|ο|οί|ές|α) είναι) [κατάσταση] <name> [<area>]
+          - τι κατάσταση έχει <name> [<area>]
+          - σε (τι|ποια) κατάσταση (είναι|βρίσκεται) <name> [<area>]
+        response: one
+        excludes_context:
+          domain:
+            - scene
+            - script
+
+      - sentences:
+          - είναι <name> {on_off_states:state} [<area>]
+        response: one_yesno
+        excludes_context:
+          domain:
+            - cover
+
+      - sentences:
+          - <exist> {on_off_domains:domain} {on_off_states:state} [<area>]
+          - <exist> {on_off_states:state} {on_off_domains:domain} [<area>]
+        response: any
+
+      - sentences:
+          - είναι [όλ(οι|ες|α)] (οι|τα) {on_off_domains:domain} {on_off_states:state} [<area>]
+          - είναι [όλ(οι|ες|α)] (οι|τα) {on_off_domains:domain} <area> {on_off_states:state}
+        response: all
+
+      - sentences:
+          - ποι(ος|α|ο|οι|ες) {on_off_domains:domain} [είναι] {on_off_states:state} [<area>]
+        response: which
+
+      - sentences:
+          - πόσ(οι|ες|α) {on_off_domains:domain} [είναι] {on_off_states:state} [<area>]
+        response: how_many
--- a/sentences/el/_common.yaml
+++ b/sentences/el/_common.yaml
@@ -52,9 +52,9 @@ lists:
         out: 1
   on_off_states:
     values:
-      - in: "on"
+      - in: "(ανοι(κ|χ)τ(ός|ή|ό|οί|ές|ά)|αναμμέν(ος|η|ο|οι|ες|α))"
         out: "on"
-      - in: "off"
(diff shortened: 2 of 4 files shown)


dsimop (approved):

dsimop:
> Btw I'm generally trying to be consistent with the current translations of state values. But I noticed that the on/off states for light entities are currently translated as:
> 
> * on: Ενεργό
> * off: Ανενεργό
> 
> I don't think that anyone really says "το φως είναι ενεργό", the usual phrases are:
> 
> * το φως είναι ανοικτό / κλειστό
> * το φως είναι αναμμένο / σβηστό
> 
> Maybe we can try to fix this? Should this change be done in lokalise?

You are right. The "Ενεργό" and "Ανενεργό" states come from Lokalise, as the naming should be general enough to fit to other entities as well. My assumption, though, is that with your commit the intent-taken actions will be the expected ones. That is, the 'on' and 'off' states will be mapped to the your given 'ανοικτό / κλειστό' or 'αναμμένο / σβηστό' state examples, as the values (which are in English) -and not the keys- would be used in the backend for the intent actions.

sentences/el/homeassistant_HassGetState.yaml
```diff
@@ -1,4 +1,39 @@
 language: el
 intents:
   HassGetState:
-    data: []
+    data:
+      - sentences:
+          # - (do you know|tell me|<what_is>) [the state of] <name> [in <area>]
+          - (ποι(ός|α|ο|οί|ές|α) είναι) [κατάσταση] <name> [<area>]
```
dsimop, line 7: Just noticed a minor typo, it should be: (ποι(ος|α|ο|οι|ες). The respective grammar rule defines in that case that these words shall be not emphasised (EL: 'δεν τονίζονται').

chatziko:
> You are right. The "Ενεργό" and "Ανενεργό" states come from Lokalise, as the naming should be general enough to fit to other entities as well. My assumption, though, is that with your commit the intent-taken actions will be the expected ones. That is, the 'on' and 'off' states will be mapped to the your given 'ανοικτό / κλειστό' or 'αναμμένο / σβηστό' state examples, as the values (which are in English) -and not the keys- would be used in the backend for the intent actions.

Indeed, I'm not worried about the intents, I just think that the translation for the state value of lights should probably be fixed.

I understand that "Ενεργό" is chosen as the generic term for arbitrary entities (makes sense), but there should be a way to specify that for `light` entities "on" should have a different translation, right?

dsimop:
I totally understand your point. Unfortunately, this is the way the translations of the words are _published_ on Lokalise. That is, there is no way to have discrete entity-based status namings.

sentences/el/homeassistant_HassGetState.yaml
chatziko: Oh, true, elementary mistake :smile: I'll fix it.

chatziko:
> I totally understand your point. Unfortunately, this is the way the translations of the words are _published_ on Lokalise. That is, there is currently no way to have discrete entity-based status namings.


After some tests I found that there are some languages for which `light` entities have different translation for "on" than generic entities. For instance, changing my language to danish (which I don't speak :smile: ), I get
- "on" for lights: "Tændt"
- "on" for several other entities: "Til"

So I started digging in Lokalise and found the following:
- There is a key `common::state::on` which in english is translated as "On".
- There are also lots of per-entity keys, for instance:
   - `component::light::entity_component::_::state::on`
   - `component::switch::entity_component::_::state::on`
   - `component::fan::entity_component::_::state::on`
- In both english and greek (and many other languages), these keys have value `[%key:common::state::on%]`, which I guess means that they link to the generic work for "on". This is why "on" for all entities is the same.
- However in some languages the per-entity keys have a different translation! For instance in danish you have
  - `common::state::on`: "Til"
  - `component::light::entity_component::_::state::on`: "Tændt"

  Explaining what I observed in my tests.

So I guess we can simply change the greek translation of `component::light::entity_component::_::state::on` to have lights shown as "Ανοικτό", right?

Should I try to change this in lokalise? I have no experience with it, are you also maintaining the greek translations there?

dsimop:
There is no Maintainer per Language in Lokalise, like here. Although, I am an active contributor there. Of course, your contribution will be really helpful there as well. In case I have some free time I will verify that and change it accordingly (if you have not done that by then).

dsimop:
Seems correct 👍

PR #1383, opened 2023-05-16
[DE] improve <irgend> for binary_sensor

steffenrapp:
Improve the `<irgend>` expansion rule for binary sensors by adding more possible commands and making it optional.
All of the changes to the expansion rule are covered by new tests for some of the sensors.
The new expansion rule `<ein>` is now also reused in `<welche>`.


sentences/de/_common.yaml
```diff
@@ -161,9 +161,10 @@ expansion_rules:
   szene: "[die] Szene"
   skript: "[das] Skript"
   batterie: "[die|der|des] (Batterie[n]|Akku[s])"
-  irgend: "(irgend(ein[e][s|r]|welche[s])|einige[s]) [der]"
+  ein: "ein[e|er|es|s]"
+  irgend: "(irgend(ein[e][s|r]|welche[s])|(einige|manche)[s]|<ein>) [der]"
```
mib1185, line 165:
```suggestion
  irgend: "(irgend(<ein>|welche[s])|(einige|manche)[s]|<ein>) [der]"
```

mib1185 (changes requested):
Hi @steffenrapp 
just one small thing, than it is ready to merge 👍

sentences/de/_common.yaml
steffenrapp: Hi @mib1185 I had it that way first but the tests were failing. It seems like it currently doesn't work when you use the same expansion rule twice in one line ... 😞

steffenrapp: I'll set it back to ready for review. I don't have an idea how to restructure it to make it working with just the expansion rule ... I also had the same issue somewhere else.

mib1185: created a fix for this tests #1398 - when this fix is accepted we can rebase the current branch and use `<ein>` multiple times

steffenrapp: @mib1185 how should we proceed? Merge this PR and once there is a solution use `<ein>` multiple times? Functionality doesn't change so maybe that's the best option?

steffenrapp: I requested review again. I'd change the rule when there is a fix available.

steffenrapp:
Hi @mib1185 as #1447 got merged I hope this one should be ready now. Please have another look.

mib1185:
Hi @steffenrapp 
you need to resolve the conflicts and update your branch (_rebase your branch or merge the current main branch back into yours_)

steffenrapp:
@mib1185 I have resolved the conflicts but with the introduction of `<artikel>` in #1419 `<ein>` is somehow redundant here which I don't like. I also have to clean up the usage of `<artikel>` a little to match better with this PR.

sentences/de/_common.yaml
```diff
@@ -170,9 +170,10 @@ expansion_rules:
   szene: "<artikel> Szene"
   skript: "<artikel> Skript"
   batterie: "<artikel> (Batterie[n]|Akku[s])"
-  irgend: "(irgend(ein[e][s|r]|welche[s])|einige[s]) <artikel>"
+  ein: "ein[e|er|es|s]"
+  irgend: "(irgend(<ein>|welche[s])|(einige|manche)[s]|<ein>) <artikel>"
```
mib1185, line 174:
```suggestion
  irgend: "(irgend(<artikel_unbestimmt>[s]|welche[s])|(einige|manche)[s]|<artikel_unbestimmt>[s]) <artikel_bestimmt>"
```

sentences/de/_common.yaml
```diff
@@ -170,9 +170,10 @@ expansion_rules:
...
   etwas: "[irgend][et]was"
-  welche: "(welche[r|s]|was für [ein[e][r|s]]) <artikel>"
+  welche: "(welche[r|s]|was für [<ein>]) <artikel>"
```
mib1185, line 176:
```suggestion
  welche: "(welche[r|s]|was für [<artikel_unbestimmt>[s]]) <artikel_bestimmt>"
```

mib1185 (commented):
you are right - `<ein>` is not needed anymore, because it is now covered by `<artikel_unbestimmt>` (_except `eins`_)

sentences/de/_common.yaml
```diff
@@ -170,9 +170,10 @@ expansion_rules:
   szene: "<artikel> Szene"
   skript: "<artikel> Skript"
   batterie: "<artikel> (Batterie[n]|Akku[s])"
-  irgend: "(irgend(ein[e][s|r]|welche[s])|einige[s]) <artikel>"
+  ein: "ein[e|er|es|s]"
```
mib1185, line 173:
```suggestion
```

mib1185 (approved):
LGTM, thx @steffenrapp 👍

steffenrapp:
Thanks :) But du you think it's a good idea to have both `<artikel_unbestimmt>` and `<artikel_unbestimmt>` generally optional (it is right now)? Could be helpful to have control over this in the sentences themselves? `<artikel>` itself is also optional which includes these - this could be enough?

PR #1384, opened 2023-05-16
[fr] add last intents and tests for binary_sensors

slegars56:
- plug
- power
- presence
- problem
- running
- safety
- smoke
- sound
- tamper
- update
- vibration


flexy2dd (approved):
thank you for your hard work  :)

PR #1408, opened 2023-05-26
[EU] Add Basque language

urtzai:
+ Basque language added with some basic intents
+ Brightness levels added to common file


synesthesiam:
Thanks!

PR #1409, opened 2023-05-27
[DE] Add binary sensors door and garage door

steffenrapp:


mib1185 (changes requested):
Hi @steffenrapp 

sorry for the late review, but during the summer months I am not as active as usual.

These additions looks good, but please extend the tests to cover all open and closed states for both types of doors. For the garage door, I would suggest to use a new area called "front"

thx 👍

steffenrapp:
@mib1185 can you have another look at this one? 😃

mib1185 (approved):
Hi @steffenrapp 
again, sorry for late reply - we are still within the Summer 🙈 
Nevertheless, the additions were already good and the additional tests looks also good now. 👍

PR #1412, opened 2023-05-29
[DE] Add binary sensor for problem

mymyke:
Added the binary Sensor "problem" for German Language


mib1185 (changes requested):
Hi @mymyke 
sorry for late review, but I'm mostly busy with other things during summer, so my time is limited here (_mostly concentrate on bug fixes_).
Nevertheless, it seems you have messed up something, because this PR also adds the connectivity binary sensor?
Please limit the PRs to only at one sensor category at once.
thx 👍

PR #1419, opened 2023-06-10
[de] Generalize German article usages

### LINKED ISSUE #948: [de] How to handle articles (in German)?
In the current German sentences, articles like `der`, `die`, `das` etc. are sometimes optional and sometimes required. If they are required, they sometimes need to be gramatically correct and sometimes all of them are accepted the same way. I guess it makes sense to unify this? Didn't want to change this right ahead but rather discuss it here first.

I feel like including the (correct) articles in the yaml files makes it easier to read. However, from a user perspective I think there is no real benefit in enforcing their usage, is it? Or does anyone have any examples which loose or change their meaning by leaving those articles?

I'm unsure whether we should enforce correct grammar here, but I tend to say this doesn't matter either.

We could just replace all occurrences with an optional expansion rule? Something like
```yml
expansion_rules:
  artikel: "[(der | die | das | den | dem | des)]"
  abdeckung: "<artikel> (Rollo[s] | Abdeckung[en] | Rollladen | Rollläden | Jalousie[n] | Raffstore[s] | Markise[n])"
```

PS:
- I'm actually referring to "bestimmte Artikel" only. Don't see a use case for "unbestimmte Artikel" like `ein`, `eine` etc. yet.
- Similar thoughts were already mentioned in https://github.com/home-assistant/intents/pull/138
- Maybe this discussion could be interesting across all languages and not just German. But I can imagine that there are languages in which the usage of articles could make a more relevant difference.

Johnson145:
Making article usage optional in general along with accepting inputs with wrong grammar.

Closes #948

Some additional notes:

I decided to include the `[...]` part right into the artikel expansion rules to encourage optional article usage whenever possible (instead of leaving it to the user as in `[<artikel>]`). Cases in which an article is actually required will probably be very rare.

A few hardcoded usages remained, because I couldn't replace them without running into an `AssertionError: Recursive rule detected` issue. This is true when using the same placeholder multiple times in the same rule (e.g. I couldn't replace the second `[der]` in `alle: "((alle|sämtliche)[r]|jede[r|s|n]|<artikel> (ganze|komplette|sämtliche)[n]) [der]`). I think I already saw an open issue for this somewhere.

I added a few new tests, but didn't want to scatter incorrect grammar usages all over the place, as those tend to increase confusion when reading.


Johnson145:
@easterapps @mib1185 May I ask, if you're still going to review this PR?

Sorry, I really don't want to be the pain in your ass. I understand that open source work may just need time. If that's the case, that's of course alright. Nevertheless, I'd actually like to propose other PRs, but don't think that makes sense before this one got handled.

easterapps:
> @easterapps @mib1185 May I ask, if you're still going to review this PR?
> 
> Sorry, I really don't want to be the pain in your ass. I understand that open source work may just need time. If that's the case, that's of course alright. Nevertheless, I'd actually like to propose other PRs, but don't think that makes sense before this one got handled.

Sorry for the late review, I will do it today.

easterapps (approved):

PR #1398, opened 2023-05-22
Fix recursive rule detection in tests

mib1185:
the current recursive rule detection also recognize multiple usage of an expansion rule within another as recursion (_see https://github.com/home-assistant/intents/pull/1383#discussion_r1199583989_).

this PR will fix this

---
- defined expansion rules (_`sentences/de/_common.yaml`_)

```yaml
  ein: "ein[e|er|es|s]"
  irgend: "(irgend(<ein>|welche[s])|(einige|manche)[s]|<ein>) [der]"
```

- test result prior this change

```
$ pytest tests/test_language_intents.py --language de

[...]

E           AssertionError: Recursive rule detected: <ein>
E           assert 'ein' not in {'ein', 'irgend'}
E            +  where 'ein' = RuleReference(rule_name='ein').rule_name

tests/test_language_intents.py:161: AssertionError
======================================================================================================================================== short test summary info =========================================================================================================================================
FAILED tests/test_language_intents.py::test_binary_sensor_HassGetState[de] - AssertionError: Recursive rule detected: <ein>
====================================================================================================================================== 1 failed, 19 passed in 0.21s ======================================================================================================================================
```

- test result with this change

```
$ pytest tests/test_language_intents.py --language de

========================================================================================================================================== test session starts ===========================================================================================================================================
platform linux -- Python 3.10.10, pytest-7.2.2, pluggy-1.0.0
rootdir: /workspaces/homeassistant-intents, configfile: pyproject.toml
collected 20 items                                                                                                                                                                                                                                                                                       

tests/test_language_intents.py ....................                                                                                                                                                                                                                                                [100%]

=========================================================================================================================================== 20 passed in 0.20s ===========================================================================================================================================
```

---

regression/negativ test

- defined expansion rules (_`sentences/de/_common.yaml`_)

```yaml
  ein: "ein[e|er|es|s]"
  irgend: "(irgend(<irgend>|welche[s])|(einige|manche)[s]|<ein>) [der]"
```

- test result with this change

```
$ pytest tests/test_language_intents.py --language de

[...]

E               AssertionError: Recursive rule detected: <irgend>
E               assert 'irgend' != 'irgend'
E                +  where 'irgend' = RuleReference(rule_name='irgend').rule_name

tests/test_language_intents.py:164: AssertionError
======================================================================================================================================== short test summary info =========================================================================================================================================
FAILED tests/test_language_intents.py::test_binary_sensor_HassGetState[de] - AssertionError: Recursive rule detected: <irgend>
[...]

PR #1447, opened 2023-07-18
Fix recursive rule detection

tetele:
This PR is an alternative to #1398 which still uses the entire stack of visited rules, only it doesn't "remember" them for each element in a sequence.

When the same rule is used multiple times in a sentence, recursive detection returns a false positive result.

Here's an example:
```
expansion_rules:
  test_sentence: "<rule1> <rule2>"
  rule1: "test"
  rule2: "sentence"
```

The above example works. However, without the fix, the one below results in an error, although it shouldn't:

```
expansion_rules:
  test_sentence: "(<rule1> <rule2> | [<rule2>] <rule1>)"
  rule1: "test"
  rule2: "sentence"
```


mib1185:
@synesthesiam does this makes #1398 obsolete?

tetele:
Yes it does. Get the latest version and test with your test sentences, you'll see that it doesn't fail anymore.

PR #1473, opened 2023-07-29
[DE] Add HassShoppingListAddItem intent

steffenrapp:
As Codespaces had a strange error I had to do some "blind" commits until checks were passing.
Also to pass checks I had to enter the slot item in lower case with a space after it (like it is in English, see #1466).


mib1185 (commented):
LGTM, thx @steffenrapp 👍

PR #1489, opened 2023-08-01
[ca] Add new shopping list, scripts and scenes

duhow:
Adds support for:
- shopping list
- scenes
- scripts


synesthesiam:
Looks good, thanks!

PR #1500, opened 2023-08-10
Fix "all" placeholders for name/area

tetele:
Some sentences hard-coded `area` slots with `"all"` values. This made the intents mismatch unless the user had areas called `all`


tetele:
Should we remove the `all` references from all languages, @synesthesiam?

PR #1506, opened 2023-08-14
pt: add more setence variants

abmantis:
Added a few variants for sentences and fixed some list values that were not translated yet.


nikito7 (approved):

nikito7:
@synesthesiam

PR #1515, opened 2023-09-03
Add English window sentences

vivobg:
Add English window sentences, following the same pattern as currently defined for other binary sensors.

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -283,8 +283,8 @@ lists:
     values:
       - in: "open"
         out: "on"
-      - in: "clear"
-        out: "closed"
+      - in: "closed"
+        out: "off"
 
   shopping_list_item:
     wildcard: true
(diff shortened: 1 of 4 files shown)


tetele:
Thank you, @vivobg, but that will not work, unfortunately. There's a conflict with the `cover` domain (`device_class: window`), same as for doors. You have to have the same sentences for both and there's no differentiator.

The proper solution (work in progress, but slow work) is to have a means of picking targets from multiple `(domain, device_class)` combinations - in this case, `(cover, window)` and `(binary_sensor, window)` so that you can ask `how many windows are open` and it can reply counting devices from both domains. I can assure you the sentences will be updated accordingly after that particular update.

vivobg:
Ah, I wasn't aware of this. Makes sense. @tetele Is there anything I can help with on the proper solution?

tetele:
Not unless you want to implement everything yourself, in core, intents and hassil.

PR #1516, opened 2023-09-03
Fix "all" response grammar and include the state

vivobg:
### Before:
**Query**: "are all windows closed?"
**Answer**: "No, Living Room Window Left and Shed Window not"

**Query**: "are all water sensors wet?"
**Answer**: "No, Kitchen leak sensor not"

### After:
**Query**: "are all windows closed?"
**Answer**: "No, Living Room Window Left and Shed Window are not closed"

**Query**: "are all water sensors wet?"
**Answer**: "No, Kitchen leak sensor is not wet"

### DIFF
--- a/responses/en/HassGetState.yaml
+++ b/responses/en/HassGetState.yaml
@@ -38,13 +38,13 @@ responses:
         {% else %}
           {% set no_match = query.unmatched | map(attribute="name") | sort | list %}
           {% if no_match | length > 4 %}
-            No, {{ no_match[:3] | join(", ") }} and {{ (no_match | length - 3) }} more not
+            No, {{ no_match[:3] | join(", ") }} and {{ (no_match | length - 3) }} more are not {{ slots.state }}
           {%- else -%}
             No,
             {% for name in no_match -%}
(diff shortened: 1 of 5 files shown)


tetele (changes requested):
Please don't mix things in the same PR. As you've seen, I've closed #1515 which would leave this PR hanging. You could leave just the response changes for `all`, although I believe there was a [good reason](https://developers.home-assistant.io/docs/voice/intent-recognition/style-guide#be-concise) for the brevity of the responses.

I will look back and get back on the reasons for avoiding more elaborate answers (such as those you are suggesting), but this PR is definitely not getting merged with the window states as well.

vivobg:
Thank you for the review. I will make the requested changes later today.

tetele (approved):
Looks good to me, with the exception of when a queried entity has a plural name (e.g. `Christmas lights`), in which case the response would be `No, Christmas lights is not on` or something like that.

@synesthesiam is there any reason we wouldn't want to be just a little bit more specific in responses? I've thought about it for a bit and as long as the binary states do not depend on number/gender (as is the case in English), I don't see a reason not to make responses sound better/less robotic in most cases.

tetele:
Thanks @vivobg

Issue #1509, opened 2023-08-20
Can't control the lights of a room in french


thomasd-:
Home assistant version : 2023.8.2

When using assist in french, the phrase "allumer les lumières du salon" is not understood, while the equivalent phrase in enlgish (with voice assistant configured in english) "turn on the lights in salon" is understood and executed as intended.

Playing with the developper tools show some subtle differences in how they are interpreted :

allumer les lumières du salon
Language: français (fr) 
```
intent:
  name: HassTurnOn
slots:
  area: salon
  domain: light
  name: all
details:
  area:
    name: area
    value: salon
    text: salon
  domain:
    name: domain
    value: light
    text: ''
  name:
    name: name
    value: all
    text: ''
targets: {}
```

turn on the lights in salon
Language: anglais (en) 
```
intent:
  name: HassTurnOn
slots:
  area: salon
  domain: light
details:
  area:
    name: area
    value: salon
    text: salon
  domain:
    name: domain
    value: light
    text: ''
targets:
  light.lumieres_ecran:
    matched: true
  light.lumieres_du_canape:
    matched: true
```

As you can see, in french, no target is found for the action while in english my two lights are found.
The only other difference I see is that the french adds a "name: all" entry in the slots.

Thanks

tetele:
Yup, that's the problem.

Could you add a PR to remove the `name: all` from [here](https://github.com/home-assistant/intents/blob/main/sentences/fr/light_HassTurnOn.yaml#L11) and [here](https://github.com/home-assistant/intents/blob/main/sentences/fr/light_HassTurnOff.yaml#L9)?

slegars56:
I tested in French my HA (2023.8.3). It's working fine even though no target is displayed in the Assist Development Tool result. HA correctly turns on the lights that belong to the wanted area.
However, I encountered a few no-match because Whisper STT returned "allumez" instead of "allumer".

PR #1519, opened 2023-09-04
fix control lights of a room

flexy2dd:
Following the issue https://github.com/home-assistant/intents/issues/1509, it seems that under some conditions the sentence in 
french "allumer les lumières du salon" is not recognized.
the problem comes from the setting **name: all** in _light_HassTurnOn.yaml_ and _light_HassTurnOff.yaml_

### DIFF
--- a/tests/fr/light_HassTurnOff.yaml
+++ b/tests/fr/light_HassTurnOff.yaml
@@ -1,6 +1,20 @@
 language: fr
 tests:
   - sentences:
+      - Éteindre les lumières du salon
+      - Éteins la lumière du salon
+      - Éteins les lumières dans le salon
+      - Éteins toutes les lumières dans le salon
+      - Désactiver la lumière du salon
+    intent:
+      name: HassTurnOff
+      slots:
+        area: salon
+        domain: light
+    response: Lumières éteintes
+
+  - sentences:
+      - Éteindre les lumières de la cuisine
       - Éteins la lumière de la cuisine
       - Éteins les lumières dans la cuisine
       - Éteins toutes les lumières dans la cuisine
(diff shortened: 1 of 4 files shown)


tests/fr/light_HassTurnOff.yaml
```diff
@@ -1,6 +1,20 @@
...
+      - Éteins toutes les lumières dans le salon
+      - Désactiver la lumière du salon
+    intent:
+      name: HassTurnOff
+      slots:
+        area: salon
+        domain: light
+    response: Lumières éteintes
+
+  - sentences:
```
piitaya, line 16:
```suggestion
```
We don't need to duplicate sentences.

tests/fr/light_HassTurnOn.yaml
```diff
@@ -1,6 +1,23 @@
...
+      - Lumière dans le salon
+      - Lumière salon
+      - Éclaire le salon
+    intent:
+      name: HassTurnOn
+      slots:
+        area: salon
+        domain: light
+    response: Lumières allumées
+  - sentences:
```
piitaya, line 19: We don't need to duplicate sentences.

piitaya (changes requested):
I agree with the name slot removal. But we don't need to add additional sentences (except the "Eteindre" one). Sentences are only here for tests so we don't need to duplicate them per room.

tests/fr/light_HassTurnOff.yaml
flexy2dd: Ok 
make sense, i have removed unnecessary sentence test

tests/fr/light_HassTurnOn.yaml
flexy2dd: i have removed unnecessary sentence test

piitaya (approved):

PR #1529, opened 2023-09-20
[en] Add alternative sentences for scenes

tjdavey:
Adds alternative sentences starting with `change` and `transition` to activate scenes. Additionally adds several alternatives to allow the activation of scenes with an area slot.

### DIFF
--- a/sentences/en/scene_HassTurnOn.yaml
+++ b/sentences/en/scene_HassTurnOn.yaml
@@ -12,9 +12,9 @@ intents:
         response: scene
       - sentences:
           - "[activate|<turn>] <area> <name> [scene] [on]"
-          - "[activate|<turn>] <name> [scene] (in|on|at) <area>"
+          - "[activate|<turn>] <name> [scene] <in> <area>"
           - "(change|transition) [to] <area> [to] <name> [scene]"
-          - "(change|transition) to <name> [scene] (in|on|at) <area>"
+          - "(change|transition) to <name> [scene] <in> <area>"
         slots:
           domain: scene
         response: scene
--- a/sentences/en/scene_HassTurnOn.yaml
+++ b/sentences/en/scene_HassTurnOn.yaml
@@ -4,8 +4,17 @@ intents:
     data:
       - sentences:
           - "[activate|<turn>] <name> [scene] [on]"
+          - "(change|transition) to <name> [scene]"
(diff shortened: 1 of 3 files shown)


sentences/en/scene_HassTurnOn.yaml
```diff
@@ -4,8 +4,17 @@ intents:
...
           - "[activate|<turn>] <name> [scene] [on]"
+          - "(change|transition) to <name> [scene]"
         requires_context:
           domain: scene
         slots:
           domain: scene
         response: scene
+      - sentences:
+          - "[activate|<turn>] <area> <name> [scene] [on]"
+          - "[activate|<turn>] <name> [scene] (in|on|at) <area>"
```
tetele, line 15: Could you please create a new common expansion rule for `(in|on|at)` called `<in>`? I think that'd be great to use for all intents.

sentences/en/scene_HassTurnOn.yaml
```diff
@@ -4,8 +4,17 @@ intents:
...
+          - "(change|transition) [to] <area> [to] <name> [scene]"
```
tetele, line 16: Why the double `[to]`? Could you please provide an example sentence?

tetele (changes requested):

sentences/en/scene_HassTurnOn.yaml
tjdavey: Yeah, it was structured this way to avoid two nearly identical senteces, I'm happy to break it out into seperate ones or revise it if you feel there is a way to make it clearer though. 

Examples: 
- "change **_to_** the kitchen party mode scene"
- "change the kitchen **_to_** party mode"

tetele:
```suggestion
          - "(change|transition) ([to] <area> <name>|<area> to <name>) [scene]"
```

sentences/en/scene_HassTurnOn.yaml
tjdavey: These changes are complete. I couldn't find any specific tests for expansion rules, but if I need to add some let me know where they are.

tetele (changes requested):

tetele (approved):
Thanks!

PR #1538, opened 2023-10-13
HassGetWeather first draft

tetele:
Since it was mentioned in Chapter 4, I don't see why not :D

### DIFF
--- a/responses/en/HassGetWeather.yaml
+++ b/responses/en/HassGetWeather.yaml
@@ -1,4 +1,4 @@
-anguage: en
+language: en
 responses:
   intents:
     HassGetWeather:
--- a/responses/en/HassGetWeather.yaml
+++ b/responses/en/HassGetWeather.yaml
@@ -1,0 +1,7 @@
+anguage: en
+responses:
+  intents:
+    HassGetWeather:
+      default: "It is now {{ state.attributes.temperature }} and {{ state.state }}"
+
+      forecast: "{{ slots.timeframe | capitalize }} it will be {{ state.attributes.forecast.condition }} with a high of {{ state.attributes.forecast.temperature }} and a low of {{ state.attributes.forecast.templow }}"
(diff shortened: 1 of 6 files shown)


TheFes:
I see you are using the forecast attribute, which is depreciated. So this will not work in 2024.3 anymore

tetele:
I am aware it's deprecated and I've tried another way around, but I haven't been able to do so.

It's not an entity state *per se*, it's whatever data the intent handler in HA decides to send back to compose a response. As such, the handler could pass the forecast as an attribute or as something else, although I'm not sure how to define that data in fixtures and access it in responses

tetele:
Maybe there should be 2 intents: one for the current state and one for the forecast. From an implementation standpoint, they do differ very much.

responses/en/HassGetWeather.yaml
```diff
@@ -0,0 +1,7 @@
+language: en
+responses:
+  intents:
+    HassGetWeather:
+      default: "It is now {{ state.attributes.temperature }} and {{ state.state }}"
+
+      forecast: "{{ slots.timeframe | capitalize }} it will be {{ state.attributes.forecast.condition }} with a high of {{ state.attributes.forecast.temperature }} and a low of {{ state.attributes.forecast.templow }}"
```
balloob, line 7: This is not going to work. I think that for each state of the weather entity we need to get a description going. 

Here are the different states that we need a description for: https://developers.home-assistant.io/docs/core/entity/weather#recommended-values-for-state-and-condition

tetele: For locks, binary sensors et. al. we are getting a translated state string from the core. Why does this need to be different?

balloob: I don't think that we can get a translation that makes a grammatically perfect sentence for each condition. It would be nicer if we can have perfect sentences.

Condition | Description
-- | --
clear-night | Clear night
cloudy | Many clouds
exceptional | Exceptional
fog | Fog
hail | Hail
lightning | Lightning/ thunderstorms
lightning-rainy | Lightning/ thunderstorms and rain
partlycloudy | A few clouds
pouring | Pouring rain
rainy | Rain
snowy | Snow
snowy-rainy | Snow and Rain
sunny | Sunshine
windy | Wind
windy-variant | Wind and clouds

PR #1539, opened 2023-10-13
[en] Add definite article for object of HassTurnOn/HassTurnOff responses

HarvsG:
To me at least, it sounds much more natural with a "the" I.e "Turned on the light(s)"
Otherwise "lights turned on" sounds more natural than "turned on lights"

Side note:

>   Google's solution is answering in the present tense:
>   "OK Google turn on the kitchen lights" : "Sure, turning 4 lights on"
>   Or "OK Google, turn on the hallway lights" : "Sure, turning the hallway lights on"
>   
>   To me that sounds more natural, however there is a subtle difference, google's solution implies an attempt, but not that the action has been successful, Assist's response implies success.

In the meantime "the" is added.


tetele (changes requested):
Fair enough, but you need to alter the expected responses in all tests, too.

HarvsG:
> Fair enough, but you need to alter the expected responses in all tests, too.

Tests updated

tetele (changes requested):
Please fix the failing tests

tetele (approved):
Thanks for this!

PR #1543, opened 2023-10-15
Add script intents for daily routines

sanyatuning:
Related MR: https://github.com/home-assistant/core/pull/102035


tetele (commented):
Like I said in the core PR, I don't see these as being generic enough, applicable to every user. They are just custom intent/sentence material.

However, we will see if the parent PR gets merged. If so, then I will gladly follow suite in the intents repo.

PR #1549, opened 2023-10-16
Add HassNevermind (not yet supported in HA)

synesthesiam:
Add English support for a future HA intent: `HassNevermind`.
This intent does nothing, it's just used to be able to cancel a request by saying something like "nevermind".

Because this intent returns no slot values, some test code needed to be a bit more robust.


tetele (approved):
Brilliant. Can't wait for conversations

PR #1558, opened 2023-10-18
[si] Added sentences for Slovenian language

andrejs2:
Added sentences for Slovenian language for ClimateGetTemperature


andrejs2 (commented):
deletation of line 5

PR #1571, opened 2023-10-20
[EL] Add HassNevermind 

apo-mak:
Add Greek support for a future HA intent: HassNevermind (see https://github.com/home-assistant/intents/pull/1549).


sentences/el/homeassistant_HassNevermind.yaml
```diff
@@ -0,0 +1,8 @@
+language: el
+intents:
+  HassNevermind:
+    data:
+      - sentences:
+          - "άστο"
+          - "ξέχνατο"
```
dsimop, line 7: Please apply this minor fix: "ξέχνα το"

tests/el/homeassistant_HassNevermind.yaml
```diff
@@ -0,0 +1,8 @@
+language: el
+tests:
+  - sentences:
+      - "άστο"
+      - "ξέχνατο"
```
dsimop, line 5: Please apply this minor fix: "ξέχνα το"

dsimop (changes requested):
Please apply this minor fix: "ξέχνα το" to the commented lines.

apo-mak:
> Please apply this minor fix: "ξέχνα το" to the commented lines.

I tried it with google TTS and it does a strange pause when I split the two.
I do not know how the STT would prefer to have it, but we also tend to say it almost like one word.
Should we put both in for better accuracy?


(I can make a new pull after this , with out the split.
This will help reverting it if necessary later.)

dsimop (approved):

dsimop:
Hi @apo-mak, sure you can add it as one word as well.

PR #1577, opened 2023-10-21
[DE] Add binary sensor gas

steffenrapp:


sentences/de/_common.yaml
```diff
@@ -186,12 +192,13 @@ expansion_rules:
...
   skript: "<artikel> Skript"
   batterie: "<artikel> (Batterie[n]|Akku[s])"
-  irgend: "(irgend(<artikel_unbestimmt>[s]|welche[s])|(einige|manche)[s]|<artikel_unbestimmt>[s]) <artikel_bestimmt>"
+  irgend: "(irgend(<artikel_unbestimmt>[s]|welche[s]|wo)|(einige|manche)[s]|<artikel_unbestimmt>[s]) <artikel_bestimmt>"
   etwas: "[irgend][et]was"
   welche: "(welche[r|s]|was für [<artikel_unbestimmt>[s]]) <artikel_bestimmt>"
   ladestand: "<artikel> [Lade][zu]Stand"
   co: "Kohlen[stoff]monoxid"
   co-sensor: "<co>[-]Sensor[en]"
+  gas-sensor: "Gas[-]Sensor[en]"
```
mib1185, line 201: i think we should go with the `lower_case_with_underscores` convention (_yeah, `co-sensor` should also be corrected, maybe in a different PR_)

```suggestion
  gas_sensor: "Gas[-]Sensor[en]"
```

mib1185 (changes requested):
Hi @steffenrapp 
as always, thanks for this addition 👍 
overall it looks good, just one small comment about naming convention

sentences/de/_common.yaml
steffenrapp: Yes makes sense. I'll address that for co-sensor in a different PR.

steffenrapp:
Hi @mib1185 thanks for the review ✌️ Naming conventions are applied and conflicts resolved.

mib1185 (approved):
LGTM, thx 👍

PR #1590, opened 2023-10-25
[sv] possibility to turn on and off individual lights

Gustavhol:
Added possibility to turn on and off individual lights in a given area in swedish.

### DIFF
--- a/tests/sv/light_HassTurnOff.yaml
+++ b/tests/sv/light_HassTurnOff.yaml
@@ -17,7 +17,7 @@ tests:
       slots:
         area: Kök
         domain: light
-    response: "Slog av"
+    response: "Släckte lamporna i kök"
 
   - sentences:
       - släck lampan bänkbelysningen
@@ -27,7 +27,7 @@ tests:
       slots:
         name: Bänkbelysning
         domain: light
-    response: "Slog av bänkbelysning"
+    response: "Släckte bänkbelysning"
   - sentences:
       - slå av bänkbelysningen i köket
       - släck bänkbelysningen i köket
@@ -38,7 +38,7 @@ tests:
         name: Bänkbelysning
         domain: light
         area: Kök
-    response: "Slog av bänkbelysning"
+    response: "Släckte bänkbelysning"
   - sentences:
       - släck vardagsrummets taklampa
       - släck taklampan i vardagsrummet
@@ -48,4 +48,4 @@ tests:
         name: Taklampa
         domain: light
         area: Vardagsrum
-    response: "Slog av taklampa"
+    response: "Släckte taklampa"
--- a/tests/sv/light_HassTurnOff.yaml
+++ b/tests/sv/light_HassTurnOff.yaml
@@ -17,6 +17,7 @@ tests:
       slots:
         area: Kök
         domain: light
+    response: "Slog av"
 
   - sentences:
       - släck lampan bänkbelysningen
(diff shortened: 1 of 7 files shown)


tests/sv/light_HassTurnOn.yaml
```diff
@@ -21,6 +21,7 @@ tests:
       slots:
         area: Kök
         domain: light
+    response: "Slog på"
```
larsdunemark, line 24: I'm not sure if "Slog på" is a correct response to turning on an areas lights. 
I think it would be:
`Påslaget` Eller `Tänt`

tests/sv/light_HassTurnOff.yaml
```diff
@@ -26,3 +27,25 @@ tests:
...
+  - sentences:
+      - släck vardagsrummets taklampa
+      - släck taklampan i vardagsrummet
+    intent:
+      name: HassTurnOff
+      slots:
+        name: Taklampa
+        domain: light
+        area: Vardagsrum
+    response: "Slog av taklampa"
```
larsdunemark, line 51: Should probably be `Släckte taklampan`

tests/sv/light_HassTurnOn.yaml
```diff
@@ -29,3 +30,25 @@ tests:
...
+      - slå på bänkbelysningen i köket
+      - tänd bänkbelysningen i köket
+      - starta kökets bänkbelysning
+    intent:
+      name: HassTurnOn
+      slots:
+        name: Bänkbelysning
+        domain: light
+        area: Kök
+    response: "Slog på bänkbelysning"
```
larsdunemark, line 44: `Bänkbelysningen är tänd` or `Tänt bänkbelysningen`

tests/sv/light_HassTurnOn.yaml
```diff
@@ -29,3 +30,25 @@ tests:
...
+  - sentences:
+      - tänd vardagsrummets taklampa
+      - tänd taklampan i vardagsrummet
+    intent:
+      name: HassTurnOn
+      slots:
+        name: Taklampa
+        domain: light
+        area: Vardagsrum
+    response: "Slog på taklampa"
```
larsdunemark, line 54: `Tänt taklampa`

larsdunemark (commented):
Thanks for the PR, added some feedback on the response.
What do you think about it?

tests/sv/light_HassTurnOn.yaml
Gustavhol: I agree. But since the default response for the light domain seemed to be "Slog på [...]", i just went with that. I wanted to keep my first PR small. If i change the response in the test the test will fail, so i decided to keep the default response.

larsdunemark: Ok, I think it just to add
```
      lights: "Tände {{ slots.name }}"
```
Or:
```
      lights: "Tände lampan"
```
To https://github.com/home-assistant/intents/blob/main/responses/sv/HassTurnOn.yaml
 
Since `lights_area` is already defined.
And the same for HassTurnOff.yaml

Don't think it will make the PR to big for this case.

larsdunemark:
Do you also update the testcase responses and verify that the tests pass.

Gustavhol:
> Do you also update the testcase responses and verify that the tests pass.

Yeah. I thought i'd just make a quick edit in the web ui, but family duty called after the first two files. I'll get to it later.

PR #1593, opened 2023-10-26
[cs] Add scene support

johny-mnemonic:
Add scene support for Czech language


johny-mnemonic:
All tests finished successfully in my local checkout, so hopefully it should work 🤞

PR #1613, opened 2023-11-03
[en] Added HassGetState for the person domain

JosephAbbey:
Added some sentences and tests for getting the state of a person (english language).

### DIFF
--- a/sentences/en/person_HassGetState.yaml
+++ b/sentences/en/person_HassGetState.yaml
@@ -1,0 +1,44 @@
+language: en
+intents:
+  HassGetState:
+    data:
+      # https://www.home-assistant.io/integrations/person/
+      - sentences:
+          - "<where_is> <name>"
+        response: where
+        requires_context:
+          domain: person
+        slots:
+          domain: person
+
+      - sentences:
+          - "is <name> at {zone:state}"
+        response: one_yesno
+        requires_context:
+          domain: person
+        slots:
+          domain: person
+
+      - sentences:
+          - "is anyone at {zone:state}"
+        response: any
+        slots:
+          domain: person
+
+      - sentences:
+          - "is everyone at {zone:state}"
+        response: all
+        slots:
+          domain: person
+
+      - sentences:
+          - "who is at {zone:state}"
+        response: which
+        slots:
+          domain: person
+
+      - sentences:
+          - "how many people are at {zone:state}"
+        response: how_many
+        slots:
+          domain: person
--- a/responses/en/HassGetState.yaml
+++ b/responses/en/HassGetState.yaml
@@ -65,3 +65,10 @@ responses:
 
       how_many: |
         {{ query.matched | length }}
+
+      where: |
+        {% if state in ("away", "not_home") %}
+          {{ slots.name | capitalize }} is away
+        {% else %}
+          {{ slots.name | capitalize }} is at {{ state.state }}
+        {% endif %}
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -289,10 +289,14 @@ lists:
   shopping_list_item:
     wildcard: true
 
+  zone:
+    wildcard: true
+
 expansion_rules:
   name: "[the] {name}"
   area: "[the] {area}"
   what_is: "(what's|whats|what is)"
+  where_is: "(where's|wheres|where is)"
   brightness: "{brightness}[ ][%|percent]"
   light: "[the] (light|lights|lighting)"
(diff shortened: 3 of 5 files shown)


sentences/en/person_HassGetState.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - sentences:
+          - "<where_is> <name>"
+        response: where
+        requires_context:
+          domain: person
+        slots:
+          domain: person
+
+      - sentences:
+          - "is <name> at {zone:state}"
```
tetele, line 15: Can you make `at` optional? I think we should be able to ask "is Bob home?"

sentences/en/person_HassGetState.yaml
```diff
@@ -0,0 +1,44 @@
...
+        response: one_yesno
+        requires_context:
+          domain: person
+        slots:
+          domain: person
+
+      - sentences:
+          - "is anyone at {zone:state}"
```
tetele, line 23: Same thing, `at` optional

sentences/en/person_HassGetState.yaml
```diff
@@ -0,0 +1,44 @@
...
+        response: any
+        slots:
+          domain: person
+
+      - sentences:
+          - "is everyone at {zone:state}"
```
tetele, line 29: And again

sentences/en/person_HassGetState.yaml
```diff
@@ -0,0 +1,44 @@
...
+        response: all
+        slots:
+          domain: person
+
+      - sentences:
+          - "who is at {zone:state}"
```
tetele, line 35: And *again*

sentences/en/person_HassGetState.yaml
```diff
@@ -0,0 +1,44 @@
...
+        response: which
+        slots:
+          domain: person
+
+      - sentences:
+          - "how many people are at {zone:state}"
```
tetele, line 41: And finally :)

tetele (changes requested):
This is a really nice addition. Just please address the `at`.

sentences/en/person_HassGetState.yaml
tetele: Maybe also add a `[the]`. E.g "Is Bob at the kindergarden?"

tetele (changes requested):

sentences/en/person_HassGetState.yaml
JosephAbbey: I could make the at optional but I would like to make the answer just use the same preposition as the question, I think this should be possible with lists but I can't see how.

JosephAbbey:
Ok I have made `at` optional but I need a better way to format the output (with or without `at`), I think it would be best to just use what is used in the question, but I don't know how.

tetele:
Nice, nice, nice, thank you!

Regarding your very valid issue with including the inclusion of the preposition in the slot, there's nothing you can do now, but I think we should support list items that can contain list items in the future. Something like

```yaml
lists:
  zone:
    wildcard: true
  zone_with_preposition:
    template: "<in> {zone}"
```

@synesthesiam could you give some feedback on both this PR and the idea above?

JosephAbbey:
> ```yaml
>   zone_with_preposition:
>     template: "<in> {zone}"
> ```

This would be perfect!

sentences/en/person_HassGetState.yaml
synesthesiam: Keeping the preposition would require extending the `HassGetState` event to accept a new slot, which I'd prefer not to do.  If you want to keep the same preposition, I'd suggest instead creating separate responses for each case: `where_at`, `where_in`, etc.

tetele:
That does not populate a slot, unfortunately, which was the whole point.

synesthesiam:
Right now, any text that gets captured into a slot has to have the slot defined for the intent itself in HA (even if it's only going to be used in the response). It would be worth considering having a way to capture text that is only intended for use in the response and not as typical slot meant for influencing the intent handling.

sentences/en/person_HassGetState.yaml
tetele: @JosephAbbey are you willing to split it into separate sentences and responses like Mike suggested or should I merge it as it is?

JosephAbbey: Currently I am only happy with the `where is...` sentence because of the `at` problem, so I will create a separate pr with just that one

tetele (approved):

PR #1615, opened 2023-11-03
[en] Allow singular use of lock or door for lock and unlock

ViViDboarder:
Some more sentence variety. Added test sentences as examples.


tetele:
Thanks!

PR #1617, opened 2023-11-05
[DE] Change naming of co-sensor

steffenrapp:
Change naming of co-sensor to co_sensor as requested in #1577


mib1185 (approved):
LGTM, thx @steffenrapp 👍

PR #1626, opened 2023-11-09
Add HassListAddItem intent

synesthesiam:
Adds a new intent: `HassListAddItem` that lets you add something to a to-do list by name ("add beer to my Trader Joe's list").

### DIFF
--- a/responses/en/HassListAddItem.yaml
+++ b/responses/en/HassListAddItem.yaml
@@ -1,0 +1,5 @@
+language: "en"
+responses:
+  intents:
+    HassListAddItem:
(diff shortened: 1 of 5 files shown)


responses/en/HassListAddItem.yaml
```diff
@@ -0,0 +1,5 @@
+language: "en"
+responses:
+  intents:
+    HassListAddItem:
+      item_added: "Added {{ slots.item }} to {{ slots.name }}"
```
tetele, line 5: Maybe this is a bit too much, according to the [style guide](https://developers.home-assistant.io/docs/voice/intent-recognition/style-guide)? I'd shorten it to `"Added {{slots.item}}"`

PR #1627, opened 2023-11-09
Add HassGetWeather intent

synesthesiam:
Adds a new intent: `HassGetWeather`. Credit to @tetele for the first draft: https://github.com/home-assistant/intents/pull/1538

In the current implementation, only the current weather is reported; either for a weather entity by name or the first weather entity found. Future revisions will add more complexity, such as forecasts.


sentences/en/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,10 @@
+language: en
+intents:
+  HassGetWeather:
+    data:
+      - sentences:
+          - "<what_is>[ the] weather[ like][( for| in| at) <name>]"
+          - "<what_is>[ the] weather[( for| in| at) <name>][ like]"
```
tetele, line 7: You can't have only permutations of all optional components, as missing them all would make the 2 sentences identical and, thus, ambiguous for the recognizer.
```suggestion
          - "<what_is>[ the] weather[( for| in| at) <name>] like"
```

tetele: I mean, you can, but you also said you shouldn't 😅

synesthesiam: Ah, good point :smile:

PR #1630, opened 2023-11-10
[en] Don't produce <N>percent

synesthesiam:
Fix template to avoid incorrect sentences like "change X brightness to Npercent" without a space between `N` and percent.


tetele (approved):
LGTM

PR #1631, opened 2023-11-10
[en] Added where is person

JosephAbbey:
From https://github.com/home-assistant/intents/pull/1613.


tetele:
I don't understand. What's this PR for?

I absolutely love #1613. The only [question](https://github.com/home-assistant/intents/pull/1613#discussion_r1389485312) there was if you want to split the sentences to keep track of preposition.

However, the more I think about it, the more I realize that the only application of including the preposition is the negative branch of the `all`/"everybody" question.

PR #1632, opened 2023-11-10
[en] Add sentences for turning lights on/off in the area of a satellite device

synesthesiam:
Adds English sentences that will turn lights on/off in the area of a satellite device. This uses:

```yaml
requires_context:
    area: null
```

to force Home Assistant to supply an area before the sentences will match. A future PR for HA Core will include the area from a satellite device as `intent_context` during recognition.

### DIFF
--- a/sentences/en/light_HassTurnOff.yaml
+++ b/sentences/en/light_HassTurnOff.yaml
@@ -28,3 +28,13 @@ intents:
         slots:
           domain: "light"
           name: "all"
+
+      # Turn off all lights in the same area as a satellite device
+      - sentences:
+          - "<turn> off all[ the| my] lights"
+          - "<turn> all[ the| my] lights off"
+        response: "lights_area"
+        slots:
+          domain: "light"
+        requires_context:
+          area: null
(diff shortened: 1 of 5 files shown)


sentences/en/light_HassTurnOff.yaml
```diff
@@ -28,3 +28,13 @@ intents:
         slots:
           domain: "light"
           name: "all"
+
+      # Turn off all lights in the same area as a satellite device
+      - sentences:
+          - "<turn> off all[ the| my] lights"
```
tetele, line 34: I feel as though "my lights" refers to all of them, not just in the area.

I'd also add an optional `[ here]` at the end

synesthesiam: I was imagining "turn off my lights" being used in the sense of "turn off the lights in **my** office"

tetele: I won't argue with that, it makes sense in a lot of contexts. If the case may be, we will receive issues and we can improve after the fact.

But i feel like `[ [in ]here]` at the end would be a good addition.

PR #1646, opened 2023-11-14
[DE] HassGetWeather

steffenrapp:
German translation of #1627

### DIFF
--- a/sentences/de/weather_HassGetWeather.yaml
+++ b/sentences/de/weather_HassGetWeather.yaml
@@ -4,7 +4,7 @@ intents:
     data:
       - sentences:
           - "wie ist [das] Wetter [(für|in) <name>]"
-          - "wie ist [das] <name>[er] Wetter"
+          - "wie ist [das] <name>[er|s] Wetter"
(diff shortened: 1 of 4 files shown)


sentences/de/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,10 @@
+language: de
+intents:
+  HassGetWeather:
+    data:
+      - sentences:
+          - "wie ist [das] Wetter [(für|in) <name>]"
+          - "wie ist [das] <name> Wetter"
```
mib1185, line 7:
```suggestion
          - "wie ist [das] <name>[er] Wetter"
```

what do you think about this addition? it should make "Wie ist das Berliner Wetter?" possible 🤔

steffenrapp: I like that idea! 😃

steffenrapp: I also added your test sentence

mib1185 (approved):
LGTM, thx @steffenrapp 👍

PR #1657, opened 2023-11-19
[DE] add support for person_HassGetState

steffenrapp:
Can get some more fine tuning but it's a start.

### DIFF
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -61,3 +61,10 @@ responses:
 
       wie_viele: |
         {{ query.matched | length }}
+
+      wo: |
+        {% if state == "not_home" %}
+          {{ slots.name | capitalize }} ist weg
+        {% else %}
+          {{ slots.name | capitalize }} ist hier: {{ state.state }}
+        {% endif %}
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -15,8 +15,6 @@ responses:
       einzeln_zone: |
         {% if query.matched %}
           Ja
-        {% elif state.state == 'not_home' %}
-          Nein, ist nicht zu Hause
         {% elif state.state == 'home' %}
           Nein, ist zu Hause
         {% else %}
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -12,6 +12,17 @@ responses:
           Nein
         {% endif %}
 
+      einzeln_zone: |
+        {% if query.matched %}
+          Ja
+        {% elif state.state == 'not_home' %}
+          Nein, ist nicht zu Hause
+        {% elif state.state == 'home' %}
+          Nein, ist zu Hause
+        {% else %}
+          Nein, ist im Bereich {{ state.state_with_unit }}
+        {% endif %}
+
       irgendeins: |
         {% if query.matched %}
           {% set match = query.matched | map(attribute="name") | sort | list %}
@@ -66,7 +77,7 @@ responses:
         {% if state.state == "not_home" %}
           {{ slots.name | capitalize }} ist weg
         {% elif state.state == "home" %}
-          {{ slots.name | capitalize }} ist zuhause
+          {{ slots.name | capitalize }} ist zu Hause
         {% else %}
           {{ slots.name | capitalize }} ist im Bereich {{ state.state }}
         {% endif %}
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -63,8 +63,10 @@ responses:
         {{ query.matched | length }}
 
       wo: |
-        {% if state == "not_home" %}
-          {{ slots.name | capitalize }} ist weg
-        {% else %}
-          {{ slots.name | capitalize }} ist hier: {{ state.state }}
-        {% endif %}
+         {% if state.state == "not_home" %} 
+           {{ slots.name | capitalize }} ist weg
+         {% elif state.state == "home" %} 
+           {{ slots.name | capitalize }} ist zuhause 
+         {% else %} 
+           {{ slots.name | capitalize }} ist im Bereich {{ state.state }} 
+         {% endif %} 
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -63,10 +63,10 @@ responses:
         {{ query.matched | length }}
 
       wo: |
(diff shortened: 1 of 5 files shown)


responses/de/HassGetState.yaml
```diff
@@ -61,3 +61,10 @@ responses:
 
       wie_viele: |
         {{ query.matched | length }}
+
+      wo: |
+        {% if state == "not_home" %}
+          {{ slots.name | capitalize }} ist weg
+        {% else %}
+          {{ slots.name | capitalize }} ist hier: {{ state.state }}
+        {% endif %}
```
mib1185, line 70:
```suggestion
      wo: |
         {% if state.state == "not_home" %} 
           {{ slots.name | capitalize }} ist weg
         {% elif state.state == "home" %} 
           {{ slots.name | capitalize }} ist zuhause 
         {% else %} 
           {{ slots.name | capitalize }} ist im Bereich {{ state.state }} 
         {% endif %} 
```

i think the Dutch approach fits better in this case
https://github.com/home-assistant/intents/blob/09ad6c6abbc04671d776bd8e44f1bf5b3eed3280/responses/nl/HassGetState.yaml#L83-L90

mib1185 (changes requested):
Hi @steffenrapp 
as always, thanks for contribution 👍 
I think we should adopt the Dutch approach here

tests/de/person_HassGetState.yaml
```diff
@@ -0,0 +1,86 @@
...
+tests:
+  - sentences:
+      - "wo ist Martin"
+      - "wo befindet sich Martin"
+    intent:
+      name: HassGetState
+      slots:
+        domain: person
+        name: Martin
+    response: "Martin ist hier: zuhause"
```
mib1185, line 11:
```suggestion
    response: "Martin ist zuhause"
```

tests/de/person_HassGetState.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+  - sentences:
+      - "wo ist Lisa"
+      - "wo befindet sich Lisa"
+    intent:
+      name: HassGetState
+      slots:
+        domain: person
+        name: Lisa
+    response: "Lisa ist hier: Arbeit"
```
mib1185, line 21:
```suggestion
    response: "Lisa ist im Bereich Arbeit"
```

mib1185 (changes requested):

responses/de/HassGetState.yaml
```diff
@@ -61,3 +61,12 @@ responses:
...
         {{ query.matched | length }}
+
+      wo: |
+         {% if state.state == "not_home" %} 
+           {{ slots.name | capitalize }} ist weg
+         {% elif state.state == "home" %} 
+           {{ slots.name | capitalize }} ist zuhause 
+         {% else %} 
+           {{ slots.name | capitalize }} ist im Bereich {{ state.state }} 
+         {% endif %}
```
mib1185, line 72:
```suggestion
      wo: |
        {% if state.state == "not_home" %}
          {{ slots.name | capitalize }} ist weg
        {% elif state.state == "home" %}
          {{ slots.name | capitalize }} ist zuhause
        {% else %}
          {{ slots.name | capitalize }} ist im Bereich {{ state.state }}
        {% endif %}
```

mib1185 (approved):

tests/de/_fixtures.yaml
```diff
@@ -192,3 +192,13 @@ entities:
     attributes:
       temperature: "25"
       temperature_unit: "°C"
+
+  - name: "Martin"
+    id: "person.martin"
+    state: "zuhause"
```
mib1185, line 198:
```suggestion
    state: "home"
```

steffenrapp:
@mib1185 thanks for the review and the contributions. I'm working on fixing the errors. Have to rebuild some of the logic.

steffenrapp:
@mib1185 I adapted the logic and added more tests. Hope it's fine, maybe I can do a little bit more refinement in a separate PR afterwards. Also I want to use the new expansion rule <wieviel> everywhere else where it fits.

sentences/de/person_HassGetState.yaml
```diff
@@ -0,0 +1,121 @@
...
+  HassGetState:
+    data:
+      # https://www.home-assistant.io/integrations/person/
+      - sentences:
+          - "<wo_ist> <name>"
+        response: wo
+        requires_context:
+          domain: person
+        slots:
+          domain: person
```
mib1185, line 12: i think we do not need to set the `domain` here, because above we already require the domain to be `person` 🤔

sentences/de/person_HassGetState.yaml
```diff
@@ -0,0 +1,121 @@
...
+          state: not_home
+
+      - sentences:
+          - "wer <befinden> [<lokale_praeposition>] {zone:state}"
+        response: welches
+        slots:
+          domain: person
+
+      - sentences:
+          - "<wieviel> (Personen|Menschen) <befinden> <zuhause>"
```
mib1185, line 104: i think we can make `(Personen|Menschen)` an expansion rule and add `Leute` to possible values 🙂

responses/de/HassGetState.yaml
```diff
@@ -12,6 +12,17 @@ responses:
           Nein
         {% endif %}
 
+      einzeln_zone: |
```
mib1185, line 15: i think this is to complex and may lead to confusing replies like:
question: Ist Jonas auf Arbeit?
answer: Nein, ist nicht zu Hause

so we should just reply with ja/nein. if the user want's to know, where a person is, the user can ask for it ^^

mib1185 (changes requested):
Hi @steffenrapp there are some comments from my side, thx 👍

sentences/de/person_HassGetState.yaml
mib1185: this should also count for the sentences below 😉

mib1185: after some testing it figured out, that setting `slot.domain` is still needed, even we already require domain to be person 🤔

mib1185:
Hi @steffenrapp 
based on https://github.com/home-assistant/intents/discussions/1768#discussioncomment-7835392 this branch has been rebased on [untranslated-state](https://github.com/home-assistant/intents/tree/untranslated-state) and is going to be merged there.

mib1185 (approved):

PR #1658, opened 2023-11-20
[NL] ToDo add item intent

TheFes:
SSIA

### DIFF
--- a/responses/nl/HassListAddItem.yaml
+++ b/responses/nl/HassListAddItem.yaml
@@ -2,4 +2,4 @@ language: "nl"
 responses:
   intents:
     HassListAddItem:
-      item_added: "{{ slots.item | capitalize }} toegevoegd"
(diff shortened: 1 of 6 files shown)


responses/nl/HassListAddItem.yaml
```diff
@@ -0,0 +1,5 @@
+language: "nl"
+responses:
+  intents:
+    HassListAddItem:
+      item_added: "{{ slots.item | capitalize }} toegevoegd"
```
frenck, line 5: Is this actually correct to filter? It will mangle the input? (including abbreviations)

TheFes: Didn't think of abbreviations, I have removed the filter (also for the shopping list intent)

frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #1660, opened 2023-11-20
[NL] Turn lights on/off in the area of a satellite

TheFes:


TheFes:
@synesthesiam any idea why the recognition fails? Did I miss something?

PR #1668, opened 2023-11-24
[EL] Fix default response to HassTurnOn/HassTurnOff

paravoid:
The current default response is "Άνοιξαν τα φώτα στο {{ slots.name }}" ("Lights were turned on in {{ slots.name }}") resulting in nonsensical responses like "Άνοιξα τα φώτα στο καφετιέρα" ("Lights were turned on in coffee machine").

Change this to "Άνοιξε", to mimic what most other languages do.

### DIFF
--- a/responses/el/HassTurnOff.yaml
+++ b/responses/el/HassTurnOff.yaml
@@ -2,7 +2,7 @@ language: el
 responses:
   intents:
     HassTurnOff:
-      default: "Έκλεισαν τα φώτα στο {{ slots.name }}"
+      default: "Έκλεισε"
       lights_area: "Έκλεισαν τα φώτα"
       fans_area: "Έκλεισαν οι ανεμιστήρες"
       cover: "{{ slots.name }} έκλεισε"
--- a/responses/el/HassTurnOn.yaml
+++ b/responses/el/HassTurnOn.yaml
@@ -2,7 +2,7 @@ language: el
 responses:
   intents:
     HassTurnOn:
-      default: "Άνοιξαν τα φώτα στο {{ slots.name }}"
+      default: "Άνοιξε"
       lights_area: "Άνοιξαν τα φώτα"
       fans_area: "Άνοιξαν οι ανεμιστήρες"
       cover: "{{ slots.name }} άνοιξε"


dsimop:
> The current default response is "Άνοιξαν τα φώτα στο {{ slots.name }}" ("Lights were turned on in {{ slots.name }}") resulting in nonsensical responses like "Άνοιξα τα φώτα στο καφετιέρα" ("Lights were turned on in coffee machine").
> 
> Change this to "Άνοιξε", to mimic what most other languages do.

This is unfortunately a common/known problem occurring with other languages as well. Your generalization approach makes sense. Just a thought from my side to consider. Maybe the word "Άνοιξα..."/"Έκλεισα..." could be a better fit as this is what HA is supposed to give as its answer.

paravoid:
So first of all, it seems in some languages `{{ slot.name }}` is included (e.g. English, German), while in others it's not (e.g. French, Dutch). I went for the second option here, but it seems there is no standard way.

Going back to the drawing board, allow me to use the same example: the response to «Άνοιξε την καφετιέρα». What are our options here?
- Άνοιξε. (_What was implemented here_)
- Καφετιέρα άνοιξε.
- Άνοιξα.
- Άνοιξα καφετιέρα.
- Άνοιξα την καφετιέρα. (_Sounds great, but I'm not sure if it is possible to reuse the article from the prompt?_)

dsimop:
> So first of all, it seems in some languages `{{ slot.name }}` is included (e.g. English, German), while in others it's not (e.g. French, Dutch). I went for the second option here, but it seems there is no standard way.
> 
> Going back to the drawing board, allow me to use the same example: the response to «Άνοιξε την καφετιέρα». What are our options here?
> 
> * Άνοιξε. (_What was implemented here_)
> * Καφετιέρα άνοιξε.
> * Άνοιξα.
> * Άνοιξα καφετιέρα.
> * Άνοιξα την καφετιέρα. (_Sounds great, but I'm not sure if it is possible to reuse the article from the prompt?_)

Your point about the lack of a standardized approach is valid. The success of implementation depends on the chosen language and optimizing it for coherent and proper output. Regarding your pull request (PR), I plan to merge it for the general/default output. The 'article' issue is unresolved, a common problem in other languages too. From your list, I prefer the last two options, though I have concerns about the success of the final option's implementation as well (A targeted ML approach would likely be more effective for the future in addressing the linguistic nuances of the affected languages.).

PR #1690, opened 2023-11-29
[en] Fix weather sentences without a name

synesthesiam:
Sentences like "what is the weather" were not working because I forgot to split them out. Additionally, the test intent "handling" logic needed to be updated to match weather entities.


tetele:
@synesthesiam if you have multiple `weather` entities, will a random one of them simply be queried if you omit the name? How do we pick what response the tests should produce, in this case? Should it be the first entity in `_fixtures`?

PR #1706, opened 2023-12-04
[RO] Fix temperature values for TTS

tetele:
In Romanian, floating point numbers are formatted using a comma as a decimal separator

The TTS engines I've tested don't properly pronounce numeric values if a dot is used as a decimal separator. This PR aims to fix pronunciation issues with numerals in Romanian

Before: "-0.4 grade" -> "zero. patru grade"
After: "-0,4 grade" -> "minus zero virgula patru grade"


protusi:
hi!

solution for me:

{% set state = state_attr('weather.yandex_weather','feels_like') %}
{% if is_number(state) and state | float < 0 %}
  минус
{% endif %}

PR #1730, opened 2023-12-06
[de-CH] slight adjustment of GetTemperature

dontinelli:
Slight adjustment of GetTemperature-sentences


dontinelli:
Added support for ListAddItem as well

PR #1743, opened 2023-12-08
[DE] added 'et' option to 'schalten' expansion rule + test

ChrM3r:
I've added an 'et' option to the expansion_rule 'schalten' because whisper often understands 'schaltet' instead of 'schalte'.

ChrM3r


mib1185 (approved):
Hi @ChrM3r 
thanks again for your contribution. This sounds reasonable to me 👍

PR #1744, opened 2023-12-08
[DE] added 'nichts' as an option for HassNevermind

ChrM3r:
I've added 'nichts' as an option to the HassNevermind intent, as it is a valid "nevermind-sentence" in other voice assistants like Siri

ChrM3r


mib1185 (approved):
Hi @ChrM3r 
thanks for your contribution.
TBH I'm not sure when this could be a valid reply, but it might just be my personal feeling 🙈

ChrM3r:
Haha - that's how different users are. I was actually surprised the first time I tried "nichts" and wasn't understood by Assist.

PR #1747, opened 2023-12-08
[GL] Add weather sentences

svillar:
Adds weather related sentences for Galician


svillar:
This must be updated in case #1748 lands first

cibernox (approved):

PR #1748, opened 2023-12-09
[GL] Replace <name> expansion rule by <nome>

svillar:
Seems to be a common practice to use the local language for the rules' names instead of english.

### DIFF
--- a/sentences/gl/_common.yaml
+++ b/sentences/gl/_common.yaml
@@ -47,7 +47,7 @@ lists:
       - in: "f"
         out: "fahrenheit"
 expansion_rules:
-  name: "[o | a] {name}"
+  nome: "[o | a | os | as] {name}"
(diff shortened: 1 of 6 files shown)


sentences/gl/_common.yaml
```diff
@@ -47,7 +47,7 @@ lists:
       - in: "f"
         out: "fahrenheit"
 expansion_rules:
-  name: "[o | a] {name}"
+  nome: "[o | a | os | as] {name}"
```
cibernox, line 50: My understanding is that keys must remain in english, only values are internationalized.

cibernox: on a second thought, that's not necessary for expansion rules, only for other keys. I'll approve it.

svillar: Yep I checked other languages and that's indeed the case.

PR #1749, opened 2023-12-09
[GL] Shopping list sentences

svillar:


sentences/gl/shopping_list_HassShoppingListAddItem.yaml
```diff
@@ -0,0 +1,11 @@
+language: "pt"
```
cibernox, line 1: pt -> gl

svillar: :smile:

PR #1753, opened 2023-12-09
[DE] add degrees sign to temperature expansion rule

hannemann:
Sometimes whisper return 20°C and sometimes 20 Grad


sentences/de/_common.yaml
```diff
@@ -183,7 +183,7 @@ expansion_rules:
   luefter: "<artikel> (Ventilator[en]|Lüfter)"
   alle_luefter: "<alle> (Ventilatoren|Lüfter)"
   brightness: "{brightness} [Prozent|%]"
-  temperature: "{temperature} [Grad] [{temperature_unit}]"
+  temperature: "{temperature} [Grad] [[°]{temperature_unit}]"
```
mib1185, line 186:
```suggestion
  temperature: "{temperature}[ ][°|Grad][ ][{temperature_unit}]"
```

i think we should adopt the English approach
https://github.com/home-assistant/intents/blob/bfe1eb8feb4a6ed623c4fcdd670fc200f52f23a2/sentences/en/_common.yaml#L304

mib1185 (changes requested):
Hi @hannemann 
thanks for your contribution, just one small comment 👍

sentences/de/_common.yaml
hannemann: I agree :+1:

tests/de/climate_HassClimateSetTemperature.yaml
```diff
@@ -2,8 +2,11 @@ language: de
 tests:
   - sentences:
       - Stelle im Wohnzimmer die Temperatur auf 30 Grad Celsius
+      - Stelle im Wohnzimmer die Temperatur auf 30 °C
```
mib1185, line 5:
```suggestion
      - Stelle im Wohnzimmer die Temperatur auf 30 °C
      - Stelle im Wohnzimmer die Temperatur auf 30°C
```

tests/de/climate_HassClimateSetTemperature.yaml
```diff
@@ -2,8 +2,12 @@ language: de
...
+      - Stelle im Wohnzimmer die Temperatur auf 30°C
```
mib1185, line 6:
```suggestion
```

mib1185 (approved):

hannemann:
My findings are that no matter what i try if the temperature and unit are something like `30°C` without a space in between the slot temperature contains `30°c` instead of just `30`. IMO that's why the tests fail and should not be the case.

I also found that in the hassil library recognize.py line 1050 the `text` property of the `MatchEntity` is returned unchanged but the `value` was cast to an integer .

In the response template the `text` property is used.
Do we have access to that `value` property in the response template?

PR #1771, opened 2023-12-12
[NL] Fix intents with untranslated states

TheFes:
As requested by @tetele in https://github.com/home-assistant/intents/discussions/1768#discussioncomment-7830669

based on https://github.com/home-assistant/intents/pull/1767

waiting for: https://github.com/home-assistant/core/pull/105524

### DIFF
--- a/responses/nl/HassGetState.yaml
+++ b/responses/nl/HassGetState.yaml
@@ -14,13 +14,16 @@ responses:
           Nee, die is {{ state.state_with_unit }}
         {% endif %}
 
-      one_zone: |
+      someone_yesno: |
         {% if query.matched %}
           Ja
-        {% elif state.state == 'not_home' %}
-          Nee, die is niet thuis
         {% else %}
-          Nee, die is in de zone {{ state.state_with_unit }}
+          Nee,
+          {% if state.attributes.get("untranslated_state") in ["not_home", "home"] %}
+            {{ state.state }}
+          {% else %}
+            in zone {{ state.state }}
+          {% endif %}
         {% endif %}
 
       bs_yesno: |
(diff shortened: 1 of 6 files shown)


responses/nl/HassGetState.yaml
```diff
@@ -14,13 +14,16 @@ responses:
...
-      one_zone: |
+      someone_yesno: |
         {% if query.matched %}
           Ja
-        {% elif state.state == 'not_home' %}
-          Nee, die is niet thuis
         {% else %}
-          Nee, die is in de zone {{ state.state_with_unit }}
+          Nee,
+          {% if state.attributes.get("untranslated_state") in ["not_home", "home"] %}
```
mib1185, line 22: i think this could result in confusing replies - let's expect Andrew is `not_home` 
question: Is Andrew at work?
reply: No, he is not home

or do i oversee something?

tetele: That depends on how "Away" is translated into NL in core. In English, that would be "No, away". Which means "away from any zone".

TheFes: I based it on the English version, which does the same.
The only thing I do different in NL is, is that I treat `home` and `not_home` the same, as we don't say `he is at home` but `hij is thuis`

TheFes: > That depends on how "Away" is translated into NL in core. In English, that would be "No, away". Which means "away from any zone".

In Dutch it is similar, `afwezig` is used, which has the same meaning as `away`

mib1185: ok, then I got confused by `not_home` does not mean "is not at home" but "don't know where, but not at home" 🙈

TheFes: It's the state a person entity or device tracker has even it's not in the `home` zone or any other zone

PR #1774, opened 2023-12-12
[SL] ] Fix intents with untranslated states

andrejs2:
Slovenian weather_HassGetWeather and person_HassGetState fix with untranslated states

As requested by @tetele in [#1768 ](https://github.com/home-assistant/intents/discussions/1768#discussioncomment-7830669)

based on https://github.com/home-assistant/intents/pull/1767

### DIFF
--- a/responses/sl/HassGetState.yaml
+++ b/responses/sl/HassGetState.yaml
@@ -14,6 +14,18 @@ responses:
           Ne, {{ state.state_with_unit }}
         {% endif %}
 
+      someone_yesno: |
+        {% if query.matched %}
+          Da
+        {% else %}
+          Ne,
+          {% if state.attributes.get("untranslated_state") == "not_home" %}
+            {{ state.state }}
+          {% else %}
+             {{ state.state }}
+          {% endif %}
+        {% endif %}
+
       any: |
         {% if query.matched %}
           {% set match = query.matched | map(attribute="name") | sort | list %}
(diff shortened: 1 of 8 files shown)


responses/sl/HassGetState.yaml
```diff
@@ -14,6 +14,18 @@ responses:
           Ne, {{ state.state_with_unit }}
         {% endif %}
 
+      someone_yesno: |
+        {% if query.matched %}
+          Da
+        {% else %}
+          Ne,
+          {% if state.attributes.get("untranslated_state") == "not_home" %}
```
tetele, line 22: There's no need for this `if` if there's no difference. In English there's `John is away` and `John is at work`

responses/sl/HassGetState.yaml
```diff
@@ -49,6 +61,32 @@ responses:
...
+            Ne,
+            {% for name in no_match -%}
+              {% if not loop.first and not loop.last %}, {% elif loop.last and not loop.first %} in {% endif -%}
+              {{ name }}
+            {%- endfor %} {% if no_match | length > 1 %}in{% else %}is{% endif %} ni
+            {% if state.attributes.get("untranslated_state") == "not_home" %}
+              {{ state.state }}
+            {% else %}
+               {{ state.state }}
+            {% endif %}
```
tetele, line 86: Same here, there's no need for the if

responses/sl/HassGetState.yaml
andrejs2: I am aware that not all responses are correct. I'll have to invest additional time to extensively work on improving the extensive grammar of Slovenian.

Since there are almost no users for these intents except for me, I will have ample time to test it thoroughly.
If you think it's inadequate, that's fine with me too. I just wanted to hurry because intents and templates in yaml will take up too much of my time during the week.

responses/sl/HassGetState.yaml
andrejs2: Hm, yes, there was "at" which doesn't sound right if I just put "v" (as stands in Slovenian for at).
`John is away`- > `John je proč`
`John is at work` -> `John je v službi`

tetele: https://github.com/home-assistant/intents/pull/1769/files#diff-05a49728660604fda371e824b9dfcd1487b062f2e9f804130b8460a9a3ddaa14R23

Same in Romanian, here's the fix

responses/sl/HassGetState.yaml
tetele: Whatever you think is OK. My point was that the `if` is redundant.

tetele:
@andrejs2 can you please merge the PR after addressing any changes you feel are necessary?

PR #1775, opened 2023-12-12
[EL] Add HassGetWeather intents

dsimop:


tetele:
@dsimop unfortunately, this won't work. See #1768

dsimop:
It seems the problem you mentioned is solved (https://github.com/home-assistant/intents/discussions/1768#discussioncomment-7835433). Thank you for your comment, though.

tetele:
It's not solved in Greek sentences, you need to use that new functionality here https://github.com/home-assistant/intents/blob/954d3821ee69e359199f8ffc0bc8185cdf36ec99/responses/el/HassGetWeather.yaml#L24

dsimop:
To be updated with #1778.

PR #1776, opened 2023-12-12
[NL] Turn lights on/off in the area of a satellite

TheFes:
SSIA


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #1778, opened 2023-12-12
[EL] Fix intents with untranslated state

dsimop:


tetele:
As per #1768 this was supposed to be merged into the branch `untranslated-states`, not `main`. If all goes well, that won't matter, but if there are delays, it may break the usage of `HassGetWeather`.

PR #1779, opened 2023-12-13
[PL] Fix intents with untranslated state (#1767) AND MORE

Uriziel01:
**WARNING**
This pull request contains more than just ported and translated changes from #1767, as some sentences where missing from PL version I've decided to also implement them in the same PR to avoid pushing half broken version to the repository.

I would really appreciate if any other Polish speaker would sanity check all the changes, this is more than I've initially intended but this is what had to be done in order to implement changes.

@witold-gren , @dziadeck47 , @Misiu, @fejrr, @panpawelgit

### DIFF
--- a/responses/pl/HassGetState.yaml
+++ b/responses/pl/HassGetState.yaml
@@ -21,11 +21,64 @@ responses:
         Nie, {{ slots.name }} jest {{ state.state_with_unit }}
         {% endif %}
 
+      someone_yesno: |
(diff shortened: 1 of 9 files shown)


Uriziel01:
#TODO
Yes I'm aware the naive sentence in HassGetState.yaml has to be implemented to accommodate for corner cases with changes from the comment, but at this point the PR is already to large and this can be easily improved in the next iteration

PR #1783, opened 2023-12-14
[PT-BR] Add support for HassListAddItem

luyzfernando08:
[PT-BR] Add HassGetWeather
[PT-BR] Add HassNevermind


v1k70rk4:
@luyzfernando08 Hi, at first glance, there's no newline character at the end of the `tests/pt-br/_fixtures.yaml` file.

If you're developing on your machine, it's a good idea to create a test environment with the `script/setup` command, which is available in the repository.
Then, you can use the `script/lint` command to fix the files, ensuring there's no unnecessary whitespace and that the file's format is correct.
Finally, you can test with the `script/test` command.
If all these are done, it should pass here as well.

luyzfernando08:
> @luyzfernando08 Hi, at first glance, there's no newline character at the end of the `tests/pt-br/_fixtures.yaml` file.
> 
> If you're developing on your machine, it's a good idea to create a test environment with the `script/setup` command, which is available in the repository. Then, you can use the `script/lint` command to fix the files, ensuring there's no unnecessary whitespace and that the file's format is correct. Finally, you can test with the `script/test` command. If all these are done, it should pass here as well.

Thank you for your help.

PR #1784, opened 2023-12-14
[EN] Improve light_HassLightSet sentences

joshuayuen99:


sentences/en/light_HassLightSet.yaml
```diff
@@ -22,6 +22,7 @@ intents:
           - "[<numeric_value_set>] <area> brightness [to] <brightness>"
           - "[<numeric_value_set>] <area> [to] <brightness> brightness"
           - "[<numeric_value_set>] <area> [to] <brightness>"
+          - "[<numeric_value_set>] [all] [the] lights [in | of] <area> to <brightness> [brightness]"
```
tetele, line 25:
```suggestion
          - "[<numeric_value_set>] [all] [the] lights [<in> | of] <area> to <brightness> [brightness]"
```

Also, could you provide an example where `of` makes sense? I can't think of any atm.

tetele (changes requested):
Nice catch

sentences/en/light_HassLightSet.yaml
joshuayuen99: I mostly included that based on the existing sentences. An example could be: "set the lights of the bedroom to 50%".

tetele: I haven't really heard that phrasing before, but it does make sense. If so, I think it would be better to add `of` directly to the `<in>` expansion rule, don't you think?

tetele:
@joshuayuen99 how do you feel [about this](https://github.com/home-assistant/intents/pull/1784#discussion_r1426253046)?

sentences/en/light_HassLightSet.yaml
joshuayuen99: This could make sense. I'm not too familiar with the structure of the rest of the intents, so I'm not sure if there could be any potential negative impact elsewhere. What do you think?

tetele: It's fine by me. Move the `of` to the `<in>` general expansion rule and remove it from here.

joshuayuen99: Done.

tetele:
Can you please update the branch? If all tests go well after that, we'll merge it

joshuayuen99:
> Can you please update the branch? If all tests go well after that, we'll merge it

Updated!

tetele (approved):

tetele:
Thank you @joshuayuen99

PR #1785, opened 2023-12-14
[EN] Context area aware intents

tetele:
Added the area-aware ability to all action intents, not just `light_HassTurnXX` ones.

Also optimized the `light` sentences, as completely optional components in a permutation may result in duplicate matches.

### DIFF
--- a/sentences/en/cover_HassTurnOff.yaml
+++ b/sentences/en/cover_HassTurnOff.yaml
@@ -21,3 +21,14 @@ intents:
         slots:
           domain: cover
         response: cover_device_class
+
+      - sentences:
+          - "<close> [the] {cover_classes:device_class}[<in_here>]"
+        expansion_rules:
+          in_here: "[ in] here"
+        slots:
+          domain: cover
+        response: cover_device_class
+        requires_context:
+          area:
+            slot: true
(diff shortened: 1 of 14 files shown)


TheFes (commented):
wouldn't it be easier to define `<in_here>` in _common.yaml as you use it quite some places now.

tetele:
I've thought about it, but decided that it's used in relatively few places at the moment, given the total number of sentences.

If you ask me, there are **way** too many `_common` lists and expansion rules, particularly since we have local expansion rules and since I've discussed with @synesthesiam about local lists, which seem like a good idea as well.

synesthesiam (approved):

synesthesiam:
This looks good to me. I think we will need to have guidelines (possibly language-specific) about the use of words like "all" and if they affect the current area or the entire house.

I personally think that anything affecting the entire house should be harder to say, like "turn off all the lights in the house", but I'm not familiar enough with other commercial voice assistants to know what's common.

tetele:
> I personally think that anything affecting the entire house should be harder to say, like "turn off all the lights in the house"

Well, saying "turn on/off all the lights" won't turn on all of them, just the ones in the area, so we're safe for now. We can tweak the sentences and leave out "all" later if needed.

PR #1787, opened 2023-12-14
[CA] Update sentences for State and Weather

duhow:
Relates to #1768

### DIFF
--- a/sentences/ca/_common.yaml
+++ b/sentences/ca/_common.yaml
@@ -104,10 +104,10 @@ lists:
         out: unlocked
   presence_states:
     values:
-      - in: home
-        out: a casa
-      - in: away
-        out: fora de casa
+      - in: a casa
+        out: home
+      - in: fora de casa
+        out: not_home
   shopping_list_item:
     wildcard: true
 expansion_rules:
--- a/sentences/ca/_common.yaml
+++ b/sentences/ca/_common.yaml
@@ -102,6 +102,12 @@ lists:
         out: locked
       - in: desbloquejat
         out: unlocked
+  presence_states:
+    values:
+      - in: home
+        out: a casa
+      - in: away
+        out: fora de casa
   shopping_list_item:
     wildcard: true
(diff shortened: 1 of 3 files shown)


sentences/ca/_common.yaml
```diff
@@ -102,6 +102,12 @@ lists:
         out: locked
       - in: desbloquejat
         out: unlocked
+  presence_states:
+    values:
+      - in: home
+        out: a casa
+      - in: away
+        out: fora de casa
```
tetele, line 110: These should be the other way around, but keep in mind that `not_home` means "away", not "not home", i.e. out of any known zone

```suggestion
      - in: a casa
        out: home
      - in: fora de casa
        out: not_home
```

tetele (changes requested):

sentences/ca/_common.yaml
jmceara: Actually, I'm not sure why this is in 'ca' language, but it's from pt-br language and it's my fault, I've inverted words there. Should I place a new PR or you just changed there?
Your sugestion are almost right in portuguese:
```
      - in: home
        out: casa
      - in: away
        out: fora_de_casa
```

If I'm right, `in` means the original word and `out` is translated word, right?
casa => home
fora_de_casa => not in home.

**_away_** means **_far away_** in portuguese.

tetele: I don't understand how the Catalan language leader could have committed something in the `/ca/` folder(s) as a result of this being for `pt-BR`. A quick Google Translate states that "a casa" and "fora de casa" are correct in Catalan as well.

@duhow ?

duhow: In a generic way, I can say "out" `fora` to mean "outside but without knowing where". with no context, assuming it is specific to home, it is "out of home".
"away" can be `lluny` as "far" but again missing the context, so i'd keep this iteration first to see impressions.

PR #1828, opened 2023-12-25
Fix Hebrew response for HassClimateGetTemperature

phoenixperson-c137:
The Hebrew response for `HassClimateGetTemperature` returned the climate's state, rather than the current temp.

Also, when a sentence in Hebrew starts with a number, or an english letter, it jumbles it.

Fixed the value and added the name of the area to the beginning of the sentence to avoid it getting jumbled.

The new sentence translates to  "In {area} {temperature} degrees" (no need for ["it is"] between area and temperature in Hebrew)


leranp:
Thanks @phoenixperson-c137

PR #1832, opened 2023-12-27
[LT] Update HassTurnOff.yaml

KarolisKirna:
Not correct translation of lights_area.


frenck (approved):
Thanks, @KarolisKirna 👍 

../Frenck

PR #1836, opened 2023-12-28
Add Harry Potter 'Lumos and Nox ' spells for controlling lights   

omaramin-2000:
Use "Lumos" work for turning on the lights, and "Nox" for turning off the lights.


tetele:
But why

omaramin-2000:
> But why

Well, I found it's a nice idea to add them in the built-in sentences, as they're enabled in other platforms like Siri, Google Assistant, and alexa. So what do you think about that? 🙂

tetele:
> as they're enabled in other platforms like Siri, Google Assistant, and alexa

No, they're not, at least no more than [custom sentences](https://www.home-assistant.io/voice_control/custom_sentences/), for which this is perfect material.

Sorry, no go.

PR #1838, opened 2023-12-31
[NB] Add Norwegian translations for GetWeather

andebor:


sentences/nb/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,10 @@
+language: nb
+intents:
+  HassGetWeather:
+    data:
+      - sentences:
+          - "Hvordan [s]er været[ ut][ i dag]"
+      - sentences:
+          - "Hvordan [s]er været[ ut] i <navn> [ i dag]"
```
LaStrada, line 8: This does not handle `på`, only `i`. I belive this will fix the issue:
```suggestion
          - "Hvordan [s]er været[ ut] <i_på> <navn> [ i dag]"
```

andebor: Yes, you're absolutely right! Forgot about that option. I'll add it in.

LaStrada (approved):
Great job, thanks! 🚀

PR #1844, opened 2024-01-02
Fix context awareness

jlpouffier:
The way context awareness is processed changed recently (See commit for EN sentences here https://github.com/home-assistant/intents/commit/eda5cba86371a7145124b1aa4d5c19e2351f9d6f) 

I mentioned that FR and NL were still using the old syntax to @synesthesiam and we agreed that we will try to fix this before tomorrow's release.

We now need to explicitly tell in the sentence to add the area in a slot as such:
```yaml
  requires_context:
    area:
      slot: true
```
_This is what changed in the sentences._

The tests have been updated to accommodate the additional slot passed to the intent.

### DIFF
--- a/sentences/fr/light_HassTurnOff.yaml
+++ b/sentences/fr/light_HassTurnOff.yaml
@@ -27,7 +27,8 @@ intents:
         slots:
           domain: light
         requires_context:
-          area: null
+          area:
+            slot: true
 
       # name + area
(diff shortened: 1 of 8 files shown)


TheFes:
Sorry, I missed that this was changed. Changes look good

TheFes (commented):
NL changes look good to me (FR as well BTW)

PR #1846, opened 2024-01-02
[NL] Update ClimateGetTemperature

TheFes:
Make changes for NL intents to get the current temperature from a climate entity


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #1857, opened 2024-01-04
Remove area "all" slot for all languages

piitaya:
Following https://github.com/home-assistant/intents/pull/1500
`area: all` doesn't target all areas but only the area called "all" so it should not be used.


TheFes (approved):
Fine for the NL intents

LaStrada (approved):
NB is good as well

andrejs2 (approved):
Approved for `sl.` Thanks @piitaya.

synesthesiam:
Thanks!

PR #1862, opened 2024-01-05
[FR] Simplify light responses for turn on and off

piitaya:
Simplify response by using masculine singular form because we don't know what's the type of light that is turned on or off (e.g. "Allume les spots du salon").
I also capitalized the common turn off response.

### DIFF
--- a/responses/fr/HassTurnOff.yaml
+++ b/responses/fr/HassTurnOff.yaml
@@ -2,8 +2,8 @@ language: fr
 responses:
   intents:
     HassTurnOff:
-      default: "éteint"
-      lights: "Lumières éteintes"
+      default: "Éteint"
+      light: "Éteint"
       fans: "Ventilateurs éteints"
       cover: "Fermé"
       lock: "Ouvert"
--- a/responses/fr/HassTurnOn.yaml
+++ b/responses/fr/HassTurnOn.yaml
@@ -3,7 +3,7 @@ responses:
   intents:
     HassTurnOn:
       default: "Allumé"
-      lights: "Lumières allumées"
+      light: "Allumé"
       fans: "Ventilateurs allumés"
       cover: "Ouvert"
       lock: "Fermé"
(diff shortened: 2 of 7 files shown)


responses/fr/HassTurnOff.yaml
```diff
@@ -2,8 +2,8 @@ language: fr
 responses:
   intents:
     HassTurnOff:
-      default: "éteint"
-      lights: "Lumières éteintes"
+      default: "Éteint"
+      light: "Éteint"
```
jlpouffier, line 6: So I see two different paths we can take with this PR.

Either we consider that having a specific answer for light is needed, and in this case, I would at least put it to the feminine (singular form) so that it flows well voice-wise
- `Allume les lumières du salon` > `Allumée` (No difference of pronunciation with `Allumées`)
- `Éteindre les lumières du salon` > `Éteinte` (No difference of pronunciation with `Éteintes` .. but better than your current `Éteint`)


Either we do not consider that having a specific answer for light is needed, and in this case, I would go one step further and strip the light response altogether and use the `default` one.

Your current PR defines two `light` answers for `HassTurnOff` and `HassTurnOn`, but one might consider them almost useless as they are identical to the `default` ones

jlpouffier (changes requested):

responses/fr/HassTurnOff.yaml
piitaya: I was thinking of "Eteinte" but "Allume les spots" => "Eteinte" feels weird to me.

jlpouffier: Touché for "Spots" - In this case, I would lead toward option 2, the `light` answers are proving useless and we should just remove them and use the `default` ones

jlpouffier: @piitaya : Instead of having answers per domain `default`, `light`, `cover`, `lock` that are proving pretty useless for almost all of them.

What do you think of having answer per action
`ouvert`
`allumé`
and maybe in the future others. (lock could be `verrouillé` for example)

jlpouffier (changes requested):

jlpouffier (approved):

PR #1864, opened 2024-01-06
[FR] Add context light to sentence targeting `{name}`

jlpouffier:
@piitaya very strange PR, that requires a few explanations

So we have a bug on our French intents today on the intent `light_HassLightSet`.

As you can see here, the intent can be targeted by any entity, of any domain:
Screenshot 2024-01-06 at 16 37 35

So my initial plan was to add 
```yaml
        requires_context:
          domain: light
``` 
on any sentences targeting a `{name}`

and add
```yaml
        slots:
          domain: light
```
on any sentences targeting an `{area}`

The thing is, the test engine fails if I add a `slot` on the `sentence` file with this:
```
FAILED tests/test_language_intents.py::test_light_HassLightSet[fr] - AssertionError: File light_HassLightSet.yaml should only have sentences without a domain slot
```

My gues is that, if you target an area, the `domain` `light` is already targeted on the intent itself and you do not need to add it.

I checked in English, it's the same behavior... a `requires_context` in any sentences targeting a `{name}`, nothing in any sentences targeting an `{area}`

The second peculiar thing: Nothing has to be added on the test, I tried adding a `domain: light` as `context`, but the test engine fails too with
```
FAILED tests/test_language_sentences.py::test_light_HassLightSet[fr] - AssertionError: File light_HassLightSet: tests should not have a domain slot
```

So the context is not added for `light_HassLightSet` .. in comparison of `homeassistant_HassTurnOn` for example where we had to add the `domain` as `context` for all tests. 
Very strange inconsistency, I guess it is coming from the fact that the intent already targets a domain by design.

So the only thing needed is to require a context on the sentence.
I tested it and indeed, this addition makes the intent parser fails for `Allume less volets en rouge`, which is a good thing.

### DIFF
--- a/sentences/fr/light_HassLightSet.yaml
+++ b/sentences/fr/light_HassLightSet.yaml
@@ -20,6 +20,8 @@ intents:
           - "[<le>]{name} [à] {brightness}<pourcent> [de] luminosité"
           # Sapin luminosité 30%
           - "[<le>]{name} luminosité [à] {brightness}<pourcent>"
+        requires_context:
+          domain: light
         response: brightness
 
       # brightness (area)
@@ -60,6 +62,8 @@ intents:
           - "[<le>]{name} [<dans>] [<le>]{area} {brightness}<pourcent> [de] luminosité"
           # Mirroir salle de bain luminosité 40%
           - "[<le>]{name} [<dans>] [<le>]{area} luminosité {brightness}<pourcent>"
+        requires_context:
+          domain: light
         response: brightness
 
       # color (name)
@@ -68,6 +72,8 @@ intents:
           - "<regle> la couleur [<de>] [<le>]{name} [en] {color}"
           # Règle le mirrior en vert
           - "(<regle>|<allume>) [<le>]{name} [avec la couleur | de couleur | en] {color}"
+        requires_context:
+          domain: light
         response: color
 
       # color (area)
@@ -88,4 +94,6 @@ intents:
           - "<regle> la couleur [<de>] [<le>]{name} [<dans>] [<le>]{area} [en] {color}"
           # Allume le sapin de l'exterieur en vert
           - (<regle>|<allume>) [<le>]{name} [<dans>] [<le>]{area} [avec la couleur | de couleur | en] {color}
+        requires_context:
+          domain: light
         response: color


piitaya (approved):
I wonder if we can add test to test that it doesn't work with another domain than light. 
"Règle la luminosité du volet à 50%" should fail.

jlpouffier:
I think so. A test is just matching an intent. I guess we can make it match to nothing. I can try that yeah. Good idea

jlpouffier:
So, I did not find a way. @synesthesiam : Do you know if we can create a test to check if something is **NOT** matching?
We want to create a test for "set the door color to red" (That works in French, that PR fixes that)

tetele:
> Very strange inconsistency, I guess it is coming from the fact that the intent already targets a domain by design.

Let me try to explain this.

Intents **do not** target a domain by design necessarily, but given that the definition of targeting sentences was in `light_HassLightSet`, the `domain` slot is already populated with the first part of the file name (`light`, in this case). The `domain` is implicitly passed as a `slot`. This does not apply to the `homeassistant` domain, where entities can be anything, really.

The difference between `context` and a `slot` is that `context` only gets populated in the conversation agent side of things, with the exception of when a piece of context explicitly gets promoted to a `slot` [like here](https://github.com/home-assistant/intents/blob/71e996fe522e733b42e5a8f2a4e85e659b8244b8/sentences/en/light_HassTurnOn.yaml#L36).

`requires_context` and `excludes_context` are evaluated against `context`, not `slots`.

Now circling back...

> I checked in English, it's the same behavior... a `requires_context` in any sentences targeting a `{name}`, nothing in any sentences targeting an `{area}`

The reason for this is that [`name` is evaluated](https://github.com/home-assistant/core/blob/840089b8ac86ea22c1ebc1fd58f5b938c2951051/homeassistant/components/conversation/default_agent.py#L327) before `area` or anything else. As such, the first state with the entity name `name` will be recognized, which is a problem if you have entities pertaining to multiple domains with the same name. We use `requires_context` to fix that.

For other search criteria, where a `domain` slot is implicitly passed (since the sentences are defined in a file beginning with `domain_`), the matching takes that `domain` into account and there's no need to infer it from context.

I hope that makes a bit more sense.

jlpouffier:
All clear(er) @tetele !
Thx

PR #1865, opened 2024-01-07
[EN] Better error messages

tetele:
Friendlier error messages in the context of https://github.com/home-assistant/core/pull/107151

I picked "called" instead of "named" to hint that aliases are included as well, as opposed to just names.

I picked "I am not aware" instead of "I couldn't find" as I think that, traditionally, inability to find something hints to its non-existence, whereas the issue here might be a lack of exposure which can be fixed by exposing an entity to Assist.

I used `in the {{area}} area` instead of `in {{area}}` because of prepositionless area names like `Outside` or other area names that might require different prepositions, like `[on the] Lawn`, `[on the] Porch` etc.

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -2,10 +2,10 @@ language: "en"
 responses:
   errors:
     no_intent: "Sorry, I couldn't understand that"
-    no_area: "No area named {{ area }}"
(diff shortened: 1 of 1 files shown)


synesthesiam (approved):

synesthesiam:
Thanks! Not to future self: update guidelines to mirror this PR.

PR #1886, opened 2024-01-20
[HU - untranslated state] Hungarian version prepared for untranslated state

v1k70rk4:
Hungarian version prepared for untranslated state


v1k70rk4:
Could someone help me understand why my `script/lint` runs smoothly on my own machine but always fails here? :)

v1k70rk4:
"I visually searched for the error, but I'm really curious about what I'm doing wrong? I run a `script/setup` every day and the local `lint` didn't catch the problem.

Issue #1602, opened 2023-11-01
[ES] Unable to make some combinations on shopping list


arpia49:
Hello, I'm unable to make the following sentences work:

- Añadir manzanas a la lista de la compra
- Añadir manzanas a mi lista de la compra

![image](https://github.com/home-assistant/intents/assets/126833/7a6a2d7d-dfc8-4a99-92c6-a3329649b7fa)

While the following works fine:

- Añadir manzanas en la lista de la compra
- Añadir manzanas en mi lista de la compra

![image](https://github.com/home-assistant/intents/assets/126833/ebf2c465-0d20-4bfd-82c3-198a4f04485d)

@davefx  I think something may be wrong in the definition, maybe spacing?

[sentences/es/shopping_list_HassShoppingListAddItem.yaml](https://github.com/home-assistant/intents/blob/main/sentences/es/shopping_list_HassShoppingListAddItem.yaml)

luca-angemi:
Having the same issue, I think the problem is because of this

![image](https://github.com/home-assistant/intents/assets/9977955/1c395648-5c42-4877-98ea-2ff867500f3b)

so basically

```
    data:
      - sentences:
          - <añadir> <item> (en|a)<my_list>
        response: item_added
        expansion_rules:
          my_list: "[ mi| la] lista [de] [la] [compra]"
```

HA gets **```manzas a l```** as ```<item>``` then **```a```** as ```(a)``` then **```lista de la compra```** as ```<my_list>```.

To workaround this I extended the intent:

```
language: "es"
intents:
  HassShoppingListAddItem:
    data:
      - sentences:
          - "apunta {shopping_list_item:item}."
        response: item_added
```

With this I just have to say **```apunta manzanas.```**

To be noticed that the final ```.``` avoids having in your list ```manzanas.```

davefx:
Reopening the issue. The problem is not fixed and still happens in the last version, both with to-do lists and with the shopping list.

The problem is caused because of the wildcard rule, that currently matches partial words.
In order to make it work, I've made a change in the to-do list sentence definition so we always require the definite article "la lista" in order to avoid false recognitions (but then ignoring sentences without the definite article).

@synesthesiam Could we make some kind of change so the `{shopping_list_item:item}` wildcard only matches full words (and never partial words), so `Añadir manzanas a la lista de la compra` never matches `manzanas a l` as the item name?

tetele:
Related to https://github.com/home-assistant/hassil/issues/92

PR #1893, opened 2024-01-24
[ES] Update shopping_list_HassShoppingListAddItem.yaml

JaviElio:
Temporary fix for  #1602 to avoid partial words as  <item>


davefx:
Gracias, @JaviElio !

PR #1895, opened 2024-01-25
[FR] Error message improvements

jlpouffier:
## New error message improvements 🎉 

We renamed `area` to `pièce` instead of `zone`.
`zone` is already used for `zone` (An external concept like your kid's school or your work place)

We felt the wording `Aucun appareil ou entité appelé ...` was a bit heavy for a voice interface so we simplified it to `je ne connais pas l'appareil ...`

The biggest change is that we now translate `domains` and `device_classes` for some errors.

Here are a few examples of interactions you will get from this PR:

- `Ouvre les rideaux du jardin`
  - Before:  `Jardin ne continent pas de curtain`
  - Now ✨: `Désolé, je n'ai pas trouvé de rideaux dans jardin`
- Allume les lumières du jardin
  - Before:  `Jardin ne continent pas de light`
  - Now ✨: `Désolé, je n'ai pas trouvé de lumières dans jardin`

## What's "not perfect"
`{area}` cannot be prefaced with the correct article. 

We cannot say:

  - `Dans le jardin`
  -  `Dans la cuisine`
  -  `Dans l'entrée`

 ... because we simply do not know 

For `Cover_HassTurnOn` and `Cover_HassTurnOff`, we have a hybrid mapping of a single word: "Store" matching 3 device classes: 
```yaml
lists:
  ...
  cover_classes:
    values:
      - in: "store[s]"
        out:
          - "shade"
          - "blind"
          - "awning"
      - ....
 ```
 When we say `Ouvre les stores dans {area}` ... If `area` does not contain any of these device classes, then unfortunately, the response will only return the first one alphabetically (awning) .... 
As such 
image

Note: I tried to move the order so that `shade` would be the first because `shade` is translated as "Store" .. but I think HASIL is re-ordering them alphabetically...

Note 2: Our hybrid mapping **does work** in case you are wondering, we are really targeting all 3 device classes as shown here (And also on the tests of this repo)
```yaml
intent:
  name: HassTurnOn
slots:
  device_class:
    - shade
    - blind
    - awning
  area: jardin
  domain: cover
details:
  device_class:
    name: device_class
    value:
      - shade
      - blind
      - awning
    text: stores
  area:
    name: area
    value: jardin
    text: jardin
  domain:
    name: domain
    value: cover
    text: ''
targets: {}
match: true
sentence_template: <ouvre> [<tous>] [<le>]{cover_classes:device_class} [<dans>] [<le>]{area}
unmatched_slots: {}
```
This only issue is how the response is constructed, and this is not something we can control here unfortunately 

**_EDIT_**
After talking to @piitaya , we are simplifying our cover device classes mapping
```yaml
      - in: "store[s] [vénitien[s]]"
        out:
          - "blind"
          - "shade"
```
"Store" targets both type of "Store" and both device classes are now translated as "store" (and not "Store vénitien")

JLo

### DIFF
--- a/sentences/fr/_common.yaml
+++ b/sentences/fr/_common.yaml
@@ -2,10 +2,54 @@ language: fr
 responses:
   errors:
     no_intent: "Désolé, je n'ai pas compris"
-    no_area: "Aucune zone appelée {{ area }}"
-    no_domain: "{{ area }} ne contient pas de {{ domain }}"
-    no_device_class: "{{ area }} ne contient pas de {{ device_class }}"
-    no_entity: "Aucun appareil ou entité appelé {{ entity }}"
+    no_area: "Désolé, je ne connais pas la pièce {{ area }}"
+    no_domain: |
+      {% set translations_domains_with_article = {
+        "button": "de boutons",
+        "camera": "de caméras",
+        "input_button": "de boutons",
+        "alarm_control_panel": "d'alarmes",
+        "automation": "d'automatisations",
+        "fan": "de ventilateurs",
+        "climate": "de thermostats",
+        "humidifier": "d'humidificateurs",
+        "input_boolean": "de commutateurs",
+        "siren": "de sirènes",
+        "water_heater": "de ballon d'eau chaude",
+        "light": "de lumières",
+        "switch": "de commutateurs",
+        "script": "de scripts",
+        "remote": "de télécommandes",
+        "lock": "de verrous",
+        "vacuum": "d'aspirateurs",
+        "scene": "de scènes",
+        "media_player": "de lecteurs multimédia",
+        "lawn_mower": "de tondeuses à gazon",
+        "valve": "de vannes"
+        } %}
+      {% if domain in translations_domains_with_article %}
+        Désolé, je n'ai pas trouvé {{ translations_domains_with_article[domain] }} dans {{ area }}
+      {% else %}
+        Désolé, je n'ai rien trouvé de correspondant  dans {{ area }}
+      {% endif %}
+    no_device_class: |
+      {% set translations_cover_device_classes_with_article = {
+        "awning": "d'auvents",
+        "blind": "de stores vénitiens",
+        "curtain": "de rideaux",
+        "door": "de porte",
+        "garage": "de portes de garage",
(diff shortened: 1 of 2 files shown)


jlpouffier:
@piitaya 
Question for you regarding the limitation on the cover device classes.
We can also simply remove the mapping for "awning", "blind" and "shade" 

It would work for the other but not for our weird "Hybrid mapping"
Example:
- Ouvre les rideaux du jardin
  -  Désolé, je n'ai pas trouvé de rideaux dans jardin
- Ouvre les volets du jardin
  -  Désolé, je n'ai pas trouvé de volets dans jardin
- Ouvre les stores du jardin
  -  Désolé, je n'ai rien trouvé de correspondant  dans jardin

(We fallback to the generic one)

tetele:
Here's a thought from a poor-French-speaking-intruder :D In Romanian, I named all areas "spaces", both in the core/frontend translations and in the intents. Does "espace" make more sense in French than "pièce", which AFAIK literally means room?

jlpouffier:
@tetele I think overall "Espace" would be better as it fits open floor.
But unfortunately, this ship has sailed for the Frenchies a long time ago when we decided to name area "Pièce" everywhere.

I guess we could, one day, make an effort to change all translations in one shot to keep consistency 🙈

sentences/fr/_common.yaml
```diff
@@ -2,10 +2,54 @@ language: fr
...
+        Désolé, je n'ai pas trouvé {{ translations_domains_with_article[domain] }} dans {{ area }}
+      {% else %}
+        Désolé, je n'ai rien trouvé de correspondant  dans {{ area }}
+      {% endif %}
+    no_device_class: |
+      {% set translations_cover_device_classes_with_article = {
+        "awning": "d'auvents",
+        "blind": "de stores",
+        "curtain": "de rideaux",
+        "door": "de porte",
```
piitaya, line 40:
```suggestion
        "door": "de portes",
```

piitaya (changes requested):

sentences/fr/_common.yaml
jlpouffier: oops

piitaya (approved):

PR #1903, opened 2024-01-29
Remove old package directory

synesthesiam:


adrianschroeter:
From where is this supposed to come now?

HA 2024.2.0b0 has currently multiple failing components for me, because get_intents is not available anymore

adrianschroeter:
ah, sorry, just found https://github.com/home-assistant/intents-package ....

PR #1906, opened 2024-02-01
Migrate errors

synesthesiam:
We plan to add some new error messages, and migrate previous errors to be more precise. The following core PR in HA implements some of the new errors: https://github.com/home-assistant/core/pull/109269

The existing `no_domain` and `no_device_class` errors have been renamed to `no_domain_in_area` and `no_device_class_in_area`. Additionally, `no_entity` has been renamed to `no_device`.

A complete list of the proposed errors are here:

* `no_intent` - Intent was not recognized.
* `handle_error` - Unexpected error while handling intent.
* `no_area` - Area does not exist.
* `no_domain` - No devices exist for a domain.
* `no_domain_exposed` - No devices are exposed for a domain.
* `no_domain_in_area` - No devices in an area exist for a domain.
* `no_domain_in_area_exposed` - No devices in an area are exposed for a domain.
* `no_device_class` - No devices of a class exist.
* `no_device_class_exposed` - No devices of a class are exposed.
* `no_device_class_in_area` - No devices of a class exist in an area.
* `no_device_class_in_area_exposed` - No devices of a class are exposed in an area.
* `no_device` - Device does not exist.
* `no_device_exposed` - Device is not exposed.
* `no_device_in_area` - Device does not exist in area.
* `no_device_in_area_exposed` - Device in area is not exposed.
* `duplicate_devices` - More than one device matched with the same name.
* `duplicate_devices_in_area` - More than one device in an area matched with the same name.


hertzg (approved):
:shipit:

davefx (approved):
New sentences already translated into Spanish.

nagyrobi:
@v1k70rk4

dontinelli (approved):
Looks great. I will add de-CH translation for new error messages shortly.

zubir2k (commented):
I have reviewed ms-MY ok. Thanks 👍🏼

script/intentfest/add_language.py
```diff
@@ -99,12 +99,23 @@ def run() -> int:
                 "language": language,
                 "responses": {
                     "errors": {
-                        "no_intent": "TODO: Sorry, I couldn't understand that",
-                        "no_area": "TODO: No area named {{ area }}",
-                        "no_domain": "TODO: {{ area }} does not contain a {{ domain }}",
-                        "no_device_class": "TODO: {{ area }} does not contain a {{ device_class }}",
-                        "no_entity": "TODO: No device or entity named {{ entity }}",
-                        "handle_error": "TODO: An unexpected error occurred while handling the intent",
+                        "no_intent": "TODO Sorry, I couldn't understand that",
```
jlpouffier, line 102: Why `TODO` at the beginning of every sentence here?

sentences/en/_common.yaml
```diff
@@ -1,12 +1,30 @@
...
     no_area: "Sorry, I am not aware of any area called {{ area }}"
-    no_domain: "Sorry, I am not aware of any {{ domain }} in the {{ area }} area"
-    no_device_class: "Sorry, I am not aware of any {{ device_class }} in the {{ area }} area"
-    no_entity: "Sorry, I am not aware of any device or entity called {{ entity }}"
-    handle_error: "An unexpected error occurred while handling the intent"
+    no_domain: "Sorry, I am not aware of any {{ domain }}"
+    no_domain_in_area: "Sorry, I am not aware of any {{ domain }} in the {{ area }} area"
+    no_device_class: "Sorry, I am not aware of any {{ device_class }}"
+    no_device_class_in_area: "Sorry, I am not aware of any {{ device_class }} in the {{ area }} area"
+    no_device: "Sorry, I am not aware of any device called {{ device }}"
```
jlpouffier, line 14: Technically the error should be `no_entity`, as we do not target `device` with sentences.

I know it's confusing because, for most languages, the spoken error will be "I am not aware of this device" because an entity is a very technical term and should not be used while conversing with a voice assistant.

Nevertheless, it should be `no_entity`

I have no strong opinion wether we should change it back or not. I know a core PR is also pending, and If I had to chose, I would prefer to move fast (this time)

dsimop (approved):

andrejs2 (approved):
`sl` is OK

jlpouffier:
@synesthesiam 
Quesiton: 
If I query a device that does not exist AT ALL in an area that does exist, for example `Turn on DLKFJGS in the living room`

Will I have a `no_device` error or a `no_device_in_area` error?

I am asking because we are translating the errors in French with @piitaya and we wanted to translated `no_device_in_area` as `Sorry, {{device}} is not in this area`

v1k70rk4:
Hello,

Something happened that I just rewrote these error messages last night. :) https://github.com/home-assistant/intents/pull/1905 And this is exactly the previous version. Should I wait for this to be merged and rewrite it again? Or can I still rewrite this? But I'm afraid I've rewritten a bunch of responses, the test would even get stuck with the Hungarian.

zubir2k (approved):
Looks ok for ms-MY. Thanks 👌🏻

synesthesiam:
@tetele From the user's perspective, I don't think the device/entity distinction is important. This is only for error responses, and doesn't mean that the intents themselves will not receive entity names or anything.

@v1k70rk4 Go ahead and merge. I will wait as long as I can and clean up on my end.

@jlpouffier It would be `no_device_in_area` in this case. A confusing case is when you have both a device **and** and area that don't exist: "turn on the monkey in the spaceship". In this case, I went with `no_area`.

tetele:
@synesthesiam it's fine for the sentences, those can be perfected later on. My issue was with the keys, i feel as though they create confusion between devices and entities

jlpouffier:
@synesthesiam Great, with @piitaya  we created another PR, we will merge later (https://github.com/home-assistant/intents/pull/1909), after yours

nagyrobi (approved):

jlpouffier (approved):

synesthesiam:
> @synesthesiam it's fine for the sentences, those can be perfected later on. My issue was with the keys, i feel as though they create confusion between devices and entities

Do you see any benefit to having both `no_device` and `no_entity` keys? If it was combined into `no_device_or_entity`, then there would be confusion about which variable to use: `{{ device }}` or `{{ entity }}`.

tetele:
I have referenced that PR in which i have made devices available (actual HA devices, not entities) as a prefilled slot. I still want to do that, although i will have to rewritw the PR.

As such, we will need both `no_entity` *and* `no_device` errors going forward. I think it's a bad idea to use `no_device` as a key for something that is a missing entity, not a missing device. The lingo exposed to the user through the error message is less relevant in this argument.

synesthesiam:
OK, I am changing "device" to "entity" in the relevant keys. We'll add device keys later as needed.

tetele (approved):

tetele:
@synesthesiam I've edited the PR description to reflect the rollback to `entity`

PR #1428, opened 2023-06-15
Add binary sensor for safty

mymyke:

PR #1926, opened 2024-02-03
[DE] Add binary sensor safety

steffenrapp:
#1428 got closed so I'm adding the safety states.


sentences/de/binary_sensor_HassGetState.yaml
```diff
@@ -371,3 +371,14 @@ intents:
           domain: binary_sensor
           device_class: heat
           state: "on"
+
+      # Safety
+      - sentences:
+          - "(ist|sind) <name> {bs_safety_states:state} [<area>]"
```
mib1185, line 377: we need also some tests for area aware sentences

mib1185 (changes requested):
Hi @steffenrapp 
thanks for this contribution, just one comment from my side

sentences/de/binary_sensor_HassGetState.yaml
steffenrapp: Sure, done 😃

sentences/de/binary_sensor_HassGetState.yaml
```diff
@@ -384,3 +384,16 @@ intents:
           domain: binary_sensor
           device_class: problem
           state: "on"
+
+      # Safety
+      - sentences:
+          - "(ist|sind) <name> {bs_safety_states:state} [<area>]"
```
mib1185, line 390: i think sind is not needed here, since we are responding with `einzeln_janein`.
if we want to ask in plural, we need also to respond with `irgendeins` or `alle`

mib1185 (changes requested):

sentences/de/binary_sensor_HassGetState.yaml
steffenrapp: Yes and no. The name itself could be a plural even if it's just one entity or maybe a group of sensors.
In the end it's just a translation of the English version where there's also both. But I see possibilities where 'sind' may be required even if it's unlikely.

mib1185: mhhh ... ok, this somehow convince me. maybe we should have at least one test sentence which uses "sind"

steffenrapp: I've added a test for a sentence with plural.

mib1185 (approved):
LGTM, thanks 👍

PR #1928, opened 2024-02-03
[DE] Add binary sensor problem

steffenrapp:
#1412 got closed so I'm adding the problem binary sensor.


mib1185 (approved):
HI @steffenrapp 
thanks for this, LGTM 👍

PR #1931, opened 2024-02-04
[DE] Add binary sensor light

steffenrapp:


mib1185 (changes requested):
Hi @steffenrapp 
thanks for this ... even sentences like "wie viel Licht ist an?" or "ist irgendwo Licht ausgelöst?" sounds weird to me 🙈
we still need some more test sentences, since we add 15 sentences, but only 12 test sentences ... also they use many expansion rules and optional words.
maybe we should create two new expansion rules:
```yaml
ist_wurde: "(ist|wurde)"
sind_wurden: "(sind|wurden)"
```
than replace `(ist|wurde)` by `<ist_wurde>` and `(ist|sind|wurde[n])` by `(<ist_wurde>|<sind_wurden>)` ... with this the amount of test sentences can be reduced, since we do not need to check all these optionals multiple times

steffenrapp:
Hey @mib1185 thank you for the feedback. I know, not every combination sounds natural. I think if it doesn't do anything wrong it shouldn't be a problem. Many voice assistants understand sentences that are not grammatically correct.
I've implemented those expansion rules (may use them at more places in a separate PR) and added more test sentences with area to cover them.

mib1185 (approved):
thanks @steffenrapp 👍 i think this is ok for now

PR #1936, opened 2024-02-05
Improve light control in french

piitaya:
Add plural form because some STT writes "Jour" as "Jours" because it's only one word. WDYT @jlpouffier?


jlpouffier:
We need to ask what jackouille thinks but I am ok with your proposal 😅

jlpouffier (approved):

PR #1938, opened 2024-02-05
[SV] Translating new error messages from #1906 to Swedish and added missing colors

HLandstrom:
Added the new error messages from #1906 and added missing colors from the english


larsdunemark:
LGTM

PR #1957, opened 2024-02-14
Correct error in the italian expansion rule for the

malorezzanzo64:
In the previous version nelgli was allowed instead of negli


auanasgheps:
@tetele apologiees. I merged some stuff today, and I added [IT] only to some of these. Will doublecheck it in the future!

PR #1962, opened 2024-02-14
Delete expansion rule for temp in Italian

malorezzanzo64:
Expansion rule for temp is useful in english where it can be expressed as temp or temperature, but not in italian where it is called only temperatura. I deleted the expansion rule and substituted with [la] temperatura like other languages already do


auanasgheps:
Grande, grazie!

PR #1980, opened 2024-02-22
Fi lists rewrite

salleq:
added todo and fixed shopping list so they both should work now


salleq:
I'll create a new branch and create a new PR from there, this isn't working...

PR #1983, opened 2024-02-22
[CS] Add singular for area light

bruxy70:
Adding a singular form of "light" to area in CS language (e.g. Turn on the light in the living room)


halecivo (approved):
I was wondering whether we should group singular and plural to not need to write each sentence in two variants, but its okay to go like this for now.

bruxy70:
I thought about that, but the optional "all" word only makes sense for plural. So it would become quite complex....

PR #1985, opened 2024-02-23
[DE] Add a few more variations to shopping list add item

ctron:
While one can argue if it's grammatically correct, it's in use.


mib1185 (changes requested):
Hi @ctron could you please also add some additional test sentences to cover these additions?

ctron:
> Hi @ctron could you please also add some additional test sentences to cover these additions?

Sure, done that. And rebased.

mib1185 (approved):
Thanks @ctron 👍

mib1185 (changes requested):
Hi @ctron 
sorry, did not recognize that there are two different intents `HassShoppingListAddItem` and `HassListAddItem`. You should add these additions to both sentence files (_`shopping_list_HassShoppingListAddItem.yaml` and `todo_HassListAddItem.yaml`_) and also extend the tests for both.

ctron:
> sorry, did not recognize that there are two different intents `HassShoppingListAddItem` and `HassListAddItem`.

Me neither :grinning: … thanks for pointing that out. I think I fixed it.

mib1185 (approved):
Thanks @ctron 👍

PR #1986, opened 2024-02-23
Suggested custom sentences for media players

OzGav:
I'm hoping the syntax is ok as I have modified these from my working custom sentences that were developed for Music Assistant users.

Removed STOP in media player pause as there is a separate `media_stop` service call.

### DIFF
--- a/sentences/en/media_player_HassMediaNext.yaml
+++ b/sentences/en/media_player_HassMediaNext.yaml
@@ -4,6 +4,6 @@ intents:
     data:
       - sentences:
           - "next [track|item] [on|for] <name>"
-          - "Skip [(to [the] next | (the|this)) [(song|track)]] [on] <name>"
+          - "(skip [(to [the ]next [(song|track)]|([the ](song|track)|this [(song|track)]) )];[on ]<name>)"
         requires_context:
           domain: media_player
--- a/sentences/en/media_player_HassMediaNext.yaml
+++ b/sentences/en/media_player_HassMediaNext.yaml
@@ -4,5 +4,6 @@ intents:
     data:
       - sentences:
           - "next [track|item] [on|for] <name>"
+          - "Skip [(to [the] next | (the|this)) [(song|track)]] [on] <name>"
         requires_context:
           domain: media_player
--- a/sentences/en/media_player_HassSetVolume.yaml
+++ b/sentences/en/media_player_HassSetVolume.yaml
@@ -3,9 +3,9 @@ intents:
   HassSetVolume:
     data:
       - sentences:
-          - "(set|change|increase|decrease) <name> volume to <volume> [percent]"
-          - "Turn <name> volume (up|down) to <volume> [percent]"
-          - "((set|change|increase|decrease) the volume to <volume> [percent];[on ]<name>)"
-          - "(Turn the volume (up|down) to <volume> [percent];[on ]<name>)"
+          - "<numeric_value_set> <name> volume to <volume> [percent]"
+          - "turn <name> [volume] (up|down) to <volume> [percent]"
+          - "(<numeric_value_set> the volume to <volume> [percent];[on ]<name>)"
+          - "(turn (the volume;(up|down)) to <volume> [percent];[on ]<name>)"
         requires_context:
           domain: media_player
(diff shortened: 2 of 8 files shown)


OzGav:
Thanks. First time doing this. Looks like there aren't any tests from the initial PR. I will see if I can fabricate some based on existing tests.

v1k70rk4:
You can find the tests here: https://github.com/home-assistant/intents/tree/main/tests/en

OzGav:
https://github.com/home-assistant/intents/pull/1987 tests added

sentences/en/media_player_HassMediaNext.yaml
```diff
@@ -4,5 +4,6 @@ intents:
     data:
       - sentences:
           - "next [track|item] [on|for] <name>"
+          - "Skip [(to [the] next | (the|this)) [(song|track)]] [on] <name>"
```
tetele, line 7:
```suggestion
          - "skip [(to [the ]next|(the|this)) (song|track)][ <in>] <name>"
```
Sentences should be normalized to use only lowercase characters, although that normalization happens behind the scenes as well. But for consistency and readability, use lowercase letters.
Optional components should encompass the spaces before/after them.
Spaces at the beginning/end of alternatives improve readability, but also increase processing time.
Not sure if [`<in>`](https://github.com/home-assistant/intents/blob/e5c75313e96a1ffb7008e945d7c784402ff54362/sentences/en/_common.yaml#L339) is a good fit here instead of `on`

But the biggest issue is that your proposal would match `skip the on <name>`, which makes no sense. If you really want to have `(song|track)` optional, make it so only for the `to [the ]next` alternative.

sentences/en/media_player_HassSetVolume.yaml
```diff
@@ -3,6 +3,9 @@ intents:
   HassSetVolume:
     data:
       - sentences:
-          - "set <name> [volume] to <volume>"
+          - "(set|change|increase|decrease) <name> volume to <volume> [percent]"
```
tetele, line 6:
```suggestion
          - "<numeric_value_set> <name> volume to <volume>"
```

sentences/en/media_player_HassSetVolume.yaml
```diff
@@ -3,6 +3,9 @@ intents:
...
+          - "Turn <name> volume (up|down) to <volume> [percent]"
```
tetele, line 7:
```suggestion
          - "turn <name> volume (up|down) to <volume>"
```

`<volume>` already is `{volume}[percent]` https://github.com/home-assistant/intents/blob/e5c75313e96a1ffb7008e945d7c784402ff54362/sentences/en/_common.yaml#L341

sentences/en/media_player_HassSetVolume.yaml
```diff
@@ -3,6 +3,9 @@ intents:
...
+          - "((set|change|increase|decrease) the volume to <volume> [percent];[on ]<name>)"
```
tetele, line 8: use `<numeric_value_set>` and consider changing `on` to `<in>`

sentences/en/media_player_HassSetVolume.yaml
```diff
@@ -3,6 +3,9 @@ intents:
...
+          - "(Turn the volume (up|down) to <volume> [percent];[on ]<name>)"
```
tetele, line 9:
```suggestion
          - "(turn the volume (up|down) to <volume>;[on ]<name>)"
```

sentences/en/media_player_HassSetVolume.yaml
tetele: Also, shouldn't `volume` be optional here?

`turn the office speaker up to 11%`

sentences/en/media_player_HassSetVolume.yaml
tetele: Consider using `(the volume;(up|down))`, as both
`turn up the volume to 12 on X`
`turn the volume up to 12 on X`
could work

tetele (changes requested):
Thanks a lot for these!

As @v1k70rk4 said, you also need tests, but these are a great start and an excellent first contribution! I've left a bunch of comments, but nothing major.

tetele:
@OzGav i've run the CI pipeline but it failed. Before submitting the PR again, please make sure that the tests pass locally. `pytest tests --language en`

OzGav:
Yeah sorry the tutorial just said to run the local tests. All should be working now.

tetele:
Thank you very much @OzGav!

PR #1987, opened 2024-02-23
Add tests to support PR#1986

OzGav:
https://github.com/home-assistant/intents/pull/1986

### DIFF
--- a/tests/en/media_player_HassMediaNext.yaml
+++ b/tests/en/media_player_HassMediaNext.yaml
@@ -2,6 +2,12 @@ language: en
 tests:
   - sentences:
       - "next item on TV"
+      - "Skip song on the TV"
+      - "Skip track on the TV"
+      - "Skip to the next song on the TV"
+      - "Skip to next track on the TV"
+      - "Skip this song on the TV"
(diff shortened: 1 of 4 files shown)


tetele:
This won't work, as the CI pipeline needs tests to pass **per PR** so you have to commit your tests to #1986

v1k70rk4:
Oops :) Sorry for the many requests, but this is not right. The essence of PRs is that you submit them and after approval, the system runs a test on them. Now, #1986 will not run either, because there is no test in that PR, and this one does not have a sentence to test. It would be wise to close #1987 and add it to #1986. If that doesn't work, close both and put everything together in a third one :)

But it's also possible that @tetele is a guru and will solve it :) But the correct procedure is definitely the above method, where a PR must be valid on its own.

v1k70rk4:
Well, that was faster :D

tetele:
> Now, #1986 will not run either, because there is no test in that PR, and this one does not have a sentence to test.

It would run, but being @OzGav 's first contribution, a maintainer has to approve the CI workflow. Which I will as soon as he commits the missing tests.

PR #1970, opened 2024-02-21
[DE] Add new intents for cover, valve, vacuum, media player

mib1185:
- Reference: #1967

PR #1991, opened 2024-02-24
[DE] Implement additions from #1970

steffenrapp:
Implement feedback from #1970 and create test cases for it.


mib1185 (approved):
Thanks @steffenrapp 👍

Issue #976, opened 2023-02-10
support for controlling entities in an area based on a partial name


ruifung:
for example
entity name: person room ceiling lights
area: person room

sentence: turn on the ceiling lights in person room

I notice that at present, it expects a mostly exact match with the entity name or aliases, and this sentence form would be more natural in my opinion.

Not sure if this is the right place to be asking this though.

dontinelli:
As a workaround you could define an alias "ceiling lights". I assume this should work, as long as you indeed only call the intent with the area as well.

ruifung:
You sure about that?
I checked https://github.com/home-assistant/intents/blob/main/sentences/en/homeassistant_HassTurnOn.yaml and it doesnt seem to support areas for generic entities.

And it definitely doesnt work.
![SmartSelect_20230213_082025_Telegram](https://user-images.githubusercontent.com/1073945/218346175-70f464ca-42d5-44de-b72c-66dfffc38e15.jpg)

dontinelli:
> You sure about that? I checked https://github.com/home-assistant/intents/blob/main/sentences/en/homeassistant_HassTurnOn.yaml and it doesnt seem to support areas for generic entities.
> 
> And it definitely doesnt work. ![SmartSelect_20230213_082025_Telegram](https://user-images.githubusercontent.com/1073945/218346175-70f464ca-42d5-44de-b72c-66dfffc38e15.jpg)

It should work, if you define an alias "alcove lights"  (or in the first example "ceiling lights") for person room ceiling lights. So no support for generic entities but you can rename entities to match your needs.

ruifung:
so the home assistant matching logic can match "area" "entity" as an entity name?
Well, it's good to know I guess.

But IMO, <action> <entity> in <area> should probably be suported too, if only for consistency with the other actions. And feels more natural in an english sentence to me. But that's probably out of scope of this particular issue I suppose.

And I still think having some kind of partial/fuzzy matching for entities would probably be beneficial for the user experience. Assuming it doesn't perform an action on a low-confidence match I suppose.

synesthesiam:
Partial/fuzzy matching will come later when we use the sentences from this repo to train actual machine learning models. For example: generating all possible sentences and using [snips-nlu](https://snips-nlu.readthedocs.io/en/latest/) to train an English model.

PR #1993, opened 2024-02-24
[PL] Add basic weather sentences

deejay1:
Add basic support for fetching weather


Uriziel01 (dismissed):

Uriziel01:
@witold-gren I think the weather responses here are more polished than in https://github.com/home-assistant/intents/pull/1996, I think we should merge this and just overwrite some other changes where it makes sense.

deejay1:
> @witold-gren I think the weather responses here are more polished than in #1996, I think we should merge this and just overwrite some other changes where it makes sense.

@Uriziel01 I could do the weather upgrades from #1996 in here, WDYT?

Uriziel01:
> > @witold-gren I think the weather responses here are more polished than in #1996, I think we should merge this and just overwrite some other changes where it makes sense.
> 
> @Uriziel01 I could do the weather upgrades from #1996 in here, WDYT?

I would do it the other way around and migrate changes from here into #1996 as that PR is absolutely enormous and I would like to avoid conflicts as much as possible just to be safe.

deejay1:
@Uriziel01 surę, review comments fixed, waiting for workflow approvals :/

sentences/pl/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,10 @@
+language: pl
+intents:
+  HassGetWeather:
+    data:
+      - sentences:
+          - "<what_is> pogoda"
+      - sentences:
+          - "<what_is> pogoda (w|dla) <name>"
```
witold-gren, line 8: Please add extra word `we` .. correct sentence is say: "Jaka jest pogoda we Wrocławiu" 😀

tests/pl/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,25 @@
+language: pl
+tests:
+  - sentences:
+      - "jaka jest pogoda"
+    intent:
+      name: HassGetWeather
+    response: 8 °C i pada deszcz
+
+  - sentences:
+      - "jaka jest pogoda w Warszawa"
```
witold-gren, line 10: The correct sentence should be "jaka jest pogoda w Warszawie"

tests/pl/weather_HassGetWeather.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - "jaka jest pogoda dla Warszawa"
+
+    intent:
+      name: HassGetWeather
+      slots:
+        name: Warszawa
+    response: 24 °C bez zachmurzenia
+
+  - sentences:
+      - "jaka jest pogoda w Poznań"
```
witold-gren, line 20: The correct sentence should be "jaka jest pogoda w Poznaniu"

witold-gren (changes requested):
good update 💪🏻 I only have a request to make some cosmetic changes

tests/pl/weather_HassGetWeather.yaml
deejay1: I'm not sure you can do automatic declination of entities. This could work with aliases though, but by using default entity names here you match the UI

tests/pl/weather_HassGetWeather.yaml
deejay1: see above, this would need aliases or automatic declination. I'll see if I can set up aliases in test fixtures

sentences/pl/weather_HassGetWeather.yaml
deejay1: done

tests/pl/weather_HassGetWeather.yaml
deejay1: ok, added declinated versions together with the base version also if, I think we should test for most possibilities than can happen (although right now it's just a dumb slot replacement)

tests/pl/weather_HassGetWeather.yaml
witold-gren: Yes, you're right but the future these declensions will be used to create sentences for the ML model. I would prefer it to sound correctly in Polish 😀 check this issue https://github.com/home-assistant/intents/issues/976

witold-gren (approved):

jlpouffier (approved):
Approved automagically, as I am not speaking Polish
I do not see anything out of the ordinary structure-wise.

deejay1:
@Uriziel01 can you reapprove it ?

PR #1995, opened 2024-02-25
[CS] Add sentence for turning on/off lights in the satelite area. Added expansion rule for lights in plural/singular

bruxy70:
Added sentence for turning on all lights in the same area as the same are as the satelite device.
Also added an expansion rule for lights, accepting both plural an dsingular (as suggested in #1983)

Whilst doing that, I made the sentences consistent so they make sense in the CZ:
rozsvitit means turn on the light, zhasnout means turn off the light. So when using these two words, I made the word svetlo (light) optional - you can say rozsvit or rozsvit svetlo, both is correct. But when using more generic zapni/vypni (turn on/turn off), the word light should not be optional, to make it clear we are switching the light, as there might be other devices in the room with the same name. This is consistent with EN.
I made an expansion for lights (<svetla>). But I only use it in the area. The device is always singular, and for all the lights, the wold "uplne" (completely) only makes sense for plural.

### DIFF
--- a/responses/cs/HassTurnOn.yaml
+++ b/responses/cs/HassTurnOn.yaml
@@ -4,6 +4,7 @@ responses:
     HassTurnOn:
       default: "Zapnuto"
       lights_area: "Světla rozsvícena"
+      light_all: "Všechna světla byla rozsvícena"
       fans_area: "Větráky zapnuty"
       fan: "Větrák zapnut"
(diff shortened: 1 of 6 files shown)


halecivo (approved):
Thanks for implementing the suggestion and overall fixing the sentences to make more sense.

PR #1997, opened 2024-02-25
[CS] Add person state

bruxy70:
Adding person_HassGetState.yaml, tests and updating related files.


responses/cs/HassGetState.yaml
```diff
@@ -78,3 +78,10 @@ responses:
 
       how_many: |
         {{ query.matched | length }}
+
+      where: |
+        {% if state == "not_home" %}
+          {{ slots.name | capitalize }} je pryč
+        {% else %}
```
balloob, line 85: You need to process `state == "home"` too.

balloob (approved):
Awesome. 🐬

PR #1999, opened 2024-02-26
Update HassGetWeather.yaml (Partly Cloudy mismatch)

khalob:
When the weather is "Partly Cloudy" the HassGetWeather intent is not correctly matching the map value because it searches for "partly cloudy" instead of "partlycloudy". To keep things consistent, we can replace all spaces with hyphens, since that's how the map keys are formatted.

### DIFF
--- a/responses/en/HassGetWeather.yaml
+++ b/responses/en/HassGetWeather.yaml
@@ -12,7 +12,7 @@ responses:
           'hail': 'with hail',
           'lightning': 'with lightning',
           'lightning-rainy': 'with lightning and rain',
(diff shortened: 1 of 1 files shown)


tetele:
Thanks, but your PR is not enough to fix the issue. First off, you have the same issue which you left unaddressed for `clear-night`, `lightning-rainy`, `snowy-rainy` and `windy-variant`.

But more importantly, the issue runs deeper than this. See https://github.com/home-assistant/intents/discussions/1768 for clarification. We need a better solution which makes both translated and untranslated states available in the responses.

frenck:

../Frenck

PR #2003, opened 2024-02-26
Test all YAML sentence files

balloob:
Previously, for each language, we would only test the files that existed under the English language. Now it will find all files for all languages and make sure they are tested.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ import yaml
 from hassil import Intents
 from hassil.util import merge_dict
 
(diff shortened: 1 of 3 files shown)


balloob:
@tetele I had to comment out some Romanian tests that previously didn't get run in the test suite and contained some test failures.

tetele:
@balloob thanks for the heads up, i will look into it. There were (still are, actually) some issues with how context is provided in tests and i tried to force them to work, maybe that was the case.

I will try to investigate and get them to run again.

PR #2005, opened 2024-02-26
[FR] Add excluded domains in homeassistant_HassTurnOn/Off

slegars56:
I noticed problems on my main HA instance if a script name or alias matches homeassistant_HassTurnOn/Off intents.

This PR aligns with the English setup by excluding the following domains besides cover, vacuum, and valves:
            - binary_sensor
            - lock
            - scene
            - script
            - sensor

PR created as a draft: @jlpouffier, this is food for thought, I don't want to mess up what you're working on with context awareness


jlpouffier:
Even if the PR is in draft, looking at it I see no issue moving forward.

You added these domains in the excluding list:
- binary_sensor
- lock
- scene
- script
- sensor

`binary_sensor` and `sensor` makes sense as you cannot turn them on or off
And `lock`, `script`, `scene` all have corresponding domain files with the correct `requires_context` field.

I don't want to look too much into these files as they are really bad to be honest: We do not use expansion rules for verbs for example.

But one thing at a time.
We're in the process of making a pass on every domain and cleaning the whole thing, we just did not reach yet these domains.

So if you pass this PR ready for review, it will be a 👍 for me.

The next step will be to clean the `lock`, `script`, and `scene` files.

sentences/fr/homeassistant_HassTurnOff.yaml
```diff
@@ -7,6 +7,11 @@ intents:
...
           domain:
+            - binary_sensor
             - cover
+            - lock
+            - scene
+            - script
+            - sensor
             - vacuum
-            - valve
+            - valve
```
jlpouffier, line 17: Add back the empty line to make the linter happy :) 
Your tests are not passing for cosmetic issues

jlpouffier (changes requested):

sentences/fr/homeassistant_HassTurnOff.yaml
slegars56: Thank you, I struggled a little to fix it :)

jlpouffier (approved):

PR #2018, opened 2024-02-28
[FR] Cleaning of scene/script_HassTurnOn

slegars56:
Cleaning of files I made a few weeks ago.

1/ Two new expansion rules to comply with the fr rule (infinitive and imperative forms)
- <active> for scenes
- <execute> for script

2/ Intents split into two rows for readability

3/ Responses updated:
I think we can return more detailed responses:

scene: "Activée" --> "Scène activée"
script: "Démarré"  --> "Script exécuté"

"Script exécuté" might not sound natural or user-friendly. I've got a doubt about this one.

4/ verbs expansion rules sorted by alphanumeric order in _common.yaml


responses/fr/HassTurnOn.yaml
```diff
@@ -6,6 +6,6 @@ responses:
       fans: "Ventilateurs allumés"
       cover: "Ouvert"
       lock: "Fermé"
-      scene: "Activée"
-      script: "Démarré"
+      scene: "Scène activée"
+      script: "Script exécuté"
```
jlpouffier, line 10: I would still keep this feeling of something that is starting but not done.
A script can take time to complete, you can have delays, or even pause until a specific trigger is matched.

I just tried with a simple example:
```yaml
alias: Test script
sequence:
  - delay:
      hours: 0
      minutes: 0
      seconds: 10
      milliseconds: 0
  - service: light.toggle
    metadata: {}
    data: {}
    target:
      entity_id: light.bureau
```

When I say "Lance test script", the response is immediate (Today `Démarré`), but my office light is toggled after some delay, 10 seconds.

Here are a few proposals from me
- Script démarré
- Script lancé

sentences/fr/_common.yaml
```diff
@@ -524,19 +524,21 @@ expansion_rules:
   partout: ([de] partout)|(<dans> [<tous>] [<le>]<maison>)|(<dans> toutes les pièces)
 
   # Verbs
-  regle: "(règle|régler|met|mets|mettre|ajuste|ajuster|change|changer)"
+  active: "(active|activer|joue|jouer|exécute|exécuter)"
```
jlpouffier, line 527: Don;t you find it weird that `active` contains `exécute` and that `execute` is also an expansion rules containing other verbs 🤔 ?

I wonder if we really need to split these two in fact.
The main difference between a scene and a script is that a script is action-based (Do this) and a scene is state-based (Put that device in that state)

But when it comes to triggering them by voice, I wonder if you cannot have a single verb list.

- Joue le script suivant.
- Joue a scène suivante.

It does not bother me to be honest.

(English as the same verb list too ;) )

jlpouffier (changes requested):

responses/fr/HassTurnOn.yaml
slegars56: It makes sense, I thought it was the same as automation with custom responses e.g. triggered after the script completion.

sentences/fr/_common.yaml
slegars56: You're right: the difference between scripts and scenes is subtle. I merge in one unique expansion rule (active).

However, am I correct in this approach? A verb expansion rule is a verb + all synonyms used in a specific context or with specific domains.

It's quite tricky to find the right balance without over-complicating things :)

jlpouffier: I would say an expansion rule should be verbs + synonyms without a particular context.

But I agree with you, the balance is quite complex.
For example, most of the tokens in the expansion rule `<allume>` are now part of `<active>`

I think this is it: A balance.
If we want to be perfectly correct, always, then we should not use expansion rules, and define what verb we want to use in every sentence, but that would make the whole repo too complex to read.

I think today the balance is OK, we do have some weird sentences because of our expansion rule, for example, you can say `démarre la lumière`... That one could be worth adapting a bit our expansion rules.

I do like the idea of having expansion rules per domains... 
I'll think about it cc @piitaya

PR #2045, opened 2024-03-06
[EN] Separate HassSetPosition into domains

tetele:
Following our French colleagues' initiative (#2002), I've split `HassSetPosition` in the 2 `domain`s that support it, so that we can have greater detail for the `cover` domain, such as targeting a specific `device_class` (which is the only supported particularization, for now).


balloob (approved):
Thanks. For follow up PR, can we get some more sentences that hardcode the word blinds instead of name etc

tetele:
> can we get some more sentences that hardcode the word blinds

I'm not sure I follow @balloob 

Doesn't `{cover_classes:device_class}` do just that without the hard-coding? What's indeed missing is area-awareness (i.e. context-to-slot injection for both referencing by just `name` and just `device_class`) and maybe some area-agnostic `device_class` calls, although who says "set all blinds everywhere to 60%"?

v1k70rk4:
@tetele When I leave home, I set all the blinds to 10% so that even the cat can see something :D
I do this with a automation, but it could well have a justification.

PR #2072, opened 2024-03-10
vacuum intents creation

dziadeck47:
Added vaccuum intents and excluded vacuum domain from generic hassturnon (as it was catching the vacuum turn on incorrectly)

### DIFF
--- a/responses/pl/HassVacuumReturnToBase.yaml
+++ b/responses/pl/HassVacuumReturnToBase.yaml
@@ -1,0 +1,5 @@
+language: pl
+responses:
+  intents:
+    HassVacuumReturnToBase:
+      default: "Powrót do bazy"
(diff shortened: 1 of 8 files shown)


sentences/en/vacuum_HassVacuumReturnToBase.yaml
```diff
@@ -1,4 +1,4 @@
-language: en
+language: pl
```
tetele, line 1: Modified the wrong file here

tetele (changes requested):

Uriziel01:
@dziadeck47 This PR is duplicating most of the changes from https://github.com/home-assistant/intents/pull/1996/files isn't it?

frenck:

../Frenck

PR #2063, opened 2024-03-08
Add more media player service calls

OzGav:

PR #2080, opened 2024-03-13
[en] Add previous_track intent

OzGav:
These have been removed from here https://github.com/home-assistant/intents/pull/2063 to make it more managable. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new intent, `HassMediaPrevious`, to control media players and skip to the previous item.
  - Added English language response for the `HassMediaPrevious` intent with the default message "Playing previous".
  - Included various command phrases for replaying the previous track in a media player context.
  - Implemented test cases for the `HassMediaPrevious` intent related to media playback control.


intents.yaml
```diff
@@ -380,3 +380,63 @@ HassSetVolume:
...
+    name:
+      description: "Name of a device or entity"
+      required: false
+    area:
+      description: "Name of an area"
+      required: false
+  slot_combinations:
+    name_only:
+      - "name"
+    area_only:
```
tetele, line 418: Are you sure it's OK to "clear playlist in the living room"? Shouldn't this be name-only? You haven't even added sentences for any other scenario

responses/en/HassMediaPrevious.yaml
```diff
@@ -0,0 +1,5 @@
+language: en
+responses:
+  intents:
+    HassMediaPrevious:
+      default: "Going back"
```
tetele, line 5: I think this sounds ambiguous. Maybe "Playing previous", for consistency with `HassMediaNext`?

sentences/en/media_player_HassMediaClearPlaylist.yaml
```diff
@@ -0,0 +1,8 @@
+language: en
+intents:
+  HassMediaClearPlaylist:
+    data:
+      - sentences:
+          - "(clear;<name>) (queue|playlist)"
```
tetele, line 6: Would "delete" work as a verb here as well?
Also, I'd add `clear [the ](queue|playlist) <in> <name>`

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,10 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "Go back [to the (previous|last) (song|track)] on <name>"
```
tetele, line 6:
```suggestion
          - "go back [to the (previous|last) (song|track) ]on <name>"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,10 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
```
tetele, line 5: Maybe also use "play"/"replay" as verbs.

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,10 @@
...
+          - "Go back [to the (previous|last) (song|track)] on <name>"
+          - "<name> go back [to the (previous|last) [(song|track)]]"
+          - "<name> play the (previous|last) [(song|track)] [again]"
```
tetele, line 8:
```suggestion
          - "<name> play [the ](previous|last)[ (song|track)][ again]"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,10 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "Go back [to the (previous|last) (song|track)] on <name>"
+          - "<name> go back [to the (previous|last) [(song|track)]]"
```
tetele, line 7:
```suggestion
          - "<name> go back[ to the (previous|last)[ (song|track)]]"
```

sentences/en/media_player_HassMediaPrevious.yaml
tetele: And I think just `previous track on <name>` should work, without the verb

tetele (changes requested):
Thanks!

sentences/en/media_player_HassMediaClearPlaylist.yaml
OzGav: fixed

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: fixed

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: fixed

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: Have a look at what I have changed here and see if I have captured what you were thinking.

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: fixed

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: Not sure about omitting the verb, that wouldn't be correct grammat and I thought that was desired....

tetele: @synesthesiam any thoughts on this?

intents.yaml
tetele: I agree about the area awareness, that most intents should act immediately on one/all of the targeted entities in the satellite's area.

However, this may be a bit more tricky, as clearing/deleting is a destructive action, with potentially frustrating implications. Imagine the scenario where you have 2 `media_player`s and a voice satellite in the `office` area. You say `clear the playlist` but forget you have 2 players. The first `media_player` matches and gets its playlist cleared, but you actually wanted to perform the action on the second. That sounds annoying.

What I'd do is that for this intent (and any other destructive/irreversible ones) if that I'd only allow targeting by device name so as to prevent accidental triggerings. Maybe in the future we will have a short conversation for confirmation (`Are you sure you want to clear the playlist on <media player name>?` `Yes` `Cleared`).

tetele:
@OzGav I think you forgot to push the fixes, I can't see anything new, so I can't review anything new :D

OzGav:
I am still a github noob and I was working on the wrong branch. Fixed now (I thnk!)

tetele (commented):
Except for the comments that are still open, I'm good with this.

Being a set of new intents, we will wait for Mike's input.

OzGav:
Just gently checking in as to whether there is any news?

intents.yaml
OzGav: @tetele I see in the latest beta area support. Are there changes I should make to this PR now as a result of that?

tetele:
@synesthesiam in the context of 2024.6, will you consider this PR too?

intents.yaml
OzGav: Thanks, looking at it now. Looking at Mike's recent PRs it seems I should remove slot combinations throughout intents.yaml and add similar sentence variations to what he has done for each of .

synesthesiam:
I'm torn on this. Maybe @jlpouffier will have more thoughts?

First, I'm hesitant to have both "pause" and "stop" intents for media players since it's not always clear how these are different. Sometimes, you can resume from a stopped state; sometimes, you can't. It's confusing, and so I've avoided it by only including "pause" :see_no_evil: 

Second, we need to be careful about introducing too many intents since this puts a translation burden on language leaders, as well as a maintenance burden in HA. Our original goal was to cover the "most common" voice assistant use cases, but there's no formal definition of what this means.

However we define it, I think we have a problem right now with intents requiring code inside HA and the sentences not being sharable in a blueprint style. So we can't have a set of community-supported intents that could get promoted to "official" status given enough use :slightly_frowning_face: 

What do you think?

OzGav:
I wasn't across the desire to only have the "most common" intents so I approached this by creating intents for all the media player service calls. I help out with the Music Assistant project and we have shared custom sentences that cover all of these (including a form of area awareness). I then thought to bring these across to core and so here we are. If HA will remain only supporting this subset of calls that is OK we will just amend our docs to advise everyone of this fact and that they should use the supplied sentences or create their own.

As for pause and stop you are correct. In MA "stop" will stop playback and clear the queue whereas pause just stops playback

intents.yaml
jlpouffier: We should allow targeting areas for all media players. We already have custom logic for play, pause, next and set volume :) 

We just need to think about what could be the logic for clearing a queue ☺️

OzGav: Yes I have made changes to the previous command to (I think) support the area awareness of 2024.6.0. The question you will see below is whethertheese commands should be supported by core at all?

jlpouffier:
I agreed with @synesthesiam unfortunately.

Out of the 3 new intents proposed, I think only "Previous" makes sense to add to core for now.
It's simple and when I think about it, since we have Next, why not have Previous? This won't add too much of a burden on our translator, and the area targeting logic is already in place for Next. We just need to copy it.

Stop colliding with pause too much - some languages already added sentences to "Stop the playback" for the "Pause" intent.
I think for a voice-specific interface, the only thing that matters for the user is to stop the playback now (ie. Pause it). 
Anything more specific (Clearing the queue, etc. Needs to be done via a UI component for now)

tetele:
This discussion revives an older idea/question of mine: should (custom) integrations be allowed to define their own intents/sentences? So when loading the intent set for the default conversation agent, should we allow each integration to register its own set?

What do you think @synesthesiam @jlpouffier ?

OzGav:
FYI there is a plan to move the MA Integration out of HACS and into core sometime this year... Not sure how/if that changes anything?

OzGav:
OK PR pared back to just the `media_previous_track` service call

OzGav:
Merge conflict resolved

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: Just this question remains

intents.yaml
```diff
@@ -377,6 +377,18 @@ HassSetVolume:
...
+  domain: media_player
+  description: "Skips a media player back to the previous item"
+  slots:
+    name:
+      description: "Name of a device or entity"
+      required: false
+    area:
+      description: "Name of an area"
+      required: false
+
```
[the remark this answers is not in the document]
OzGav: Awaiting other PR

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - sentences:
+          - "(go back[ to the (previous|last) (song|track)];[on ] <name>)"
+          - "replay[ the (previous|last) (song|track)] on <name>"
+          - "<name> (play|replay) [the ](previous|last)[ (song|track)][ again]"
+        requires_context:
+          domain: media_player
+      - sentences:
+          - "go back[ to the (previous|last) (song|track)]"
+          - "replay[ the (previous|last) (song|track)][ again]"
+          - "play[ the] (previous|last) [(song|track) ][again]"
```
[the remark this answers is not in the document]
OzGav: tweaked space placement per existing next intent

OzGav (commented):
Ready for further review

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "(go back [to the (previous|last) (song|track)];[on ]<name>)"
```
tetele, line 6:
```suggestion
          - "(go back[ to the (previous|last) (song|track)];[on ]<name>)"
```

tetele (changes requested):
Sorry for the long delay again.

Generally speaking, whitespace should be part of the optional part, i.e. `unlock [the ]{name}` instead of `unlock [the] {name}`. This also prevents inefficiency, as internally, the 2 spaces before and after `[the]` would have to match a single space in the transcribed sentence, which is slower than it having to match a single whitespace.

Apart from that, I don't see anything which would need obvious fixing.

sentences/en/media_player_HassMediaPrevious.yaml
OzGav: Thanks @tetele I have rolled that in.

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "(go back[ to the (previous|last) (song|track)];[on ]<name>)"
+          - "replay [the (previous|last) (song|track)] on <name>"
```
tetele, line 7:
```suggestion
          - "replay [the (previous|last) (song|track) ]on <name>"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+          - "<name> (play|replay) [the ](previous|last) [(song|track) ][again]"
```
tetele, line 8:
```suggestion
          - "<name> (play|replay) [the ](previous|last)[ (song|track)][ again]"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+        requires_context:
+          domain: media_player
+      - sentences:
+          - "go back [to the (previous|last) (song|track)]"
```
tetele, line 12:
```suggestion
          - "go back[ to the (previous|last) (song|track)]"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+          - "replay [the (previous|last) (song|track) ][again]"
```
tetele, line 13:
```suggestion
          - "replay[ the (previous|last) (song|track)][ again]"
```

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+          - "play [the ](previous|last) [(song|track) ][again]"
```
tetele, line 14:
```suggestion
          - "play [the ](previous|last)[ (song|track)][ again]"
```

tetele (changes requested):

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+          - "go back[ to the (previous|last) (song|track)]"
+          - "replay[ the (previous|last) (song|track)][ again]"
+          - "play[ the] (previous|last) [(song|track) ][again]"
+        requires_context:
+          area:
+            slot: true
+      - sentences:
+          - "go back[ to the (previous|last) (song|track)] [in ]<area>"
+          - "(replay[ the (previous|last) (song|track)][ again];[in ]<area>)"
+          - "(play[ the] (previous|last) [(song|track) ][again];[in ]<area>)"
```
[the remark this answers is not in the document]
tetele: The semicolon within parantheses is used to signal a permutation, i.e. any order of the members either side of the semicolon.

tetele:
@OzGav please resolve the merge conflicts so we can finally merge this

sentences/en/media_player_HassMediaPrevious.yaml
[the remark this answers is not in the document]
tetele: Here's [the documentation for the sentence template syntax](https://developers.home-assistant.io/docs/voice/intent-recognition/template-sentence-syntax#sentence-templates-syntax), including the permutations I've mentioned, represented as permutable members between parentheses, separated by semicolons.

OzGav (commented):
Thanks.

OzGav (commented):
Thanks again.

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
...
+          - "go back[ to the (previous|last) (song|track)]"
+          - "replay[ the (previous|last) (song|track)][ again]"
+          - "play [the ](previous|last)[ (song|track)][ again]"
+        requires_context:
+          area:
+            slot: true
+      - sentences:
+          - "go back[ to the (previous|last) (song|track)] [in ]<area>"
+          - "(replay[ the (previous|last) (song|track)][ again];[in ]<area>)"
+          - "(play[ the] (previous|last) [(song|track) ][again];[in ]<area>)"
```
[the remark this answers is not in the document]
OzGav: The semicolon within parantheses is used to signal a permutation, i.e. any order of the members either side of the semicolon.

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "(go back[ to the (previous|last) (song|track)];[on ]<name>)"
+          - "replay [the (previous|last) (song|track) ]on <name>"
+          - "<name> (play|replay) [the ](previous|last)[ (song|track)][ again]"
```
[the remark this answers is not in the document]
OzGav: I think due to the particular arrangement of items in this sentence this is acceptable

sentences/en/media_player_HassMediaPrevious.yaml
```diff
@@ -0,0 +1,21 @@
+language: en
+intents:
+  HassMediaPrevious:
+    data:
+      - sentences:
+          - "(go back[ to the (previous|last) (song|track)];[on ]<name>)"
```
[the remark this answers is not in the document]
OzGav: No this would result in a potentional double space.

OzGav:
@tetele I don't see any merge conflicts. I do see a "Change Requested" but when I click on "See Review" it doesn't take me to anything that is unresolved? I am not sure what to do from here although I have moved it out of draft again.

tetele (approved):

tetele:
I've approved the PR, but i will ask @synesthesiam to take a final look before merging, as this is a new intent.

synesthesiam:
Looks good, thanks!

PR #2081, opened 2024-03-13
[CA] Update sentences for opening the lights

edurenye:
The same way that you can light on or open a light, in Catalan you can 'encendre' or 'obrir' the lights.

It was set like this for singular, but not for plural. This was making me be able to close all the lights but just be able to close one light, not all of them. (In some regions we just say 'obrir' not 'encendre')

The singular was added later in https://github.com/home-assistant/intents/pull/804 that is why this was missed.

### DIFF
--- a/sentences/ca/light_HassTurnOn.yaml
+++ b/sentences/ca/light_HassTurnOn.yaml
@@ -3,9 +3,9 @@ intents:
   HassTurnOn:
     data:
       - sentences:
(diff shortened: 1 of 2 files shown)


duhow (commented):
Pots afegir un parell de tests pel llum, si us plau?

edurenye:
He actualitzat els tests @duhow

PR #2101, opened 2024-03-21
[PT] Improve cover intent

marcocunha:
Add 'Estores'

### DIFF
--- a/sentences/pt/_common.yaml
+++ b/sentences/pt/_common.yaml
@@ -79,9 +79,9 @@ lists:
         out: switch
   cover_states:
     values:
-      - in: "(aberto|aberta|abertas)"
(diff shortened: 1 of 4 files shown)


marcocunha (commented):
I added a test case, but it fails, so I added it as a comment. Can you please check what might be the reason?

abmantis:
Thanks for the improvements!
Unfortunately the cover lists had some issues, since they cannot have duplicated "in" values for the same "out". 
While testing this out, I've ended up updated them to match the ones that are currently displayed by HA UI in PT, and fixing your test. Can you check my patch please? I don't have permissions to apply it to your branch.

marcocunha:
That's great @abmantis . Just updated this pull request.

abmantis (approved):

Issue #2112, opened 2024-03-30
[FR] Issue turning off the lights


crazyactions:
Hi,

I'm currently testing the intents for my home assistant, and it's performing well when it comes to turning things on, with "allume" being easily recognized. However, I'm encountering difficulties when trying to turn things off. For instance, when I say "éteins," it seems to interpret it as "Etais," "été," "Et t'as," or "état." How can we incorporate these variations into the standard intent to improve its service ?

Thank you

jlpouffier:
ON it.

piitaya:
Fixed with https://github.com/home-assistant/intents/pull/2120

PR #2120, opened 2024-04-03
[FR] Floor basic support for lights TurnOn and TurnOff

jlpouffier:
## Floor supports
Floors are now supported in French. (A basic support for now)

Please note that, as of today, the only intent supported by `floor` is `HassTurnOn` and `HassTurnOff`.
The only slot combination is `domain` + `floor`
This restricts what is possible.

This PR adds support for floors to lights.
- Turning on all lights on a floor
- Turning off all lights on a floor 

This is it for now. 
Once more intents, and slot combinations are supported, more sentences will come.

## Dirty hack support for the word "éteins" 
I am personally not a fan of introducing dirty hacks in our sentences, but this one is big, and is starting to get attention (https://github.com/home-assistant/intents/issues/2112).

The word "éteins" (Which means "Turn off") is often misunderstood as other words (étant among others)
I decided to create a "dirty" expansion rule and append it to the clean one.

I, on purpose, decided not to add the misunderstood words on the clean expansion rule because I want this gone in the future, once we find a better STT engine, or can perform some sort of fuzzy matching of words.

As of today, these words are now understood as a turn off action on top of the "Clean" verbs
```yaml
eteins_dirty: "(étant|étends|étend|étendre|état|et tant|et teins|et teint|et teints|et t'as|été|étais|était)"
```

### DIFF
--- a/sentences/fr/_common.yaml
+++ b/sentences/fr/_common.yaml
@@ -101,7 +101,7 @@ responses:
       {%- endif %}
     no_entity: "Désolé, je ne connais pas l'appareil {{ entity }}"
     no_entity_in_area: "Désolé, je ne connais pas l'appareil {{ entity }}"
-    no_floor: "Désolé, je ne connais pas l'étage {{ area }}"
+    no_floor: "Désolé, je ne connais pas l'étage {{ floor }}"
     no_domain_in_floor: |
       {% set translations = {
         "button": "de boutons",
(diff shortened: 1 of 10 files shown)


sentences/fr/_common.yaml
```diff
@@ -101,7 +101,54 @@ responses:
...
+        Désolé, je n'ai rien trouvé de correspondant dans cet étage
+      {%- endif %}
+    no_device_class_in_floor: |
+      {% set translations = {
+        "awning": "d'auvents",
+        "blind": "de stores",
+        "curtain": "de rideaux",
+        "door": "de portes",
+        "garage": "de portes de garage",
+        "gate": "de portes",
```
ElieMillon, line 141: Why translate "gate" as "porte" instead of "portail" ?

piitaya: You're right, it should be "portail" 👍

PR #2121, opened 2024-04-03
[DE] Add initial support for floors

steffenrapp:
German support for floors, see #2111

### DIFF
--- a/responses/pt-br/HassTurnOff.yaml
+++ b/responses/pt-br/HassTurnOff.yaml
@@ -11,3 +11,4 @@ responses:
       cover_device_class: "Fechado {{ slots.device_class }}"
       lock: "Destrancado"
       valve: "Fechada"
+      lights_floor: "Luzes apagadas"
--- a/responses/pt-br/HassTurnOn.yaml
+++ b/responses/pt-br/HassTurnOn.yaml
@@ -13,3 +13,4 @@ responses:
       script: "Iniciado {{slots.name}}"
       lock: "Trancado"
       valve: "Aberta"
+      lights_floor: "Acendendo luzes"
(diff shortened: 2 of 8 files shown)


sentences/de/_common.yaml
```diff
@@ -229,6 +237,7 @@ expansion_rules:
   artikel: "[<artikel_bestimmt>|<artikel_unbestimmt>]"
   name: "<artikel> {name}"
   area: "[in|an|auf|bei] [<artikel>|im|am|<von_dem>] {area}[s]"
+  floor: "[in|auf] [<artikel>|im|<von_dem>] [Stock[werk]|Etage|Geschoss] {floor}"
```
mib1185, line 240: i'm not sure, if we really need `[Stock[werk]|Etage|Geschoss]` in here (_you can see it on the weird sounding test sentences_ 🙈 ) - we may just keep `[Geschoss]` as suffix to allow a floor called `Keller` be also targeted by "Keller Geschoss" 🤔 

```suggestion
  floor: "[in|auf] [<artikel>|im|<von_dem>] {floor} [Geschoss]"
```

mib1185 (changes requested):
Hi @steffenrapp 
thanks for this 👍 just one comment about (_IMHO_) unneeded additional extensions.

sentences/de/_common.yaml
steffenrapp: What if someone calls his floors just by numbers - like "2"? 😄 Then it would be useful to have it as a prefix but I don't insist on having it 😃

mib1185: I was also thinking about just numbered floor names ... in English it is easy with `{floor} [floor]` but thanks to the German complex grammar we need to have something like `[Stock[werk]|Etage|Geschoss] {floor}[. Stock[werk]|Etage|Geschoss]` 🙈

PR #2123, opened 2024-04-04
[PT-BR] Initial support for floors

luyzfernando08:


bernabap:
Hi Fernando, could you take a look at these sentences. When we say a sentence like "Ligar a luz", it's treating like if we want to turn on all lights from the house. I was thinking that to turn on all lights from the house we should be forced to explicitly say something like "da casa" in the end of the sentence. Maybe create a expansion rule and replace the [da casa] with <na_casa>?

na_casa: "(n[o|a]|d[a|o]) (casa|apartamento|loft)"

Thank you!

PR #2136, opened 2024-04-13
[DE] add sentences for activating scenes in the area of a satellite device

andreasbrett:
Added awareness of a scene's area based on the area of the triggering satellite device. This might be controversial since this has not yet been added to any other language (at least not EN) and raises the question if scenes should be required to be assigned to an area for them to be triggered via Assist.

Intents for _lights_ require that a light must be in an area. Sentences without an area will then use the satellite's area, sentences with an area will target lights in that area. For _scenes_ this is different, since scenes control not only lights but also other domains. It needs to be discussed if we can require scenes to be assigned to an area for them to be triggered via Assist (just like lights).

For now I circumvented this by requiring the word "hier" to use the satellite's area for targeting. I'm not particularly fond of this though. I'd rather have scenes behave as lights so I can simply ask "aktiviere partymodus" instead of "aktiviere partymodus hier". There might be objections though for those that use scenes area-independent (is that a thing?).


mib1185 (changes requested):
Hi @andreasbrett 
I think this should be discussed first in the [ideas](https://github.com/home-assistant/intents/discussions/categories/_ideas) or [general](https://github.com/home-assistant/intents/discussions/categories/_general) section of the discussions board

frenck:

../Frenck

andreasbrett:
Re-opening because this is on halt because of missing feedback from @synesthesiam. I'd be happy to "pick up work" anytime but have been waiting for months for a response in https://github.com/home-assistant/intents/discussions/2147.

PR #2141, opened 2024-04-17
[FI] Don't exclude script from the generic HassTurnOn

daniellandau:
There doesn't seem to be any way currently to start scripts via voice, and "käynnistä" is a very natural Finnish language word to use to start a script.

### DIFF
--- a/sentences/fi/homeassistant_HassTurnOn.yaml
+++ b/sentences/fi/homeassistant_HassTurnOn.yaml
@@ -15,6 +15,5 @@ intents:
             - cover
             - lock
             - scene
-            - script
             - sensor
             - valve


daniellandau:
Yeah adding aja as an additional option is good, but IMO käynnistä makes sense to work for a script too

salleq:
I had little time to check this out. Technically I think your solution might work, but I'd rather follow how it's done in English and other translations to have similar file structure - to translate the HassTurnOn intent for scripts in a different file and make own tests for it. 
I don't think it should take too long to do these - I can do it hopefully in the weekend if I find the time, or if you know how to do everything needed go ahead. This isn't the only intent that needs love, but translators have been a bit scarce :D

salleq:
@daniellandau I basically did the changes already but I'm still a git noob, so I don't know how to add them to this PR. Right now I have the changes in my fork. Can I somehow merge my changes to this PR or should I just close this one and create a new PR from my changes?

daniellandau:
Probably easiest for you to just make them in a new pull request.

salleq:
Allright, I'll close this one.

PR #2152, opened 2024-04-23
[he] added female pronounce to `sentences`

ofirsnb:
In Hebrew,  there are different pronounces when addressing male or female.
So, if my wake-word is "Alexa" (female),  I need to speak to her in a female form.
Example for non-Hebrew speakers - the term `Turn On` is different for male (`הדלק`) and female (`הדליקי`) in Hebrew.
Currently,  the Hebrew intents are set to be called in a male only form.
This PR adds female form as well,  so when calling "Alexa" for example,   the rest of the conversation can be natural.
Note - this PR does **not** alter the `responses`  -  which will remain in neutral form.
Thank you.


leranp (approved):
Thank you for helping with the Hebrew project 😀

ofirsnb:
Thank you very much for the (very) quick review! 🙏

After about a year using Assist in English,  I want my household to use it as well,  but they're not English speakers. 

So just as heads up 😃 - I'm gonna work on the Hebrew further to add more capabilities (domains and such),  my goal for it is to be as close to English as possible. 

Thanks again and happy holiday :)

PR #2154, opened 2024-04-28
[pt] lights room awareness

P2mb2:


sentences/pt/light_HassTurnOff.yaml
```diff
@@ -6,3 +6,12 @@ intents:
           - "<desliga> [(todas | todos)] [(o | os | a | as)] (luz | luzes | candeeiro | candeeiros) <na_zona>"
         slots:
           domain: "light"
+
+      # Desligar luzes na mesma divisão que um dispositivo de voz
+      - sentences:
+          - "<desliga> [(todas | todos)] [(o | os | a | as)] (luz | luzes | candeeiro | candeeiros)"
```
abmantis, line 12: I think this sentence will be better for turning all lights, not only the ones in the current area.
Maybe add "aqui"? "Desliga todas as luzes aqui"
Also, a separate one could be in the singular form: "Desliga a/o luz/candeeiro", since that is a common way people use to refer to the light in the current area.

abmantis (changes requested):

abmantis (approved):
Obrigado!

PR #2161, opened 2024-04-29
[DE] Added areas to homeassistant_HassTurnOn.yaml

SaftigeSteaks:
Now sentenses like "Schalte den Computer im Wohnzimmer an" will work


SaftigeSteaks:
While testing within Home Assistant it doesn't like these lines:

          - [<area>] <name> <aktivieren>
          - [<area>] <name> <an>[<schalten>|<machen>]

so I closed pull request

PR #2180, opened 2024-05-20
Add enhanced media sentences for English

synesthesiam:
HA now supports "enhanced" media control via voice (see https://github.com/home-assistant/core/pull/115445).

This will use the area/floor of the voice satellite and state/features of media players to figure out which media player(s) to target (when no name is provided). For example, "pause" will pause any playing media players in the current area.


tetele (approved):
LGTM

PR #2182, opened 2024-05-21
Add timer and state/feature errors

synesthesiam:
Add error messages for timers and state/feature

## Summary by CodeRabbit

- **New Features**
  - Introduced new error codes for enhanced validation: "entity_wrong_state," "feature_not_supported," "timer_not_found," "multiple_timers_matched," and "no_timer_support."
  - Added corresponding error messages to provide clearer feedback for users encountering these issues.

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -17,7 +17,7 @@ responses:
     no_entity: "Sorry, I am not aware of any device called {{ entity }}"
     no_entity_in_area: "Sorry, I am not aware of any device called {{ entity }} in the {{ area }} area"
     no_entity_in_floor: "Sorry, I am not aware of any device called {{ entity }} in the {{ floor }} floor"
-    entity_wrong_state: "Sorry, no device is in the expected state"
+    entity_wrong_state: "Sorry, no device is {{ state | lower }}"
     feature_not_supported: "Sorry, no device supports the required features"
 
     # Errors for when user is logged in and we can give more information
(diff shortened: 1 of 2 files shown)


tetele:
What's the functionality here?

synesthesiam:
There are a few new errors possible now in HA for timers and media players, so I've added English "translations" here.

The timer errors are straightforward. The state/feature errors are for when you target media players without a name. If you say "pause" but no media players are playing, that's a state error. If you try to set the volume of a media player that doesn't support it, that's a feature error.

sentences/en/_common.yaml
```diff
@@ -17,6 +17,8 @@ responses:
     no_entity: "Sorry, I am not aware of any device called {{ entity }}"
     no_entity_in_area: "Sorry, I am not aware of any device called {{ entity }} in the {{ area }} area"
     no_entity_in_floor: "Sorry, I am not aware of any device called {{ entity }} in the {{ floor }} floor"
+    entity_wrong_state: "Sorry, no device is in the expected state"
+    feature_not_supported: "Sorry, no device supports the required features"
```
tetele, line 21: These feel very clunky as responses. I wouldn't have the slightest clue what I can do to fix anything if I tried to pause playback and I got back a "Sorry, no device is in the expected state".

Any chance they can be overridden at sentence/intent level with some `metadata` and keep these just as defaults? I feel like "Sorry, no device is currently playing" sounds much clearer.

synesthesiam: Hmmm, I could slip the expected state in there. I'll need to translate it, of course.
The features is more of a problem, as there don't seem to be strings for them anywhere I can find.

tetele:
Is this not perfect `metadata` territory? Or is the `metadata` not passed to error responses?

synesthesiam:
It could be as well. So you're suggesting that we add something like this to sentences:

```yaml
intents:
  HassSetVolume:
    data:
      - sentences:
          - "set volume to 100"
        metadata:
          feature: "set volume"
```

and then in the error:

```yaml
responses:
  errors:
    feature_not_supported: "Sorry, no device supports the feature {{ metadata.feature }}"
```

Is that correct?

tetele:
Yup, that's what i had in mind. I think it's a better approach, and easily translatable by language leaders.

PR #2219, opened 2024-06-06
[EN] Add more english sentences for lock get state queries

ViViDboarder:
Add some new permutations of lock sentence queries.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the `HassGetState` intent to support more flexible and varied sentence structures for querying the state of locks, doors, and windows in specific areas.

- **Improvements**
  - Refined language for querying lock states to improve user interaction and clarity.

### DIFF
--- a/sentences/en/lock_HassGetState.yaml
+++ b/sentences/en/lock_HassGetState.yaml
@@ -3,7 +3,7 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {lock_states:state} [in <area>]"
+          - "(is|are) <name> {lock_states:state} [in <area>]"
         response: one_yesno
         requires_context:
           domain: lock
@@ -11,26 +11,35 @@ intents:
           domain: lock
 
       - sentences:
-          - "(is|are) [there] any door[s] {lock_states:state} [in <area>]"
-          - "(is|are) [there] any {lock_states:state} door[s] [in <area>]"
+          - "(is|are) [there] any (door|lock|window)[s] {lock_states:state} [in <area>]"
+          - "(is|are) [there] any {lock_states:state} (door|lock|window)[s] [in <area>]"
+          - "do I have any {lock_states:state} (door|lock|window)[s] [in <area>]"
         response: any
         slots:
           domain: lock
 
       - sentences:
-          - "are all [the] door[s] {lock_states:state} [in <area>]"
+          - "(is|are) all [the] (door|lock|window)[s] [securely] {lock_states:state} [in <area>]"
         response: all
         slots:
           domain: lock
 
       - sentences:
-          - "(which|what) door[s] (is|are) {lock_states:state} [in <area>]"
+          - "(which|what) (door|lock|window)[s] (is|are) [currently] {lock_states:state} [in <area>]"
         response: which
         slots:
           domain: lock
 
       - sentences:
-          - "how many door[s] (is|are) {lock_states:state} [in <area>]"
+          - "how many (door|lock|window)[s] (is|are) {lock_states:state} [in <area>]"
         response: how_many
+        slots:
+          domain: lock
+
+      - sentences:
+          - "tell me the status of <name> [in <area>]"
+        response: one
+        requires_context:
+          domain: lock
         slots:
           domain: lock
--- a/sentences/en/lock_HassGetState.yaml
+++ b/sentences/en/lock_HassGetState.yaml
@@ -3,8 +3,7 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "(is|are) <name> [<currently>] [securely] {lock_states:state} [in <area>]"
-          - "(is|are) <name> [securely] {lock_states:state} [in <area>] [<currently>]"
+          - "(is|are) <name> ([<currently>];{lock_states:state} [in <area>])"
         response: one_yesno
         requires_context:
           domain: lock
@@ -12,31 +11,26 @@ intents:
           domain: lock
 
       - sentences:
-          - "((is|are) [there]|do I have) any <lockable> [<currently>] [securely] {lock_states:state} [in <area>]"
-          - "((is|are) [there]|do I have) any <lockable> [securely] {lock_states:state} [in <area>] [<currently>]"
-          - "((is|are) [there]|do I have) any [<currently>] [securely] {lock_states:state} <lockable> [in <area>]"
-          - "((is|are) [there]|do I have) any [securely] {lock_states:state} <lockable> [in <area>] [<currently>]"
+          - "((is|are) [there]|do I have) any <lockable> ([<currently>];{lock_states:state} [in <area>])"
+          - "((is|are) [there]|do I have) any ([<currently>];{lock_states:state} <lockable> [in <area>])"
         response: any
         slots:
           domain: lock
 
       - sentences:
-          - "are all [the] <lockable> [<currently>] [securely] {lock_states:state} [in <area>]"
-          - "are all [the] <lockable> [securely] {lock_states:state} [in <area>] [<currently>]"
+          - "are all <lockable> ([<currently>];{lock_states:state} [in <area>])"
         response: all
         slots:
           domain: lock
 
       - sentences:
-          - "(which|what) <lockable> (is|are) [<currently>] [securely] {lock_states:state} [in <area>]"
-          - "(which|what) <lockable> (is|are) [securely] {lock_states:state} [in <area>] [<currently>]"
+          - "(which|what) <lockable> (is|are) ([<currently>];{lock_states:state} [in <area>])"
         response: which
         slots:
           domain: lock
 
       - sentences:
-          - "how many <lockable> (is|are) [<currently>] [securely] {lock_states:state} [in <area>]"
-          - "how many <lockable> (is|are) [securely] {lock_states:state} [in <area>] [<currently>]"
+          - "how many <lockable> (is|are) ([<currently>];{lock_states:state} [in <area>])"
         response: how_many
         slots:
           domain: lock
--- a/sentences/en/lock_HassGetState.yaml
+++ b/sentences/en/lock_HassGetState.yaml
@@ -4,6 +4,7 @@ intents:
     data:
       - sentences:
           - "(is|are) <name> [<currently>] [securely] {lock_states:state} [in <area>]"
+          - "(is|are) <name> [securely] {lock_states:state} [in <area>] [<currently>]"
         response: one_yesno
         requires_context:
           domain: lock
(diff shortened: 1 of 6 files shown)


sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,43 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {lock_states:state} [in <area>]"
+          - "(is|are) <name> {lock_states:state} [in <area>]"
```
tetele, line 6: Could you please add a test sentence for an entity with a plural name?

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,43 @@ intents:
...
+          - "(is|are) [there] any (door|lock|window)[s] {lock_states:state} [in <area>]"
+          - "(is|are) [there] any {lock_states:state} (door|lock|window)[s] [in <area>]"
+          - "do I have any {lock_states:state} (door|lock|window)[s] [in <area>]"
         response: any
         slots:
           domain: lock
 
       - sentences:
-          - "are all [the] door[s] {lock_states:state} [in <area>]"
+          - "(is|are) all [the] (door|lock|window)[s] [securely] {lock_states:state} [in <area>]"
```
tetele, line 22: There is no grammatical context in which "is" can be used with "all". Unless you can provide an example, please remove it.

```suggestion
          - "are all [the] (door|lock|window)[s] [securely] {lock_states:state} [in <area>]"
```

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,43 @@ intents:
...
         response: all
         slots:
           domain: lock
 
       - sentences:
-          - "(which|what) door[s] (is|are) {lock_states:state} [in <area>]"
+          - "(which|what) (door|lock|window)[s] (is|are) [currently] {lock_states:state} [in <area>]"
```
tetele, line 28: Why would we add "currently" only to this sentence and "securely" only to the previous one? Would it not be better to alter the `lock_state` list in order to optionally match these 2 adverbs as well?

tetele (changes requested):
All in all good changes, but there are one or two things which could be improved. Thanks!

sentences/en/lock_HassGetState.yaml
ViViDboarder: Maybe, but `currently` is an adjective that could probably apply to _any_ query. Do you know how `skip_words` are used? They look like they are currently words that may show at the beginning of a query, however there are probably a lot of adverbs that could be added to the list.

* currently
* securely
* presently
* already

tetele: `skip_words` are simply stripped from the input sentence before any matching is done. I'd not add adverbs to that list, as common as they may be

IamTheFij: In that case, possibly an `adverbs` expansion would be useful? I can see your point to not stripping though as that's probably more blunt. It's possible someone may have an adverb in a script name or something.

ViViDboarder: Made some changes to address this. One thing I don't quite like is that to make sentences flexible, it requires two sentences for each intent. One where the adverb is at the front and one where it's at the end. I'm not sure there's a good way around that to make something even more generic.

tetele: [Permutations](https://developers.home-assistant.io/docs/voice/intent-recognition/template-sentence-syntax#sentence-templates-syntax), perhaps?

sentences/en/lock_HassTurnOff.yaml
```diff
@@ -3,14 +3,14 @@ intents:
...
+          - "unlock [the] <name> [in <area>]"
         requires_context:
           domain: lock
         response: lock
 
       - sentences:
-          - "unlock [all] [the] [(lock|door)[s]] [in] <area>"
-          - "unlock [all] <area> [(lock|door)[s]]"
+          - "unlock [all] [the] [<lockable>] [in] <area>"
+          - "unlock [all] <area> [<lockable>]"
```
tetele, line 13: If all optional parts are missing, these 2 sentence templates are the same: `unlock <area>`. Please avoid this.

tetele (changes requested):
I haven't addressed each individual line, but you have 3 types of issues:

1. Removing all optional parts in a set of sentences basically produces the same sentence template. This makes for very inefficient (slow) matching.
2. Whitespace should be part of the optional part, i.e. `unlock [the ]{name}` instead of `unlock [the] {name}`. This also prevents inefficiency, as internally, the 2 spaces before and after `[the]` would have to match a single space in the transcribed sentence, which is slower than it having to match a single whitespace.
3. The `securely` adverb should probably be an optional part of the [input string](https://github.com/home-assistant/intents/blob/0d5de750b97092bb2d703f6f306d211a8c79c0d4/sentences/en/_common.yaml#L127) of the `lock_states` list, rather than included in each sentence. You can't call something `securely unlocked`.

sentences/en/lock_HassGetState.yaml
ViViDboarder: Giving that a try!

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {lock_states:state} [in <area>]"
+          - "(is|are) <name> ([<currently>];{lock_states:state} [in <area>])"
```
tetele, line 6: You can't have a completely optional permutation item, as skipping `<currently>` would match both the optional part at the beginning and at the end, which is inefficient. Also, watch out for spaces before/after optional parts.
```suggestion
          - "(is|are) <name> (<currently>;{lock_states:state}[ in <area>])"
          - "(is|are) <name> {lock_states:state}[ in <area>]"
```

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
...
         response: one_yesno
         requires_context:
           domain: lock
         slots:
           domain: lock
 
       - sentences:
-          - "(is|are) [there] any door[s] {lock_states:state} [in <area>]"
-          - "(is|are) [there] any {lock_states:state} door[s] [in <area>]"
+          - "((is|are) [there]|do I have) any <lockable> ([<currently>];{lock_states:state} [in <area>])"
```
tetele, line 14: Same thing, you can't have a completely optional permutation item. Please refactor and keep in mind the whitespace in optional parts.

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
...
+          - "((is|are) [there]|do I have) any ([<currently>];{lock_states:state} <lockable> [in <area>])"
         response: any
         slots:
           domain: lock
 
       - sentences:
-          - "are all [the] door[s] {lock_states:state} [in <area>]"
+          - "are all <lockable> ([<currently>];{lock_states:state} [in <area>])"
```
tetele, line 21: Completely optional permutation item and misplaced whitespace.

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
...
         response: all
         slots:
           domain: lock
 
       - sentences:
-          - "(which|what) door[s] (is|are) {lock_states:state} [in <area>]"
+          - "(which|what) <lockable> (is|are) ([<currently>];{lock_states:state} [in <area>])"
```
tetele, line 27: Completely optional permutation item and misplaced whitespace.

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
...
         response: which
         slots:
           domain: lock
 
       - sentences:
-          - "how many door[s] (is|are) {lock_states:state} [in <area>]"
+          - "how many <lockable> (is|are) ([<currently>];{lock_states:state} [in <area>])"
```
tetele, line 33: Completely optional permutation item and misplaced whitespace.

sentences/en/lock_HassGetState.yaml
```diff
@@ -3,34 +3,42 @@ intents:
...
         response: how_many
         slots:
           domain: lock
+
+      - sentences:
+          - "tell me the [current] status of <name> [in <area>] [<currently>]"
```
tetele, line 39: Misplaced whitespace in optional components

sentences/en/lock_HassTurnOff.yaml
```diff
@@ -9,8 +9,8 @@ intents:
         response: lock
 
       - sentences:
-          - "unlock [all] [the] [(lock|door)[s]] [in] <area>"
-          - "unlock [all] <area> [(lock|door)[s]]"
+          - "unlock [all] <lockable> [in] <area>"
+          - "unlock [all] <area> [<lockable>]"
```
tetele, line 13: Misplaced whitespace in optional components

sentences/en/lock_HassTurnOn.yaml
```diff
@@ -9,8 +9,8 @@ intents:
         response: lock
 
       - sentences:
-          - "lock [all] [the] [(lock|door)[s]] [in] <area>"
-          - "lock [all] <area> [(lock|door)[s]]"
+          - "lock [all] <lockable> [in] <area>"
+          - "lock [all] <area> [<lockable>]"
```
tetele, line 13: Misplaced whitespace in optional components.

tetele (changes requested):
Please review my comment here https://github.com/home-assistant/intents/pull/2219#pullrequestreview-2118870868

sentences/en/lock_HassGetState.yaml
ViViDboarder: I'm not seeing any significant difference in performance when I run benchmarks over 324,000 sentences.

The performance seems to flip flop a bit between the two versions and, even taking worst case difference (0.5s for a benchmark iteration), that comes out to be 0.5, which is one micro second.

With that in consideration, it feels like optimizing for ease of maintenance is more important to giving users a good experience. My worry is not so much making this one change, but doing this for every intent type and then expanding for other common utterances.

Methodology:
```fish
# Number of generated sentences
$ python3 -m hassil.sample ./split_permutations.yaml --areas kitchen --names "front door" | jq -r .text | sort | uniq | wc -l
# Benchmarking with split sentences (your diff)
$ set PERMUTATIONS (python3 -m hassil.sample ./split_permutations.yaml --areas kitchen --names "front door" | jq -r .text)                                                                                                                             [.venv]
  time for i in (seq 1000)
      echo $PERMUTATIONS | hassil ./split_permutations.yaml --areas kitchen --names "front door" 2&> /dev/null
  end

________________________________________________________
Executed in   83.68 secs    fish           external
   usr time   56.92 secs  353.69 millis   56.56 secs
   sys time   14.81 secs  391.86 millis   14.42 secs
# Benchmarking with optional permutations
$ set PERMUTATIONS (python3 -m hassil.sample ./empty_permutations.yaml --areas kitchen --names "front door" | jq -r .text)                                                                                                                             [.venv]
  time for i in (seq 1000)
      echo $PERMUTATIONS | hassil ./empty_permutations.yaml --areas kitchen --names "front door" 2&> /dev/null
  end

________________________________________________________
Executed in   82.60 secs    fish           external
   usr time   56.22 secs  377.65 millis   55.85 secs
   sys time   14.59 secs  398.24 millis   14.20 secs
```

`empty_permutations.yaml`
```yaml
:hassil:Area names: ['kitchen']
language: en
intents:
  HassGetState:
    data:
      - sentences:
          - "(is|are) <name> ([<currently>];{lock_states:state} [in <area>])"
        response: one_yesno
expansion_rules:
  name: "[the] {name}"
  area: "[the] {area}"
  currently: "(currently|presently|right now|at the moment)"
lists:
  lock_states:
    values:
      - in: "[securely ]locked"
        out: "locked"
      - in: "unlocked"
        out: "unlocked"
```

`split_permutations.yaml`
```yaml
language: en
intents:
  HassGetState:
    data:
      - sentences:
          - "(is|are) <name> (<currently>;{lock_states:state} [in <area>])"
          - "(is|are) <name> {lock_states:state} [in <area>]"
        response: one_yesno
expansion_rules:
  name: "[the] {name}"
  area: "[the] {area}"
  currently: "(currently|presently|right now|at the moment)"
lists:
  lock_states:
    values:
      - in: "[securely ]locked"
        out: "locked"
      - in: "unlocked"
        out: "unlocked"
```

tetele: Well, that's interesting, i must admit. Simply looking at the code structure/complexity says that the split variant should be faster, but i guess that the complexity of other permutation items might have a role here as well.

Either way, I agree that for such minute performance differences, ease of maintenance should be prioritized.

@synesthesiam please take a look at the performance results above and let's discuss if we should change our approach regarding optimization enforcement.

ViViDboarder: Just wanted to see if any more thought has been given to this? I'd like to do another PR generalizing adding some more common values, then start to refactor sentences and generate new structures via an LLM.

It's pretty cool that we can now use an LLM in Assist, but I'm hoping to get things to feel more natural and flexible by letting the LLM generate sentences to match and then performing any actual sentence to intent matching offline.

tetele (approved):

PR #2220, opened 2024-06-06
[EN] Add lockable expansion

ViViDboarder:
Depends on #2219


tetele:
Why can't this be a part of #2219 ?

ViViDboarder:
I had split it to reduce the size of the PR. Happy to combine though!

ViViDboarder:
Combined with #2219

PR #2221, opened 2024-06-07
Make shade queries more flexible for more generic requests

ViViDboarder:
Add coverage for new sentences. These are ones that I try all the time naturally but end up failing.

### DIFF
--- a/sentences/en/cover_HassGetState.yaml
+++ b/sentences/en/cover_HassGetState.yaml
@@ -3,7 +3,7 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {cover_states:state} [in <area>]"
+          - "is <name> {cover_states:state} [[in ](<area>|<floor>)]"
         response: one_yesno
         requires_context:
           domain: cover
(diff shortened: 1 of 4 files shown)


sentences/en/cover_HassGetState.yaml
```diff
@@ -3,33 +3,37 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {cover_states:state} [in <area>]"
+          - "is <name> {cover_states:state} [[in ](<area>|<floor>)]"
```
tetele, line 6: I don't think "is the shutter closed in upstairs" sounds right.

The best idea would probably be to split this into 2 separate sentences instead of using the alternatives.

Also, here is probably a good place to replace `is` with `(is|are)` - e.g. `are the bedroom blinds open?`

tetele (changes requested):
The same issue of splitting the sentences must be fixed throughout this PR.

sentences/en/cover_HassGetState.yaml
ViViDboarder: That's true, but also unavoidable.

Doing a little brainstorming here:

Possible area names:
- Kitchen
- Upstairs
- Outside
- Basement

Possible floor names:
- Upstairs
- First Floor
- Basement

We'd want to match:
- "... in the kitchen" `in <area>`
- "... upstairs" -> `(<area>|<floor>)`
- "... outside" -> `(<area>|<floor>)`
- "... in the basement" `in (<area>|<floor>)`
- "... on the first floor" `on (<area>|<floor>)`

Since there is not really any strict alignment of article to `area` or `floor`, these can be generalized as:

`[(in|on) ](<area>|<floor>)`

Of course the whole thing should be optional as well. This is also very general and not specific to covers and something that I can, in another PR, add to `_common.yaml` so it's easier to add `floor` support to intents.

ViViDboarder: For now, I'm using `[[(in|on) ](<area>|<floor>)]` because it's more concise, even though it possibly matches less correct grammar. I'm happy to change it to `[([in ](<area>|<floor>)|[on ]<floor>)]`, especially by adding it to a common expansion.

tetele: I can think of stuff like `on the patio`.

I'd suggest creating a general expansion rule along the lines of `in_area_or_floor: ([(in|on) ]<area>|[on ]<floor>)`.

Then you can use `[<in_area_or_floor>]` throughout all sentences where it makes sense.

tetele (changes requested):

sentences/en/cover_HassGetState.yaml
ViViDboarder: Oh, I don't know how I missed this one... All the others were updated to cover that use case. I have a draft PR for more general expansion rules. I can add `[in_area_or_floor]` there. For now I'm using `[[<in>] <area_floor>]`, which expands to `[[(in|on|at|of)] (<area>|<floor>)]`. #2225

tetele (approved):

PR #2225, opened 2024-06-07
[EN] More common expansions

ViViDboarder:
Add some more common expansions to make generalizing sentences a bit easier. First implemented with covers but will expand out.

## Summary by CodeRabbit

- **New Features**
  - Enhanced phrase variations for controlling lights and querying states in a smart home environment for improved contextual understanding.
  - Introduced more concise and structured sentence templates for querying cover device states, increasing clarity and consistency.
  - Expanded the system's ability to interpret user queries with new and refined rules for better responsiveness.

- **Tests**
  - Added and refined test cases to cover new and varied expressions for querying the state of curtains and shades in a smart home setup.


tetele:
I need to look a bit deeper into this (particularly edge cases), but it does look like a welcome refactoring. Thanks!

sentences/en/_common.yaml
```diff
@@ -361,28 +361,36 @@ lists:
...
   brightness: "{brightness}[[ ]%| percent]"
-  light: "[the] (light|lights|lighting)"
+  light: "[the ](light|lights|lighting|lamp|lamps)"
   turn: "(turn|switch|change)"
   temp: "(temp|temperature)"
   temperature: "{temperature}[([ ]°[ ][{temperature_unit}])|( degrees[ {temperature_unit}])]"
-  open: "(open|raise)"
-  close: "(close|shut|lower)"
+  open: "(open|raise|lift)[ up]"
+  close: "(close|shut|lower)[ (up|down)]"
```
tetele, line 382: Can you provide a few examples where `up` makes sense?

sentences/en/_common.yaml
```diff
@@ -361,28 +361,36 @@ lists:
...
-  numeric_value_set: "(set|change|turn|increase|decrease|make)"
+  numeric_value_set: "(set|change|turn[ (up|down)]|increase|decrease|make)"
   in: "(in|on|at|of)"
   position: "{position}[[ ]%| percent]"
   volume: "{volume:volume_level}[[ ]%| percent]"
+  state: "[(present|current) ](state|status)"
 
   # Context awareness expansion rules
-  all: "(all|all of|every|every single|each|each and every)"
+  all: "[<are> ](all[[ of] the]|every[ single]|each[ and every])"
```
tetele, line 391: Why `[<are> ]`? Looking where this expansion rule was used prior to this PR, it makes no sense

sentences/en/_common.yaml
```diff
@@ -361,28 +361,36 @@ lists:
...
+  area: "[the ]{area}"
+  floor: "[the ]{floor}[ floor]"
+  area_floor: "(<area>|<floor>)"
+  in_area_floor: "[<in> ]<area_floor>"
+  what_is: "(what's|whats|what is|tell me)[ the]"
   where_is: "(where's|wheres|where is)"
+  which: "(which|what)[ of the]"
+  is: "(is|are)[ (there|the)]"
+  are: "<is>"
+  any: "[<are> ](any|some)[ of the]"
```
tetele, line 374: I'd like to extract `<are>` and leave `<any>` more generic

tetele (changes requested):

tetele:
Please also resolve conflicts

sentences/en/_common.yaml
ViViDboarder: "Ok Nabu, open up the kitchen shades"

Maybe this is a dialect thing, but throwing unnecessary prepositions is definitely something commonly done in the Chicago-land area where I grew up.

tetele: I meant for `close`

sentences/en/_common.yaml
ViViDboarder: This is for the `all` response type. The question being "Are all the shades open?" Or "Are all the shades open?". The sentences could use `[<are> ]<all>`, but I was hoping to make this easier from a maintainer perspective so that `all` responses are basically `<all>` and then the thing.

That said, the `all` responses could probably be refactored to be extremely generic since they can more or less be generalized as: `<are> <all> {{domain}_classes:device_class} <is> {{domain_states}:state}[ <in_area_floor>]`, if we had a way to expand to all domains or even just generate for all domains at build/deploy. That would make maintenance much simpler and negate the benefits of making it easier to maintain at a per domain level.

sentences/en/_common.yaml
ViViDboarder: Ah. That kind of depends on context. But ["Close up the house"](https://www.wordnik.com/words/close%20up) is something I've said/heard. I can't imagine someone using it with `lower`, but `close up` and `shut up` are certainly used.

sentences/en/_common.yaml
tetele: Don't get me wrong, it would be great if the expansion rule was called `<are_all>` or something, but this way it is just confusing. And `<all>` is used in sentences such as [this](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/en/light_HassTurnOff.yaml#L14).

sentences/en/_common.yaml
tetele: I won't argue with a native English speaker on this one, I'll just tale your word for it 😅

sentences/en/_common.yaml
ViViDboarder: Ah. That's a good point. `<are_all>` makes sense to me.

Have any idea where to start to implement the latter idea so we don't even have to do this for each domain? Is that feasible without an intent processing change?

tetele: IT is. Here's how I did it in Romanian:
1. create [an expansion rule](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/ro/_common.yaml#L628) that uses other undefined expansion rules
2. add the expansion rule as a sentence template in [multiple](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/ro/binary_sensor_HassGetState.yaml#L782) [sentences](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/ro/binary_sensor_HassGetState.yaml#L998) for [multiple](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/ro/binary_sensor_HassGetState.yaml#L1165) use cases
3. in each sentence, [populate](https://github.com/home-assistant/intents/blob/1242f1209b364af6a99179467ab24b488a0a8879/sentences/ro/binary_sensor_HassGetState.yaml#L786) the missing pieces using [local expansion rules](https://developers.home-assistant.io/docs/voice/intent-recognition/template-sentence-syntax#local-expansion-rules)

tetele (approved):

tetele:
Thanks a lot for this @ViViDboarder!

@TheFes @jlpouffier could you guys take a second look over this PR? It's been a long ping-pong and I don't want to be biased when I merge it.

ViViDboarder:
Any other feedback? I’m eager to cut some other PRs, but they are dependent on this.

sentences/en/cover_HassGetState.yaml
```diff
@@ -3,37 +3,45 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "is <name> {cover_states:state} [[(in|on) ](<area>|<floor>)]"
+          - "<is> <name> {cover_states:state} [<in_area_floor>]"
```
TheFes, line 6: This way only singular named entities work.
What if there is a group entity called `Bedroom Shutters`. In the previous version using `<is>` you could say `"Are the bedroom shutters closed`. With the fixed `is` this is no longer possible.

TheFes (approved):

tetele:
Sorry about the delay @ViViDboarder and thanks you again for the contribution!

PR #2230, opened 2024-06-13
[EN] Enhanced get state sentences for `media_player`

### LINKED ISSUE #2229: HassGetState and media players
I am trying to move from my rhasspy system to assist but the very first thing I tried to ask off assist makes me wonder if I actually should, it seems kinda strangely configured in places.

Scenario: 
- Behind me is a receiver, it is on and currently playing music

Question to assist:
- Is the receiver on?

Answer:
- No

This seemed strange to me so I started reading the inbuild sentences. Took me ages to figure out that the responses are in a different file to begin with. For sentences for assist to understand I could find quite a few examples of how to add a special case for some domain specific things, but I personally can't really see how I would do that for answers right now.

My assist is set to German, but from the responses it seems to me that English has mostly the same problem, though in the English response it at least returns the current state instead of just plain "No".

Since my receiver, and most likely all other media players in existence are obviously on while they are playing/idling I would like it if someone that knows how to navigate the response chaos could either work on this being answered better for media players in general, or point me in the right direction to solve it with custom_sentenses.

tetele:
An attempt to fix #2229 (at least in English)

These sentences "rephrase" some states that you ask about regarding media players.

Also, when the state is ambiguous (e.g. `is the TV on?`), the response clarifies the state, e.g. `the TV is on, but it is not playing`.

As a dependency in this PR, there is a [small fix](https://github.com/home-assistant/intents/pull/2230/commits/ce9ab25caa6d87ad2ca5240eaafe3e5a35d28a25) regarding passing a list for the `state` slot, which, although supported in hassil, would have appeared as no matches were found until now.

## Summary by CodeRabbit

- **New Features**
  - Introduced an `alternate_state` response template for more nuanced state representation.
  - Added detailed state management and querying for media players.

- **Enhancements**
  - Refined state definitions for media players, including states like "off", "on", "idle", "standby", "buffering", "playing", and "paused".
  - Updated Home Assistant intents to exclude `media_player` from specific state checks.

- **Tests**
  - Added tests for querying and managing media player states to ensure accurate and reliable responses.

### DIFF
--- a/responses/en/HassGetState.yaml
+++ b/responses/en/HassGetState.yaml
@@ -9,6 +9,14 @@ responses:
       one: |
         {{ slots.name | capitalize }} is {{ state.state_with_unit }}
 
+      alternate_state: |
+        {{ slots.name | capitalize }} is {{ slots.actual_state }}, but it is
+        {% if state.state in slots.alternate_state %}
+          {{ slots.alternate_state[state.state] }}
+        {% else %}
+          {{ state.state }}
+        {% endif %}
+
       one_yesno: |
         {% if query.matched %}
           Yes
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -128,6 +128,16 @@ lists:
         out: "locked"
       - in: "unlocked"
         out: "unlocked"
+  media_player_states:
+    values:
+      - in: "[turned ]off"
+        out: "off"
+      - in: "(idle|standing by)"
+        out: ["on", "idle", "standby", "buffering"]
+      - in: "(playing|[turned ]on)"
+        out: "playing"
+      - in: "paused"
+        out: "paused"
 
   # binary_sensor
   bs_battery_states:
@@ -372,6 +382,7 @@ expansion_rules:
   in: "(in|on|at|of)"
   position: "{position}[[ ]%| percent]"
   volume: "{volume:volume_level}[[ ]%| percent]"
+  media_players: "([media ]player[s]|speaker[s]stereo[s]|tv[s])"
 
   # Context awareness expansion rules
   all: "(all|all of|every|every single|each|each and every)"
--- a/sentences/en/media_player_HassGetState.yaml
+++ b/sentences/en/media_player_HassGetState.yaml
@@ -1,0 +1,75 @@
+language: en
+intents:
+  HassGetState:
+    data:
+      - sentences:
+          - is [the ][state of ]<name> [turned ]on[ in <area>]
+        slots:
+          actual_state: "on"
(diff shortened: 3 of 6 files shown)


sentences/en/media_player_HassGetState.yaml
```diff
@@ -0,0 +1,75 @@
...
+  HassGetState:
+    data:
+      - sentences:
+          - is [the ][state of ]<name> [turned ]on[ in <area>]
+        slots:
+          actual_state: "on"
+          alternate_state:
+            "on": "in an unknown state"
+            idle: "not playing"
+            standby: "in standby"
```
tetele, line 12: Maybe these should actually be `metadata`?

responses/en/HassGetState.yaml
```diff
@@ -9,6 +9,14 @@
       one: |
         {{ slots.name | capitalize }} is {{ state.state_with_unit }}
 
+      alternate_state: |
+        {{ slots.name | capitalize }} is {{ slots.actual_state }}, but it is
+        {% if state.state in slots.alternate_state %}
+          {{ slots.alternate_state[state.state] }}
+        {% else %}
+          {{ state.state }}
+        {% endif %}
```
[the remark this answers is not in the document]
tetele: That's a slot being passed from `sentences/en/media_player_HassGetState.yaml`, so first of all it's not an attribute and second of all it is not a problem.

sentences/en/_common.yaml
```diff
@@ -372,6 +382,7 @@ expansion_rules:
   in: "(in|on|at|of)"
   position: "{position}[[ ]%| percent]"
   volume: "{volume:volume_level}[[ ]%| percent]"
+  media_players: "([media ]player[s]|speaker[s]stereo[s]|tv[s])"
```
ViViDboarder, line 385: Missing a `|` between speaker and stereo. Should maybe add `television`? Also, this could be simplified by moving the `[s]` to the end.

```suggestion
  media_players: "([media ]player|speaker|stereo|tv|television)[s]"
```

ViViDboarder (changes requested):

sentences/en/_common.yaml
tetele: Agreed regarding `television`, but moving the `[s]` at the end makes it less readable and does not optimize matching.

synesthesiam:
Closing as stale — this PR has been inactive for a long time (over a year) and no longer applies cleanly against the current sentence structure. Thanks a lot for the contribution! If you'd like to revive it, please feel free to reopen or open a fresh PR rebased on the latest `main`.

PR #2250, opened 2024-06-24
update to language da

Dawn3888:
update to language da for light turn on and the test file. Also the common file has been updated.


MTrab (approved):
Seems good to me

PR #2286, opened 2024-07-08
[NL] Finetune cover_HassGetState

TheFes:
Changes:
* Add support for floors
* Split sentences so you can't say things like: `Hoeveel gordijnen in de woonkamer staan in de woonkamer open in de woonkamer`
* Only allow the `afdekking` expansion rule for the named sentences so for a cover named like `bloemetjes` something like `is bloemetjeslamp open` won't work
* Include spaces in the optional parts so there are no double spaces when the optional part isn't used

### DIFF
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -375,15 +375,12 @@ lists:
 
 expansion_rules:
   # generic expansion rules for sentences
-  name: "[de|het] {name}"
-  area: "[de|het] {area}"
+  name: "[de |het ]{name}"
+  area: "[de |het ]{area}"
+  floor: "[de |het ]{floor}"
   in: "[in|op|van|bij]"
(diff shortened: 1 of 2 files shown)


TheFes:
@frenck another one.
I'll just go to them one by one, and see what can be improved per file.

Thanks :)

frenck (changes requested):
There is a merge conflict @TheFes 🙈 

Could you take a look?

../Frenck

frenck:
Don't forget to mark the PR "Ready for review" when done; so it gets out of the draft state 😉 

../Frenck

TheFes:
Done, was doing other stuff while waiting for the tests to complete. Then I forgot to check back :)

frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #2287, opened 2024-07-08
[NL] Finetune cover_HassSetPosition

TheFes:
Changes:
 * Added sentences like `doe bloemetjesgordijn omhoog naar 50%`
 * Added sentences like `doe slaapkamergordijnen naar 25% omlaag`
 * make use of `numeric_value_set` expansion rule
 * add spaces inside optional parts instead of outside to avoid double spaces.
 * Add `stand` as variant for position
 * Allow `stand|postition` before the name (with optional `van`)
 * Add tests includling `stand|position`
 * Optionally add the `afdekking` expansion rule to names

It seems there is no support for floors for HassSetPosition. Is this intentional?

### DIFF
--- a/sentences/nl/cover_HassSetPosition.yaml
+++ b/sentences/nl/cover_HassSetPosition.yaml
@@ -3,17 +3,23 @@ intents:
   HassSetPosition:
     data:
       - sentences:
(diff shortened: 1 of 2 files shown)


TheFes:
@frenck and another one :)

PS No code rabbit?

PR #2290, opened 2024-07-08
[NL] Finetune HassTurnOff/On for covers and fans & align with EN

TheFes:
changes:
 * Add support for area aware intents
 * Added some additional combinations by allowing different word order
 * Some cleanup

Differences with EN
 * Add support for floors
 * The curtain intents are split off from the other device classes. If I look at my own situation, I have blinds for the sun, and curtains on the same window. I do not always want to close both, the blinds are closed on sunny days, while the curtains are closed after the sun has set.

## Summary by CodeRabbit

- **New Features**
  - Added new expansion rule for "floor" and included "hier" as an option in generic expansion rules.

- **Enhancements**
  - Refined sentence structures for the `HassTurnOff` and `HassTurnOn` intents to improve the handling and control of covers like garage doors, curtains, blinds, shutters, and shades.

- **Tests**
  - Streamlined and enhanced test cases to cover new sentence variations for turning off and on various home devices.

### DIFF
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -375,9 +375,9 @@ lists:
 
 expansion_rules:
   # generic expansion rules for sentences
-  name: "[de |het ]{name}"
-  area: "[de |het ]{area}"
-  floor: "[de |het ]{floor}"
+  name: "[de|het] {name}"
+  area: "[de|het] {area}"
+  floor: "[de|het] {floor}"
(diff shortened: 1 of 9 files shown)


TheFes:
@frenck 
next one, the split between curtain and the other device classes was already in place for these sentences, I kept it in for the reason mentioned above. Does it make sense to also apply that to other languages?

frenck (approved):

PR #2300, opened 2024-07-10
Add HassGetDate and HassGetTime

synesthesiam:
Add English sentences for two new intents (not yet supported in HA):

* `HassGetCurrentDate`
    * Receives a `date` slot value with a `datetime` object
    * Responds with the current date
* `HassGetCurrentTime`
    * Receives a `time` slot value with a `datetime` object
    * Responds with the current time

## Summary by CodeRabbit

- **New Features**
  - Introduced new intents `HassGetCurrentDate` and `HassGetCurrentTime` to retrieve the current date and time in English.
  - Added corresponding response templates for formatting and delivering date and time information.
  - Users can now query the current date and time through natural language commands.

- **Tests**
  - Added test cases to ensure accurate responses for the new date and time intents.

### DIFF
--- a/intents.yaml
+++ b/intents.yaml
@@ -180,12 +180,12 @@ HassSetPosition:
       - "domain"
       - "device_class"
 
-HassGetDate:
+HassGetCurrentDate:
   supported: true
   domain: homeassistant
   description: "Gets the current date"
 
-HassGetTime:
+HassGetCurrentTime:
   supported: true
   domain: homeassistant
   description: "Gets the current time"
--- a/intents.yaml
+++ b/intents.yaml
@@ -180,6 +180,16 @@ HassSetPosition:
       - "domain"
       - "device_class"
 
+HassGetDate:
+  supported: true
+  domain: homeassistant
+  description: "Gets the current date"
+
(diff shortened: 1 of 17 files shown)


intents.yaml
```diff
@@ -180,6 +180,16 @@ HassSetPosition:
...
 
+HassGetDate:
+  supported: true
+  domain: homeassistant
+  description: "Gets the current date"
+
+HassGetTime:
+  supported: true
+  domain: homeassistant
+  description: "Gets the current time"
```
frenck, line 191: Considering Home Assistant also has date & time entities, these names seem very unfortunate for the future.

Wouldn't it be better to use `HassGetCurrentDate` and `HassGetCurrentTime` instead?

frenck (changes requested):

intents.yaml
synesthesiam: That's a good point. I'll change them :+1:

synesthesiam: Updated

frenck (approved):
Nice! 🎉 

../Frenck

PR #2306, opened 2024-07-11
[FR] Time and Date intents

jlpouffier:
## Date
Added intent for getting the current date.
- Months are correctly translated
- I am using the ordinal form for the first day of the month (`Premier janvier` vs `2 janvier`)
- I decided to go with the response `Nous sommes le <date>`

## Time
Added intent for getting the current time.
- I kept the answer very simple: `Il est 12:45` 

Note: I tested the speech synthesis on both HA Cloud and Piper for that `XX:XX` format.
Both are generating understandable answers!

_It is still to be noted that HA Cloud provides better results for a few special cases_
Here are a few examples:
- The "hour sharp" case: `13:00`
  - HA Cloud: `Il est treize heures`
  - Piper: `Il est treize heures zéro zéro` 
- The "midnight" case: `00:15`
  - HA Cloud: `Il est minuit quize`
  - Piper: `Il est zéro heure quize`

## Others
I added one more "dirty" transcription that I keep hitting over and over again at home: 
`Ferme les volets` is almost always translated to `Faire les volets`... 
So now this form works too.

_I am so looking forward to a fuzzy matching logic at some point..._

### DIFF
--- a/responses/fr/HassGetCurrentDate.yaml
+++ b/responses/fr/HassGetCurrentDate.yaml
@@ -1,0 +1,27 @@
+language: fr
+responses:
+  intents:
+    HassGetCurrentDate:
+      default: >
+        {% set months = {
+           1: 'janvier',
+           2: 'février',
+           3: 'mars',
+           4: 'avril',
+           5: 'mai',
+           6: 'juin',
+           7: 'juillet',
+           8: 'août',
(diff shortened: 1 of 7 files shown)


piitaya (approved):
Seems to be correct 👍

Issue #2341, opened 2024-07-21
Add devenv support


nagyv:
Are the maintainers open to moving the `./scripts/setup` to [devenv](https://devenv.sh/getting-started/)?

While it's nice that the setup script creates a dedicated virtualenv, devenvs are a bit more powerful, and at the same time very easy to use for the "end user" / contributor.

frenck:
There is no need to create an issue for making a PR.

Let's discuss in the PR instead.

../Frenck

PR #2342, opened 2024-07-21
Added devenv support

nagyv:
Closes https://github.com/home-assistant/intents/issues/2341

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `.envrc` file to streamline the development environment setup using `direnv`.
  - Added a `devenv.nix` configuration file to manage development tools and scripts, enhancing workflow efficiency.
  - Created `devenv.yaml` to specify project dependencies and settings, improving the setup of development environments.

- **Chores**
  - Updated `.gitignore` to exclude unnecessary configuration and environment files, ensuring a cleaner repository.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+source_url "https://raw.githubusercontent.com/cachix/devenv/95f329d49a8a5289d31e0982652f7058a189bfca/direnvrc" "sha256-d+8cBpDfDBj41inrADaJt+bDWhOktwslgoP5YiGJ1v0="
+
+use devenv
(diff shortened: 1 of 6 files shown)


frenck:
Not sure why you created an issue instead of just providing context in the spar description?

This differs imho too much from what we use everywhere else in all our projects.

In the end, we steer towards using devcontainers handling all of this instead; so there is no need for this either imho.

Last but not least, adding another dependency, see unneeded.

nagyv:
I'm new to the HA intents project, and following the README did not mention devcontainers for development. As a result, the currently documented approach seems harder to use and is more prone to contaminate the global environment (virtualenvs are not perfect).

frenck:
Ok, let's document the devcontsiners and the enables support for codespaces here on GitHub in that case instead.

PR #2356, opened 2024-07-28
Tests: Filter response files for files ending with yaml

zyberzero:
When creating some translations I noticed that when I, in my editor of choice, had opened a response file the tests failed.
It turned out that due to my editor created a `.swp` file in the response folder the tests failed as the yaml parser doesn't (correctly) interpret the swap file as valid UTF-8 and throws an exception, for example this:
```python
_______________________________________________ ERROR at setup of test_climate_HassTurnOff[sv] _______________________________________________

language = 'sv'

    @pytest.fixture(scope="session")
    def language_responses(language: str) -> dict[str, Any]:
        """Load intent responses for a language."""
        merged_responses: dict = {}
        for response_file in (RESPONSES_DIR / language).iterdir():
>               merge_dict(merged_responses, yaml.safe_load(response_file.read_text()))

tests/conftest.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.10/pathlib.py:1135: in read_text
    return f.read()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <encodings.utf_8.IncrementalDecoder object at 0x7fc56b37fd60>
input = b'b0VIM 8.2\x00\x00\x00\x00\x10\x00\x00\x9f\xdb\xa4fR\x05I\x00?\xff\x01\x00zyber\x00\x00\x00\x00\x00\x00\x00\x00\x00\x...e }}"\x00      default: "Slog av {{ slots.name }}"\x00    HassTurnOff:\x00  intents:\x00responses:\x00language: sv\x00'
final = True

    def decode(self, input, final=False):
        # decode input (taking the buffer into account)
        data = self.buffer + input
>       (result, consumed) = self._buffer_decode(data, self.errors, final)
E       UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9f in position 16: invalid start byte

/usr/lib/python3.10/codecs.py:322: UnicodeDecodeError
```

This PR introduce a simple check that the extension of the file it tries to load is indeed a yaml-file (or at least, it checks that the file says it is a yaml-file).


balloob:
What editor do you use? I don't want to start adding workarounds for every editor under the sun. Aren't swap files short lived, and are the odds of this happening tiny?

tests/conftest.py
```diff
@@ -73,5 +73,6 @@ def language_responses(language: str) -> dict[str, Any]:
     """Load intent responses for a language."""
     merged_responses: dict = {}
     for response_file in (RESPONSES_DIR / language).iterdir():
```
synesthesiam, line 75: Instead of iterdir, let's use glob

synesthesiam (changes requested):

zyberzero:
> What editor do you use? I don't want to start adding workarounds for every editor under the sun. Aren't swap files short lived, and are the odds of this happening tiny?

In this case, it's vim. However, I don't see it as a workaround for one specific editor - the proposed change filters out only yaml-files to be parsed rather then adding exceptions specific for an editor.

With the current state, adding for example an .md-file in the directory (for example, a file that describes the tests) would make the tests fail.

PR #2357, opened 2024-07-28
[sv] Add timer intents

zyberzero:
Adds sentences, tests and responses for timer intents.

## Summary by CodeRabbit

- **New Features**
	- Added localized responses for various timer-related intents in Swedish, including starting, pausing, unpausing, increasing, decreasing, and canceling timers.
	- Enhanced user interaction for Swedish-speaking users with clear and contextually appropriate messages.

- **Tests**
	- Introduced comprehensive test cases for new Swedish language intents to ensure accurate handling of timer commands and interactions.
	- Added a section for timer objects in test fixtures to facilitate testing of multiple timers with unique settings.

- **Documentation**
	- Updated sentence templates and error messages for improved clarity in user feedback regarding timer functionalities.


responses/sv/HassTimerStatus.yaml
```diff
@@ -0,0 +1,94 @@
...
+          {% endif %}
+        {% endif %}
+
+        {% if next_timer: %}
+          {# At least one active timer #}
+          {% if (next_timer.rounded_hours_left == 1) and (next_timer.rounded_minutes_left > 1): %}
+            1 timme och {{ next_timer.rounded_minutes_left }} minuter
+          {% elif (next_timer.rounded_hours_left ==1) and (next_timer.rounded_minutes_left == 1): %} 
+            1 timme och 1 minut
+          {% elif (next_timer.rounded_hours_left > 0): %}
```
larsdunemark, line 48: This will catch next_timmer.rounded_hours_left == 2 and say "1 timme" 

Since it match it will not go to the next else-if to see if the hours is more the 1 hour.

Please add a test case for this case also.

larsdunemark (changes requested):

responses/sv/HassTimerStatus.yaml
zyberzero: You're right, I'll fix it. Is there a good way to add fixtures for specific tests? Because if I add a new timer to the fixtures file it affects some other tests (du har ~~två~~ tre pausade timers)

larsdunemark: Not sure, will check with the other language_leaders to see if there is any input on more dynamic test cases. It should have a testcase where no times exists also so we need someway to define it in the testcase.

zyberzero: I've fixed it and added a test case for it. Should I open this for review in the meantime until we figure out how we can do more specific tests?

larsdunemark: Sure, we can fix and add more testcase if we get a better solution.

larsdunemark (approved):

PR #2372, opened 2024-08-09
[EN] Add "activate" phrasing for scripts

tannisroot:
Only feels natural for the user to say that they want a certain script activated. And the light domain uses "activate" too.
Real life use-case: i have a script named "Intercom" that opens intercom door and it feels most natural to me to say "activate" instead of "start" or "run" or "turn on".

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `HassTurnOn` intent to include "activate" as an additional command option for more flexible script activation.


tetele:
Sorry for the delay, I was on vacation. Could you please also add a test sentence to https://github.com/home-assistant/intents/blob/main/tests/en/script_HassTurnOn.yaml

ViViDboarder (approved):
Looks good. Thanks!

tetele (approved):

PR #2395, opened 2024-09-25
[SV] Locks sentences

frli4797:
Basic sentences for locks in Swedish.

## Summary by CodeRabbit

- **New Features**
	- Added Swedish language responses for locking ("Låste") and unlocking ("Låste upp") actions.
	- Introduced intents for controlling locks, allowing users to lock/unlock specific doors or all doors in an area.
	- Expanded recognition of lock states and related terms in Swedish, enhancing user interaction.

- **Tests**
	- Added test cases to validate the new locking and unlocking intents in Swedish.


sentences/sv/lock_HassTurnOff.yaml
```diff
@@ -0,0 +1,17 @@
...
+        requires_context:
+          domain: lock
+        response: lock
+
+      - sentences:
+          - "lås upp [<alla>] <låsbar> <i_på> <area>"
+          - "lås upp [<alla>] <area> <låsbar>"
+        slots:
+          domain: "lock"
+          name: "alla"
```
larsdunemark, line 16: I would expect that this sentence should set the name to "all" and not "alla". Since it is not translated slot name.
Looking at DE and ES languages for example.

larsdunemark (changes requested):
Need to update the all slot name.

sentences/sv/lock_HassTurnOff.yaml
frli4797: Oh. My bad. Fixed in latest commit.

larsdunemark (approved):
LGTM

PR #2399, opened 2024-09-26
[FR&EN] Floor Support for HassLightSet

jlpouffier:
## Context
We noticed that floors were supported for HassTurnOn and HassTurnOff but not for HAssSetLight.
This meant that light support was uneven 
- `Turn on the lights on the first floor` was working ✅
- `Set the lights of the first floor to 50%` was not ❌

 ## Scope
This PR fixes this.
We can now control brightness and color on a floor level (in addition to the area level that was already supported)

## Why English and French in the same PR?
As it is a brand new set of sentences, we also decided to provide a draft of the English sentences too.
~~They are derived from the area sentences.~~
@tetele rightfully pointed out that the `area` sentences way not be in the best of shape.
So I decided to add only a single sentence in English.
Up to English LL to take it from here.

## Anything else
Yes. We noticed a few `name: all` when targeting complete area in English.
This is not required anymore: When targeting n area (or floor), all entities of the area (or floor) are by default targeted.
Thus, we simplified this particular file.
_(Note: More `name: all` could exist in the English set, we only touched the file in scope of this PR)_

## Summary by CodeRabbit

- **New Features**
  - Introduced a new optional field for specifying the floor in light settings.
  - Enhanced brightness and color control commands to allow adjustments for specific floors.

- **Bug Fixes**
  - Streamlined syntax by removing the "all" slot from brightness commands.

- **Tests**
  - Added new test cases for brightness and color control on specific floors in both English and French.

### DIFF
--- a/sentences/en/light_HassLightSet.yaml
+++ b/sentences/en/light_HassLightSet.yaml
@@ -67,12 +67,7 @@ intents:
 
       # Floor support for brightness
       - sentences:
-          - "[<numeric_value_set>] [the] brightness in <floor> to <brightness>"
-          - "[<numeric_value_set>] [the] brightness of <floor> to <brightness>"
           - "[<numeric_value_set>] <floor> brightness [to] <brightness>"
-          - "[<numeric_value_set>] <floor> [to] <brightness> brightness"
-          - "[<numeric_value_set>] <floor> [to] <brightness>"
-          - "[<numeric_value_set>] [all] [the] lights [<in>] <floor> to <brightness> [brightness]"
         response: "brightness"
 
       - sentences:
@@ -106,6 +101,5 @@ intents:
 
       # Floor support for color
       - sentences:
-          - "[<set>] [[the] color of] (<floor> | [all [the]] lights in <floor> | [all] <floor> lights) [to] {color}"
-          - "[<set>] (<floor> | [all] lights in <floor> | [all] <floor> lights) [color] [to] {color}"
+          - "[<set>] <floor> [color] [to] {color}"
         response: "color"
--- a/sentences/en/light_HassLightSet.yaml
+++ b/sentences/en/light_HassLightSet.yaml
@@ -23,14 +23,10 @@ intents:
           - "[<numeric_value_set>] <area> [to] <brightness> brightness"
           - "[<numeric_value_set>] <area> [to] <brightness>"
           - "[<numeric_value_set>] [all] [the] lights [<in>] <area> to <brightness> [brightness]"
-        slots:
-          name: "all"
         response: "brightness"
 
       - sentences:
           - "<numeric_value_set> <area> to <brightness>"
-        slots:
-          name: "all"
         response: "brightness"
 
       - sentences:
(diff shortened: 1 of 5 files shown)


sentences/en/light_HassLightSet.yaml
```diff
@@ -71,6 +65,23 @@ intents:
           area:
             slot: true
 
+      # Floor support for brightness
+      - sentences:
+          - "[<numeric_value_set>] [the] brightness in <floor> to <brightness>"
+          - "[<numeric_value_set>] [the] brightness of <floor> to <brightness>"
```
tetele, line 71: Use `<in>` which is defined in `_common.yaml`.

tetele: ...and make it optional (e.g. `set the brightness upstairs to 25%`)

sentences/en/light_HassLightSet.yaml
```diff
@@ -71,6 +65,23 @@ intents:
...
+          - "[<numeric_value_set>] <floor> brightness [to] <brightness>"
+          - "[<numeric_value_set>] <floor> [to] <brightness> brightness"
+          - "[<numeric_value_set>] <floor> [to] <brightness>"
```
tetele, line 74: This will surely clash with other sentences in the future. Or with volume at the moment. We should specifically include `brightness` in the sentence.

sentences/en/light_HassLightSet.yaml
jlpouffier: It's a copy-paste of the current sentences in English for areas.
But we can work a bit but we're not maintainer of that language

sentences/en/light_HassLightSet.yaml
jlpouffier: It's already the case for area too.
The sentence exists without `brightness`

sentences/en/light_HassLightSet.yaml
tetele: Understood. I'm not sure how to go about this, as it appears we are digging ourselves in.

For example in `core`, even if the code coverage is bad, you can't make it worse in a PR. I was hoping not to copy whatever was present for `{area}` as that's not necessarily a benchmark.

Leave this PR as you see fit and I'll try to update with the changes I myself proposed.

sentences/en/light_HassLightSet.yaml
tetele: Same thing, it was made that way when there was nothing other than brightness :) That doesn't mean it's still OK, but leave it as you see fit and I'll try to update when I can (for both floors and areas).

sentences/en/light_HassLightSet.yaml
jlpouffier: Non no it's up to you to decide. You're the English LL 😉
We just felt bad not adding "something" for English. 

I have an idea, how about we add just ONE sentence for floor support just to state the fact it is supported in English. 

Would that be ok? 
I can remove the copy paste and provide a single sentence.

sentences/en/light_HassLightSet.yaml
jlpouffier: I am simplifying sentences. The main idea was to make sure something was present in the english language, but you are correct, it's technically not my job to do so.
So I will only put a single sentence.

sentences/en/light_HassLightSet.yaml
jlpouffier: Edit: Done. Will ask you to re-review in a minute

tetele (approved):
Thanks!

sentences/en/light_HassLightSet.yaml
tetele: i.e. I'm trying to block as little creative effort as possible and at the same time to create as few future problems as possible.

PR #2407, opened 2024-10-02
[PL] - added area for running and finishing vacuum cleaning

witold-gren:
## Summary by CodeRabbit

- **New Features**
	- Enhanced Polish language support with updated error messages and expanded sentence patterns for vacuum commands.
	- Added flexibility in command structures for returning and starting the vacuum, allowing for area specifications.
	- Improved naming conventions for areas and entities in the home automation system for better representation.

- **Bug Fixes**
	- Resolved inconsistencies in error messages and command interpretations.

- **Tests**
	- Introduced new test cases for vacuum commands to validate expanded functionality and area specifications.

### DIFF
--- a/tests/pl/_fixtures.yaml
+++ b/tests/pl/_fixtures.yaml
@@ -44,7 +44,7 @@ areas:
     id: entrance_id
     floor: "Parter"
 
-  - name: "Korytarz[u]|Hol[u]"
+  - name: "Korytarz[u|a]|Hol[u]"
     id: hallway_id
     floor: "Parter"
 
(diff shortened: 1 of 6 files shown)


tests/pl/_fixtures.yaml
```diff
@@ -495,118 +495,118 @@ entities:
...
 
-  - name: "Czujnik powietrza w kuchni"
+  - name: "Czujnik[u|a] powietrza w kuchni"
     id: "sensor.kitchen_aqi_sensor"
     state: "50"
     attributes:
       device_class: aqi
 
-  - name: "Czujnik ciśnienia"
+  - name: "Czujnik[u] ciśnienia zewnętrznego"
```
Uriziel01, line 511: Shouldn't we include the `|a`? Seems identical to one above it. Can be usefull fome some tests.

tests/pl/_fixtures.yaml
```diff
@@ -495,118 +495,118 @@ entities:
...
-  - name: "Dwutlenek węgla"
+  - name: "Pomp(a|y|ę|ie) ciepła" # Dwutlenek węgla
     id: "sensor.heat_pump_co2"
     state: "50"
     attributes:
       device_class: carbon_dioxide
       unit_of_measurement: "ppm"
 
-  - name: "Tlenek węgla"
+  - name: "Pomp(a|y|ę|ie) ciepła" # Tlenek węgla
```
Uriziel01, line 532: Isn't this a problem that we now have two identical ones here? If we ask for it which one should be returned?

Uriziel01 (changes requested):

tests/pl/_fixtures.yaml
witold-gren: This is not necessary now. I start working with sensor.. if the new name will be needed then I will add it

tests/pl/_fixtures.yaml
witold-gren: This is not a problem, because we will check specific `device_class` for this name. In one case we use `carbon_dioxide` but in the other `carbon_monoxide`.

tests/pl/_fixtures.yaml
witold-gren: This name is just for running the test, and we don't have any tests in this PR at the moment 😀

Uriziel01 (approved):

PR #2418, opened 2024-10-16
[EN] Explicit and friendly domain name for HassTurnOn/HassTurnOff

tetele:
"Turn on the TV" would hare produced the response "Turned on the media_player", read as "Turned on media underscore player"

We've patched this by adding a map of how each domain should be handled in the response text.

Other approaches would have been to:
- use the `name` slot instead of domain (such as in RO)
- add the domain without underscore in the response for the `media_player` domain, which is what I wanted to do with `metadata`
- programatically convert underscores to spaces in all domains - it could encounter edge cases

## Summary by CodeRabbit

- **New Features**
	- Enhanced responses for the `HassTurnOn` and `HassTurnOff` intents with dynamic feedback based on device types.
	- Introduced a translation system for clearer and more specific messages related to various devices.
  
These updates improve user interaction by providing contextually relevant responses when controlling devices.


tetele:
@coderabbitai

> Add a fallback for unknown domains to prevent potential errors. For example:

The point of these files are to be easily interpreted and translated by the language leaders, so adding extra code complexity will harm legibility. A response such as "Turned on" is perfectly fine (i.e. without domain name) in case an unknown domain happens to be assigned to an entity.

jlpouffier (approved):
Thx very much! This will greatly help other language leaders

PR #2439, opened 2024-10-31
[en] Add more expansions for state values

ViViDboarder:
Found a bunch of state expansions that have terms that are used interchangeably. Expanding them here to improve matching.

I'm working on another PR to expand test sentences and then work towards closing gaps.

## Summary by CodeRabbit

- **New Features**
	- Enhanced input handling for device states, allowing synonyms and variations.
	- Expanded recognition for terms in `cover_states`, `lock_states`, and binary sensor states, including "closed" and "shut," as well as variations for "locked" and "unlocked."
	- Broadened input options for `bs_presence_states`, now accepting "away," "not present," or "gone."
	- Added new test cases for various binary sensors, improving the coverage of the `HassGetState` intent.
	- Introduced a new sentence variation for curtain state inquiries, enhancing natural language processing capabilities.

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -86,7 +86,7 @@ lists:
         out: "off"
   on_off_domains:
     values:
-      - in: (light|lamp)[s]
+      - in: light[s]
         out: light
       - in: fan[s]
         out: fan
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -86,7 +86,7 @@ lists:
         out: "off"
   on_off_domains:
     values:
-      - in: light[s]
+      - in: (light|lamp)[s]
         out: light
       - in: fan[s]
         out: fan
@@ -96,11 +96,11 @@ lists:
     values:
       - in: "open"
         out: "open"
-      - in: "closed"
+      - in: "(closed|shut)"
         out: "closed"
       - in: "opening"
         out: "opening"
-      - in: "closing"
+      - in: "(closing|shutting)"
         out: "closing"
   cover_classes:
     values:
@@ -134,7 +134,7 @@ lists:
     values:
       - in: "low"
         out: "on"
-      - in: "normal"
+      - in: "(normal|charged)"
         out: "off"
 
   bs_battery_charging_states:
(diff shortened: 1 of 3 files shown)


mrdarrengriffin (approved):
I like this addition. I wonder if we should also add "full" for batteries also 🤔

sentences/en/_common.yaml
```diff
@@ -86,7 +86,7 @@ lists:
         out: "off"
   on_off_domains:
     values:
-      - in: light[s]
+      - in: (light|lamp)[s]
```
tetele, line 89: The trouble with this is that when you say "turn off the bedroom lamps" without having a lamp, it will turn off LED strips or other stuff from the `light` domain, which is why I've refrained for adding anything other than `"light"` so far.

I'm not saying it's wrong to add it, but that's the end result. Do you think that's acceptable? cc @mrdarrengriffin

tetele (changes requested):
Overall excellent change, with the asterisk on lamps.

However, it would be good to add/edit some tests to use these new state labels.

sentences/en/_common.yaml
ViViDboarder: Hmm. At first I was thinking that it world do no harm, the situation you describe I think is harmless because you have no other device you could be referring to. However, if there was a lamp or two in the room and instead it turned off everything, that could be unexpected. 

I’ll take that out.

tetele:
> However, it would be good to add/edit some tests to use these new state labels. ([source](https://github.com/home-assistant/intents/pull/2439#pullrequestreview-2409648983))

:D

mrdarrengriffin:
> @mrdarrengriffin so the reason I didn't add full is because this was for the binary sensor which I believe typically only states "low" or "not low". So it may not be "full" by what I think someone might expect.

Ah yeah, good point.

ViViDboarder:
> > However, it would be good to add/edit some tests to use these new state labels. ([source](https://github.com/home-assistant/intents/pull/2439#pullrequestreview-2409648983))
> 
> :D

Yea, I can throw some simple ones in. I didn't do that originally because I plan to create a bunch more tests with LLM generated sentences.

ViViDboarder:
Added tests and found a bug with how states for window were mapped. Fixed that too.

tetele (approved):

PR #2448, opened 2024-11-03
Add HassRespond intent

synesthesiam:
Adds a new intent named `HassRespond` that simply returns the response and takes no action.
This intent is meant to hard-code responses to specific questions.

Core support PR: https://github.com/home-assistant/core/pull/129755

## Release Notes

- **New Features**
  - Introduced a new intent, `HassRespond`, allowing the Home Assistant to return responses without performing actions.
  - Added predefined responses for the `HassRespond` intent, enhancing user interaction with greetings and data handling information.
  - Implemented sentence patterns to recognize various user inquiries related to the assistant's capabilities and origin.

- **Tests**
  - Added test cases to ensure the `HassRespond` intent functions correctly, validating responses to user queries.

### DIFF
--- a/responses/en/HassRespond.yaml
+++ b/responses/en/HassRespond.yaml
@@ -2,7 +2,7 @@ language: en
 responses:
   intents:
     HassRespond:
-      default: "Hello from Home Assistant."
+      hello: "Hello from Home Assistant."
       listening: "No, I only record when you speak the wake word."
       data: "Your data is sent to your Home Assistant server."
       commands: "To learn what you can ask, visit H.A. dot I.O. slash voice."
--- a/responses/en/HassRespond.yaml
+++ b/responses/en/HassRespond.yaml
@@ -1,7 +1,7 @@
 language: en
 responses:
   intents:
-    HassHello:
+    HassRespond:
       default: "Hello from Home Assistant."
       listening: "No, I only record when you speak the wake word."
       data: "Your data is sent to your Home Assistant server."
--- a/sentences/en/homeassistant_HassRespond.yaml
+++ b/sentences/en/homeassistant_HassRespond.yaml
@@ -4,7 +4,7 @@ intents:
     data:
       - sentences:
           - "(hello|hi) [home assistant]"
-        response: default
+        response: hello
 
       - sentences:
           - "are you always (listening|recording)"
--- a/sentences/en/homeassistant_HassRespond.yaml
+++ b/sentences/en/homeassistant_HassRespond.yaml
@@ -1,6 +1,6 @@
 language: "en"
 intents:
-  HassHello:
+  HassRespond:
     data:
       - sentences:
           - "(hello|hi) home assistant"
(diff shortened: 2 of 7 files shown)


responses/en/HassRespond.yaml
```diff
@@ -0,0 +1,9 @@
+language: en
+responses:
+  intents:
+    HassRespond:
+      default: "Hello from Home Assistant."
```
balloob, line 5:
```suggestion
      default: "How can I help?"
```

sentences/en/homeassistant_HassRespond.yaml
```diff
@@ -0,0 +1,23 @@
+language: "en"
+intents:
+  HassRespond:
+    data:
+      - sentences:
+          - "(hello|hi) home assistant"
```
balloob, line 6:
```suggestion
          - "(hello|hi)[ home assistant]"
```

responses/en/HassRespond.yaml
balloob: Can we call this something else than default? Or do we always need a default?

balloob (approved):
This is good. I would suggest we also explore in a future PR adding a lint rule to ensure that each sentence block has a mandatory response key defined.

sentences/en/homeassistant_HassRespond.yaml
synesthesiam: "hello" and "hi" by themselves are too short to get recognized until we fix the VAD issue :slightly_frowning_face:

responses/en/HassRespond.yaml
synesthesiam: Maybe say something about needing the wake word, since we don't support conversations yet?

sentences/en/homeassistant_HassRespond.yaml
mrdarrengriffin: Just updated it from `(hello|hi)[ home assistant]` to `(hello|hi) [home assistant]`

mrdarrengriffin: Didn't see Mike's comment but we defo need to get standalone "Hello" or "Hi" commands before production release. Raised on Discord

responses/en/HassRespond.yaml
balloob: Nah, that will be a good time to learn it, and a good reason for us to add conversation support.

TheFes:
Dutch translations are pending in #2449. Will keep an eye on this PR for changes to the EN sentences.

witold-gren:
Polish translations are also pending https://github.com/home-assistant/intents/pull/2450

synesthesiam:
Renamed the `default` response key to `hello`, and then merge :slightly_smiling_face:

tetele:
Note: I am doubtful STTs will be able to transcribe "hello home assistant" most languages other than English

PR #2455, opened 2024-11-04
[en] Expand light_HassTurnOn and light_HassTurnOff with natural sentences

ViViDboarder:
Used an LLM to generate a whole bunch of natural sounding sentences for the intent, cleaned them up by removing some that sounded not so natural, and then updated the sentences to get tests to pass. It also ended up suggesting some new expansions synonym expansions, so those are added to `_common`.


sentences/en/_common.yaml
```diff
@@ -376,15 +376,15 @@ expansion_rules:
...
+  light: "[the ](light|lights|lighting|lamp|lamps|bulb|bulbs|light fixture|light fixtures|light switch|light switches)"
+  turn: "(turn|switch|change|bring)"
   temp: "(temp|temperature)"
   temperature: "{temperature}[([ ]°[ ][{temperature_unit}])|( degrees[ {temperature_unit}])]"
   open: "(open|raise|lift)[ up]"
   close: "(close|shut|lower)[ (up|down)]"
   set: "(set|make|change|turn)"
   numeric_value_set: "(set|change|turn[ (up|down)]|increase|decrease|make)"
-  in: "(in|on|at|of)"
+  in: "(in|on|at|of|across|around|throughout)"
```
tetele, line 387: Is there any way "around" could mean "the areas around the one i'm mentioning"? If so, let's drop it.

sentences/en/_common.yaml
```diff
@@ -376,15 +376,15 @@ expansion_rules:
   are_any: "[<are> ]<any>"
   how_many: "how many[ of the]"
   brightness: "{brightness}[[ ]%| percent]"
-  light: "[the ](light|lights|lighting|lamp|lamps)"
-  turn: "(turn|switch|change)"
+  light: "[the ](light|lights|lighting|lamp|lamps|bulb|bulbs|light fixture|light fixtures|light switch|light switches)"
```
tetele, line 379: I said this before: bulbs, LEDs, chandeliers etc. are not good words to include because it would be weird if a user would say "turn on the bulb" in an area with only LEDs and no light bulbs (or one light bulb and many other lights), then all the lights would unexpectedly turn on. It would work anyway if the user would add the name/alias "bulb" to what they know is an actual light bulb.

Also, if we really want to include it, at least let's make it `[light] bulb`. I really, really don't want to include it, though.

sentences/en/_common.yaml
```diff
@@ -394,8 +394,8 @@ expansion_rules:
   all: "(all[[ of] the]|every[ single]|each[ and every])"
   are_all: "[<are> ]<all>"
   home: "(home|house|appartment|flat)"
-  everywhere: "(everywhere|in (all|each)[[ of] the] room[s]|in every room[s]|in the <home>)"
-  here: "(here|in here|in this room|in the room)"
+  everywhere: "(everywhere|all over|[<in> ](all|each)[[ of] the] room[s]|[<in> ]every room[s]|(in|across|around|throughout) the [(entire|whole) ]<home>|<in> <all> area[s])"
```
tetele, line 397:
```suggestion
  everywhere: "(everywhere|all over|[<in> ](all|each)[[ of] the] room[s]|[<in> ]every room[s]|(in|across|around|throughout|all over) the [(entire|whole) ]<home>|<in> <all> area[s])"
```

sentences/en/_common.yaml
```diff
@@ -394,8 +394,8 @@ expansion_rules:
...
+  here: "(here|[in ]here|[in ](this|the) (room|area|space))"
```
tetele, line 398:
```suggestion
  here: "([in] here|[in] (this|the) (room|area|space))"
```
`(here|[in ]here)` is redundant and we decided to go withe the easily legible stype of `[optional] part` instead of the more optimized `[optional ]part`.

sentences/en/_common.yaml
```diff
@@ -416,4 +416,7 @@ expansion_rules:
 skip_words:
   - "please"
   - "can you"
+  - "could you"
   - "would you"
+  - "for me"
+  - "i'd like"
```
tetele, line 422:
```suggestion
  - "i'd like"
  - "i'd like to"
```

sentences/en/light_HassTurnOn.yaml
```diff
@@ -1,72 +1,83 @@
...
+          - "<turn> on <name> <light> [<in_area_floor>]"
+          - "[<turn>] <area_floor> <name> <light> [to] on"
+          - "[<turn>] <name> <light> [<in_area_floor>] [to] on"
+          - "activate <area_floor> <name> <light>"
+          - "activate <name> <light> [<in_area_floor>]"
+          - "light up <area_floor> <name> <light>"
+          - "light up <name> <light> [<in_area_floor>]"
         requires_context:
           domain: "light"
+
```
tetele, line 17: Comments for each section would be a welcome addition

tetele (changes requested):
I agree with most proposals. The stuff I don't agree with are in comments below.

Two other things:
1. This will have to be ported for `light_HassTurnOff` as well
2. Let's settle on a style for optional parts of sentences: `[optional] part` or `[optional ]part` and use that everywhere.

sentences/en/_common.yaml
mrdarrengriffin: Yeah, I agree

sentences/en/_common.yaml
mrdarrengriffin: Yeah, I would imagine that if I said "around the living room", it would affect rooms next to the living room, but not the living room itself.

ViViDboarder: That interpretation doesn't really resonate with me. Like, If I say "turn on the lights all around my house", I'm generally referring to inside, not perimeter. More like, within the perimeter.

Similar to how someone saying "Let's go around the room" typically refers to within the room, but inclusive of everyone, rather than outside the perimeter.

sentences/en/_common.yaml
ViViDboarder: We discussed this on Discord, so yea. I'll drop this one. I think it makes sense to remove given that bulb, fixture, and switch are too specific and not used generally to refer to "lights".

sentences/en/_common.yaml
tetele: I get your point. The questions is: is "turn on the lights around the bedroom" (1) that common in terms of usage of "around" with any meaning and (2) widespread enough in terms of "around" meaning "inside the room" to mandate the change?

ViViDboarder: Well, I think so. :) For what it's worth, if I ask Siri to "Turn the lights on around the kitchen", all lights in my kitchen go on. Not that it means we _must always_ do what Siri or others do, but it does at least indicate I'm not just an outlier.

tetele: Alrighty, then. Rebase, fix the tests and let's get it merged

tetele (approved):

mrdarrengriffin (approved):

PR #2449, opened 2024-11-04
Add NL HassRespond Translations

TheFes:
Linked to PR https://github.com/home-assistant/intents/pull/2448

Dutch translations for HassRespond

## Summary by CodeRabbit

- **New Features**
	- Introduced a new intent, `HassRespond`, for the Home Assistant, allowing it to provide responses without taking action.
	- Enhanced the `HassGetState` intent by adding a new slot for specifying the state to match.
	- Added predefined responses for the `HassRespond` intent in both English and Dutch, covering greetings, data handling, and community acknowledgment.

- **Tests**
	- Implemented test cases for the `HassRespond` intent to ensure accurate responses in both English and Dutch.

PR #2457, opened 2024-11-05
[NL] New attempt for HassRespond translations

TheFes:
GitHub did some strange stuff with the first attempt (#2449) so here's another go.

Provides Dutch translations for #2448


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #2463, opened 2024-11-06
[DE] Add "Beschattung[en]" to cover_classes and "geöffnet" to cover_s…

ascha191:
[DE] Add "Beschattung[en]" to cover_classes and "geöffnet" to cover_states

PR #2464, opened 2024-11-06
test sentences for "Beschattung" in cover_HassTurnOff.yaml

ascha191:


mib1185:
please add these test sentences in #2463

PR #2465, opened 2024-11-06
[DE] Test sentences for "Beschattung" in cover_HassTurnOn.yaml

ascha191:


mib1185:
please add these test sentences in #2463

PR #2468, opened 2024-11-06
[ES] Add more sentences to test getting the temperature

cibernox:
Particularly I added masculine examples and examples that used "de la / del" instead of "en la / en el", which is also a valid way command that wasn't tested.

### DIFF
--- a/shared/__init__.py
+++ b/shared/__init__.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import partial
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, Optional, List, Set, Tuple
 
 from hassil import parse_sentence
 from hassil.intents import SlotList, TextSlotList, is_template
--- a/shared/__init__.py
+++ b/shared/__init__.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import partial
-from typing import Any, Dict, Optional, List, Set, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 
 from hassil import parse_sentence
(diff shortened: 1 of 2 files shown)


cibernox (commented):
Somewhat unrelated, but I figured out why `¿Cual es la temperatura del salón?` didn't work for me. It was because my living room didn't have a thermostant entity on it, it had a temperature sensor instead.

Turns out temperature sensors are not valid targets for temperature queries, and I don't think that's right. It doesn't seem exclusive to spansh language. I have temperature sensors in my pool and in my greenhouse, none if which will ever have a thermostat. That seems like it should change.

davefx:
> Somewhat unrelated, but I figured out why `¿Cual es la temperatura del salón?` didn't work for me. It was because my living room didn't have a thermostant entity on it, it had a temperature sensor instead.
> 
> Turns out temperature sensors are not valid targets for temperature queries, and I don't think that's right. It doesn't seem exclusive to spansh language. I have temperature sensors in my pool and in my greenhouse, none if which will ever have a thermostat. That seems like it should change.

About this, I would open a different issue, as it's unrelated to the language.

shared/__init__.py
```diff
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import partial
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, Optional, List, Set, Tuple
```
cibernox, line 4: FWIW, this was a result of running the linting. I agree it's unrelated but it's possibly something desirable

cibernox:
@davefx I opened this issue: https://github.com/home-assistant/intents/issues/2469

I'm still not sure if this belongs on this repo or on HA Core, but someone will tell me either way.

davefx (approved):

PR #2494, opened 2024-11-11
[EN] Add timer length and name to responses

TheFes:
As Missy mentioned on GitHub, currently the timer length is not included in the response. This can lead incorrectly set timers of which the user will be unware.

This PR adds the timer length and timer name to the response, so the user is aware of what has been set. The set name can be used to ask about the status of the timer.

### DIFF
--- a/responses/en/HassStartTimer.yaml
+++ b/responses/en/HassStartTimer.yaml
@@ -3,5 +3,24 @@ language: en
 responses:
   intents:
     HassStartTimer:
-      default: "Timer started"
-      command: "Command received"
+      default: >
+        {% set h = slots.hours if slots.hours is defined else none %}
+        {% set m = slots.minutes if slots.minutes is defined else none %}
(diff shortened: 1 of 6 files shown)


ViViDboarder:
Rather than calculating the duration here based on the user's input, it would be better if we got the duration back from Home Assistant so we're not potentially using differing interpretations. Is that possible today? Or does that need a core or hasil change?

tetele:
@TheFes can you please update the branch? After some discussion we decided to go forward with this approach, even if it stands against the [current response style guide](https://developers.home-assistant.io/docs/voice/intent-recognition/style-guide#be-concise) due to its UX value.

frenck (approved):

TheFes:
I accidentally based the change for `NL` with this branch, which means the NL changes are now also included in this PR.
Quickly discussed with Frenck, and he agreed to just merge them together. Frenck also did the rebase.

PR #2504, opened 2024-11-11
[BG] Add current time and add to shopping list

MindFreeze:


hristo-atanasov:
Looks good. Only for the hassGetCurrentTime there might be some changes in future, as some of the TTS services does not pronounce the time as they should. I already spoke about this problem with MIke Hansen.

PR #2505, opened 2024-11-11
Add Swedish HassListAddItem

balloob:
Add support for Swedish for `HassListAddItem`

Used AI to generate this as an experiment, needs sign off from a Swedish native speaker.


balloob:
Checked with a Swede, the test sentences make sense.

PR #2512, opened 2024-11-12
Only run tests for affected languages in PRs

MindFreeze:
This should speed up PR checks and reduce the load on CI


tetele (commented):
I like the approach, but I'd add a bit more logic: if **ANY** file was modified in a PR outside of `sentences`, `responses` and `tests`, then run a full test. Otherwise, just run the tests for the modified languages.

I think it's safer that way

MindFreeze:
Alright. I made it run full tests if a language can't be matched or files outside of translations a changed. Tweaked a BG file to test if it matched it properly and it did.

.github/workflows/ci.yaml
```diff
@@ -32,6 +33,46 @@ jobs:
...
+
+          if [ -n "$LANGUAGES" ]; then
+            echo "languages=$LANGUAGES" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Run tests (PR)
+        if: github.event_name == 'pull_request'
+        run: |
+          if [ -n "${{ steps.extract-languages.outputs.languages }}" ]; then
+            script/test --language ${{ steps.extract-languages.outputs.languages }}
```
balloob, line 68: Does this arg work with multiple languages ?

tetele: It seems to work with comma separated values: `pytest tests --language he,ru`

tetele: Yes, code confirms that's the way to go about it https://github.com/home-assistant/intents/blob/f85c1f3c74fed771a0cdabded8f2cc8e0bface47/tests/conftest.py#L27

MindFreeze: one of my first CI runs wrongly matched all languages and this command worked

balloob (approved):

PR #2513, opened 2024-11-12
[DE] Add timer length and name to responses

andreasbrett:
DE version of https://github.com/home-assistant/intents/pull/2494 and https://github.com/home-assistant/intents/pull/2521

> As Missy mentioned on GitHub, currently the timer length is not included in the response. This can lead incorrectly set timers of which the user will be unware.
>
> This PR adds the timer length and timer name to the response, so the user is aware of what has been set. The set name can be used to ask about the status of the timer.

> also adds the duration added or removed from a timer when the respective intents are used


mib1185 (changes requested):
Hi @andreasbrett 
many thanks for adopting this :heart:
But it would be nice, if we could have further test sentences to cover these singular and plural variations (_`Stunde` vs `Stunden`, same for minutes and seconds_) and these `['halbe', '1/2']` variations.
thx :+1:

andreasbrett:
> Hi @andreasbrett many thanks for adopting this ❤️ But it would be nice, if we could have further test sentences to cover these singular and plural variations (_`Stunde` vs `Stunden`, same for minutes and seconds_) and these `['halbe', '1/2']` variations. thx 👍

Thanks for being a pain with the tests :-) Especially with those mere EN->DE migrations I sometimes forget how much more thorough some of the DE tests are compared to EN. Which is a good thing. Tests added.

The 'halbe', '1/2' variations were leftovers from EN that we don't have in the DE sentences. Again well spotted. I removed them, since they're outside the scope of this PR. Let someone else or future-me handle them.

mib1185 (approved):
LGTM :+1:  As always, many thanks for your contribution :heart:

PR #2516, opened 2024-11-12
[FR] A few missing simple intents

jlpouffier:
## Respond
Added the HAssRespond intent.
For now, I decided only to translate two use-case
- Hello
- Who made you

I do not think [the other use-cases](https://github.com/home-assistant/intents/blob/main/sentences/en/homeassistant_HassRespond.yaml) are worth translating as of today.

## Previous
Added the MediaPrevious intent.
A copy of the MediaNext intent

## Nevermind
Added lots of variation for the Nevermind intent

## Timer bug
Last release I added a few sentences using the `reste-t-il` syntax.
It turns out HA Cloud STT sometimes yields `reste t il` without the dashes
image

I decided to use a local expansion rule for that. 
It adds a bit of duplication but it's pretty much local to a single intent

### DIFF
--- a/responses/fr/HassRespond.yaml
+++ b/responses/fr/HassRespond.yaml
@@ -1,0 +1,6 @@
+language: fr
+responses:
+  intents:
+    HassRespond:
+      hello: "Bonjour depuis Home Assistant."
+      creator: "J'ai été créé par Home Assistant et sa communauté mondiale de passionnés de domotique."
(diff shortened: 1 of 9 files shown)


responses/fr/HassRespond.yaml
```diff
@@ -0,0 +1,6 @@
+language: fr
+responses:
+  intents:
+    HassRespond:
+      hello: "Bonjour depuis Home Assistant."
```
piitaya, line 5: Looks weird to say that but ok 😅

piitaya (approved):

responses/fr/HassRespond.yaml
jlpouffier: Do you have a better idea? I am 100% ok to change it if you find better!

PR #2521, opened 2024-11-12
[EN] Add duration added or removed from timer

TheFes:
In addition to #2494 this PR also adds the duration added or removed from a timer when the respective intents are used.

### DIFF
--- a/responses/en/HassDecreaseTimer.yaml
+++ b/responses/en/HassDecreaseTimer.yaml
@@ -3,4 +3,14 @@ language: en
 responses:
   intents:
     HassDecreaseTimer:
-      default: "Timer updated"
+      default: >
+        {% set h = slots.hours if slots.hours is defined else none %}
+        {% set m = slots.minutes if slots.minutes is defined else none %}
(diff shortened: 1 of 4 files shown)


tetele (approved):
Nice!

ViViDboarder (approved):
Looks good. I had to do a sanity check because I was thinking that the slots would already have factored in the "half" > 30 logic since it even shows as the slot value in tests, but apparently not!

TheFes:
> Looks good. I had to do a sanity check because I was thinking that the slots would already have factored in the "half" > 30 logic since it even shows as the slot value in tests, but apparently not!

```yaml
intent:
  name: HassStartTimer
slots:
  minutes: '3'
  seconds: half
details:
  minutes:
    name: minutes
    value: 3
    text: '3'
  seconds:
    name: seconds
    value: 30
    text: half
targets: {}
match: true
sentence_template: <timer_set> [a|the|my] timer for <timer_duration>
unmatched_slots: {}
source: builtin
```
It's included in the `details` but you can't use those in the response

PR #2525, opened 2024-11-12
fix(ci): check language modified to run faster tests

duhow:
Fixes file detection with `changed-files` and runs tests with `--language` as intended.
No more waiting 10 min to complete! 😌


duhow:
Above 400cd45 runs `ca` and `es` as targeted changes 👍🏻

balloob (approved):

PR #2538, opened 2024-11-15
[NL] remove unneeded spaces in lists

TheFes:
Removes unneeded spaces in lists, for example instead of `zet[de |het |een] lamp aan` it is now `zet [de|het|een] lamp aan`


frenck (approved):
Nice 👍 

../Frenck

PR #2507, opened 2024-11-11
Upgrade to hassil 2.0

synesthesiam:
Bumps hassil to 2.0.0: https://github.com/home-assistant/hassil/compare/1.7.4...v2.0.0

There are some significant changes in hassil 2.0, fixing several outstanding bugs. I've done my best to fix the existing sentences, but I need help from some language leaders.

## _test_failures.yaml

I've added a new file to `tests/<language>` called `_test_failures.yaml` which contains a list of sentences that are being tested for but are failing for some reason. Language leaders should start here when looking at their language for this PR. Once fixed, please remove the sentence from `_test_failures.yaml`. 

## Matching Changes

For testing, the biggest change is going from the **first** sentence matched to the **best** sentence matched. This mirrors the code in HA (and will soon replace it) so that certain matches are prioritized:
- Matches with a `name` slot are highest priority, with the longest matching `name` winning
- Matches with more literal text (the text outside of lists and wildcards) are higher priority
- Matches with wildcards are lower priority

In a few cases, this will cause a different sentence template to be recognized than before. For example, if you have a test sentence "open the garage door" and an entity named "garage door", it will always match the template open the {name} instead of open the {device_class} now.

## Hassil Changes

Some high level hassil changes:
- Casing is now preserved, so list values/wildcards/etc. will no longer be lower-cased
- Sentence templates are filtered with a fuzzy regex before matching is attempted. This speeds up most languages, but slows down a few (those languages have `filter_with_regex: false` in their `_common.yaml` files.
- Extra whitespace around wildcards is removed
- Number words will now be used for all genders, cases, etc. in the language (see [unicode-rbnf release](https://github.com/rhasspy/unicode-rbnf/releases/tag/v2.0.0))

Details of hassil changes:

- Allow wildcards to be followed by expansion rules and lists
- Use regular expressions to filter sentence templates
- Add `filter_with_regex` to intent settings and intent data (`false` disables regex filtering)
- Filter text slot list values by required/excluded context during matching
- Use a trie to filter range slot list values based on remaining text to be matched
- Add `required_keywords` section to intent data to skip sentences without specific keywords
- Preserve case during matching 
- Strip punctuation before text processing
- Remove extraneous whitespace from the end of wildcards
- Refactor string matching code into `string_matcher.py`

PR #2539, opened 2024-11-15
[SL] sentences and test fixed according to #2507

andrejs2:


mib1185 (commented):
There is many commented code now, this is not allowed and should be removed!

PR #2544, opened 2024-11-15
[NL] Improvements fan_HassTurnOn/Off

TheFes:
Fix some mistakes introduced in last PR where `<in>` was made optional twice for the same expansion rule.

Reintroduce `<would>` as I noticed my prefix their requests with `wil je`. I removed this in the last change to these intents.
Before they also had sentences like `zou je de ventilator aan willen doen` but these I will not add back again for now. I still don't expect they will be used.


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #2545, opened 2024-11-15
[NL] Improve cover intents

TheFes:
Add `<would>` for cover intents.
Add more options to set position


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #2547, opened 2024-11-15
[NL] Add more options to timer intents

TheFes:
Adds more sentence variants to the timer intents (including `<would>` variants

to make it easier to add more options I added the `timer_prefix`, `timer_suffix`, `my_timer` and `area_timer` expansion rules.
moved some expansion rules which were only used in one place

Also added a few more `skip_words` allowing things like `ik wil een timer van 5 minuten starten` and `kun je voor mij een timer van 5 minuten starten`


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

PR #2560, opened 2024-11-17
Add vacuum variant to match without vacuum name

duhow:
❗ NOTE: This adds a new `slot_combination` to `Vacuum` in `intents.yaml`, in order to skip specifying a name.

As the **vacuum** may have a different name than "vacuum", users will not trigger it by the entity name, but rather as "vacuum". This either forces the user to set an **alias** to the vacuum (named `vacuum`) to start it, or specify the full name for it.

### DIFF
--- a/intents.yaml
+++ b/intents.yaml
@@ -315,6 +315,7 @@ HassVacuumStart:
       description: "Name of an area"
       required: false
   slot_combinations:
+    none: []
     name_only:
       - "name"
     area_only:
@@ -335,6 +336,7 @@ HassVacuumReturnToBase:
       description: "Name of an area"
       required: false
   slot_combinations:
+    none: []
     name_only:
       - "name"
     area_only:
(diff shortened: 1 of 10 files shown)


tetele:
I don't think this will work without some modifications on the `core` repo, so that the concerned intents know how to handle no slots.

intents.yaml
```diff
@@ -315,6 +315,7 @@ HassVacuumStart:
       description: "Name of an area"
       required: false
   slot_combinations:
+    none: []
```
tetele, line 318: Are you sure this is required? The `vacuum` domain [is implied](https://github.com/home-assistant/intents/blob/8022db738785c08f4652eea8419b12fac85ed26d/tests/test_language_sentences.py#L104) by default [in the intent definition](https://github.com/home-assistant/intents/blob/8022db738785c08f4652eea8419b12fac85ed26d/intents.yaml#L308).

PR #2570, opened 2024-11-18
[NL] Translate domain and device class for error responses

TheFes:
@jlpouffier mentioned on Discord that the domains and device classes need translations for the error responses. 

This PR adds those translations.

I also made a change to the responses for covers, they now respond an active state `Wordt geopend` instead of `Geopend` as it will take time to perform the action, and they no longer return the device class. So for example `Gordijn geopend` is now `Wordt geopend`


responses/nl/HassTurnOff.yaml
```diff
@@ -5,10 +5,9 @@ responses:
       default: "Uitgezet"
       lights_area: Verlichting uitgezet
       fans_area: Ventilatoren uitgezet
-      cover: "Gesloten"
+      cover: "Wordt gesloten"
```
frenck, line 8: This seems wrong. Actions return after they have been performed. Not all integrations do, but is the general way things should be handled from our end.

This change seems to be doing the opposite of how things should work.

responses/nl/HassTurnOn.yaml
```diff
@@ -5,10 +5,9 @@ responses:
       default: "Aangezet"
       lights_area: "Verlichting aangezet"
       fans_area: "Ventilatoren aangezet"
-      cover: "Geopend"
+      cover: "Wordt geopend"
```
frenck, line 8: Same as above.

frenck (changes requested):
Let's not combine different concerns in a single PR. Now I leave a review blocking something else, while the original intent of this PR could have just been moving forward.

../Frenck

responses/nl/HassTurnOff.yaml
TheFes: The action has been performed, the integration received the command to close the cover. However, it will take some time before the cover is actually physically closed. However, I reverted the change, and it's `Gesloten` and `Geopend` now.
Like you said, it was not related to the main concern for this PR.

frenck (approved):

PR #2573, opened 2024-11-18
[CS] Make covers area and floor aware, create intents for valves

zbeky:
Update error messages
Make covers area and floor aware
Create intents for valve


zbeky:
Sorry for bigger PR, got carried away when updating HassSetPosition... If it needs to be splitted to multiple PRs, just say so

schizza (approved):

PR #2579, opened 2024-11-18
[DE] Translate error responses

mib1185:
As discussed on discord, we want to translate the possible integration domains and device classes in error responses

PR #2582, opened 2024-11-19
Add more detailed de error responses

wendevlin:
Add domain and device_class option to german error responses, like it is done in french


mib1185:
there is already a pending PR for that -> #2579

mib1185:
close this, since we're working on #2579

PR #2625, opened 2024-11-23
[DE] Add person intents

mib1185:
Add person HassGetState intents - adopted from EN

### DIFF
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -64,7 +64,7 @@ responses:
 
       wo: |
         {% if state.state == "not_home" %}
-          {{ slots.name | capitalize }} ist nicht da
+          {{ slots.name | capitalize }} ist unterwegs
         {% else %}
           {{ slots.name | capitalize }} ist im Bereich {{ state.state }}
         {% endif %}
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -61,3 +61,25 @@ responses:
 
       wie_viele: |
         {{ query.matched | length }}
+
+      wo: |
+        {% if state.state == "not_home" %}
+          {{ slots.name | capitalize }} ist nicht da
+        {% else %}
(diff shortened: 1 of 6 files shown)


responses/de/HassGetState.yaml
```diff
@@ -61,3 +61,25 @@ responses:
 
       wie_viele: |
         {{ query.matched | length }}
+
+      wo: |
+        {% if state.state == "not_home" %}
+          {{ slots.name | capitalize }} ist nicht da
```
andreasbrett, line 67: IMHO this could be misleading. If a person is neither at home nor in a defined zone, I would not expect `... ist nicht da` but rather `... ist außerhalb unterwegs`, `... ist in einem nicht definierten Bereich.` or similar. `nicht da` suggests the answer is binary (at home or not at home). Might just be me though because from a logical POV this works out just fine.

responses/de/HassGetState.yaml
```diff
@@ -61,3 +61,25 @@ responses:
...
+        {% else %}
+          {{ slots.name | capitalize }} ist im Bereich {{ state.state }}
```
andreasbrett, line 69: Proposal: if a person is in an area for quite a long time (to be defined...), provide the duration in the response. I've been using a custom_sentence for "where is xyz" for quite some time now and me and my family really appreciate when the response is something like "Maria ist seit 6 Stunden im Bereich Büro" or "Frank ist seit 1 Stunde im Supermarkt". Adds some unforeseen value ("oh, then he must be back soon").

mib1185: (_at the moment_) we only have the current state available, not the "last updated" or "laste changed" attribute.
Further adding this needs a discussion on the language-leader board.

responses/de/HassGetState.yaml
mib1185: what do you think about just `... ist unterwegs` ?

PR #2657, opened 2024-11-26
[PL] Added Polish support for new intent HassCancelAllTimers

witold-gren:


responses/pl/HassCancelAllTimers.yaml
```diff
@@ -0,0 +1,41 @@
...
+    HassCancelAllTimers:
+      default: >
+        {% if slots.canceled < 1: %}
+        Nie anulowano żadnego minutnika
+        {% else: %}
+          {% if slots.canceled | int | abs == 1 %}
+            {% set timers = 'minutnik' %}
+          {% elif slots.canceled | int | abs in [2, 3, 4] %}
+            {% set timers = 'minutniki' %}
+          {% elif (slots.canceled | int | string)[-1] | int | abs in [2, 3, 4] %}
```
Uriziel01, line 13: Same as with the other PR, currently it will resolve `12, 13, 14` to `minutniki.

Uriziel01 (changes requested):

responses/pl/HassCancelAllTimers.yaml
Uriziel01: Instead of casting from `int` to `string` back to `int` you could have just done a `% 10`, maybe that along with removing the first statement would make it more readable and shorter?

witold-gren: I added extra condition to support proper pronunciation for number 12, 13, 14.  Unfortunately the value `% 10` doesn't work for me.. To do this correctly, we would need to add an additional variable which we then convert and add to another variable which we then compare new value.. it is too complicated for me..

Uriziel01 (approved):

Uriziel01:
@witold-gren Approved, merge at your discretion 🙂

PR #2659, opened 2024-11-26
[SV] Added support for persons.

danielholm:
Added intents, commons and fixtures for person support in Swedish.


danielholm:
I don't understand how to change the response to a more fluid one where persons location are responded more than "Nej, den är x", into more like "Nej, Namn är <i_på> zone"

danielholm:
It seem to be working! pytest ran with pass.

PR #2662, opened 2024-11-26
[FR+EN] Start Vacuums per floor

jlpouffier:
Add a new sentence to allow users to start their vacuum by floor.

The vast majority of users have a single vacuum cleaner per floor that rooms freely all the areas of the floor.

This PR adds sentences to target vacuums per floor, such as
- Clean the first floor.

Done in French and English so that other languages can follow.
I kept it extremely minimal in English as it is not my language


IamTheFij (approved):

ViViDboarder (approved):
Used the wrong profile at first. 👍 

I followed the conversation on Discord the other day, this sounds good in English.

Given the likely safe assumption there is one vacuum per floor or area, a "run the vacuum" or "clean (the|this) room" are probably also useful if received from a device with an area associated. I don't think you have to make the change here though. That could be added later.

jlpouffier:
I'd like to avoid targeting area for now. I think area targeting for vacuum would be a kick ass use case WHEN we are able to actually send a vacuum to a specific area (zone cleaning) 

But today it's impossible without custom commands and actions

PR #2664, opened 2024-11-26
[GL] Add all timer intents

cibernox:
This PR implements all the intents related to adding, querying and cancelling timers.

Those are the last important intents missing.


balloob (approved):
Sweet !

PR #2667, opened 2024-11-26
[NL] fix some spelling mistake

Natfaji:
I think it was meant to be `floor` instead of `area` as every other language uses `floor` for this message and it is located inside `no_domain_in_floor`


TheFes (approved):
Yes, that was a copy/paste mistake and a weird typo.

TheFes:
Thanks @Natfaji

PR #2685, opened 2024-11-28
[FR] extend timers with the concept of reminders

bors-ltd:
Still bringing what I developed for my previous voice assistant.

If you're not interested for the sake of Home Assistant, or if I'm already asking too much too early, and the voice pipeline needs more maturation, just tell me. 

At least the first two commits should be considered. 

There's one sentence commented out, because it's being shadowed by timed commands. I'd like your insight on that.

I know it looks like I'm obsessed with naming timers, but it's just because I display them. :rofl: 
Way easier when you have three of them at the same time.

IMG_20241128_154732
IMG_20241128_154711

(These changes are pushed to my esphome wakeword fork, just no PR open yet.)


jlpouffier:
Arf, indeed you are thinking one or two steps ahead. 

So the main difference between a timer and a reminder is **the time frame.**

You set a timer for 1,2 hours max.
But you usually set reminders days or even weeks into the future.

Today, we have a somewhat OK-ish solution for timers (it has its limitations), but we do not have a great solution for reminders.
Reminders require to be managed, edited, displayed, etc, etc

So for now, we do not want to mix both concepts, unfortunately.
Because we do not want to set ourselves up for failure in the future once reminders will actually come.

I do like the plural thing on the response. This is top-notch.

Would you mind opening a PR just for that, I can merge it right away.

And do not forget, you can extend any intent you like by putting the files (The same you are working on right now) under `consutom_sentences/fr` in your home assistant configuration folder. 
I would suggest you do just that for reminders if they are close to your heart.

Just a note: We know we need to find a solution for reminders. Definitely.

Thx a lot for your time! I really appreciate seeing more people working on the french sentences!

PR #2704, opened 2024-12-02
[FR] overhaul of the weather intent, with detailed report

bors-ltd:
The current one would never give the weather condition, as the state was never "clear" or "sunny", but in French already.

The wind description is using the Beaufort scale.

Heavily inspired by the Hungarian template, kudos to RViktor!

I'd like to push it even further, with sentences like "va-t-il pleuvoir/neiger/geler ?", but it requires the Météo-France integration, which not everyone has. I stick to the built-in Met.no integration. Unless there is a way to ship intents along with the Météo-France integration itself.

A more reasonable next step would be "quelle est la météo pour demain ?".

### DIFF
--- a/responses/fr/HassGetWeather.yaml
+++ b/responses/fr/HassGetWeather.yaml
@@ -3,22 +3,189 @@ responses:
   intents:
     HassGetWeather:
       default: >
-        {% set weather_condition = {
-          'clear': 'et le temps est clair',
-          'clear-night': 'et la nuit est claire',
-          'cloudy': 'et le temps est nuageux',
-          'exceptional': 'et les conditions météos sont exceptionnelles',
-          'fog': 'avec du brouillard',
-          'hail': 'avec de la grêle',
-          'lightning': 'avec de l\'orage',
-          'lightning-rainy': 'avec de l\'orage et de la pluie',
-          'partlycloudy': 'et un temps partiellement nuageux',
-          'pouring': 'et une pluie battante',
-          'rainy': 'et de la pluie',
-          'snowy': 'et de la neige',
-          'snowy-rainy': 'avec pluie et neige mêlées',
-          'sunny': 'et un temps ensoleillé',
-          'windy': 'avec un temps venteux',
-          'windy-variant': 'et un temps variable et venteux'
-        } %}
-        {{ state.attributes.get('temperature') }} {{ state.attributes.get('temperature_unit') }} {{ weather_condition.get((state.state | string).lower(), "") }}
+        {# Reuse weather condition already formatted and translated #}
+        {# just invert telling the condition then temperature #}
+        {{ state.state }}
+
+        {% set temperature = state.attributes.get('temperature') %}
+        {% if temperature is number %}
+          {% if temperature < 0 %}
+            {% set temperature_string = 'moins ' ~ (temperature * -1) | string | replace('.', ',') %}
+          {% else %}
+            {% set temperature_string = temperature | string | replace('.', ',') %}
+          {% endif %}
+          {% set temperature_unit = state.attributes.get('temperature_unit', '') %}
+          et il fait {{ temperature_string }} {{ temperature_unit.replace('°C', 'degrés') }}.
+        {% endif %}
+
+      # Starting from here, heavily inspired by the Hungarian template, kudos to RViktor
+      wind: >
+        {% set wind_speed = state.attributes.get('wind_speed') %}
+        {% set wind_speed_unit = state.attributes.get('wind_speed_unit', '') %}
+        {% set wind_bearing = state.attributes.get('wind_bearing') %}
+        {% set wind_gust_speed = state.attributes.get('wind_gust_speed') %}
+
+        {# Compute the Beaufort scale when possible #}
+        {% set beaufort_scale = '' %}
+        {% set wind_speed_ms = none %}
+        {% if wind_speed_unit == 'm/s' %}
+          {% set wind_speed_ms = wind_speed %}
+        {% elif wind_speed_unit == 'km/h' %}
+          {% set wind_speed_ms = wind_speed / 3.6 %}
+        {% endif %}
+        {% if wind_speed_ms is number %}
+          {# Borrowed from the speed converter in core, better to have a convert filter #}
+          {# Rounding is important, we want 3.9 to become 4, not 3 #}
+          {% set beaufort_force = (((wind_speed_ms / 0.836) ** 2) ** (1 / 3)) | round | int %}
+          {% set beaufort_scale = {
+            0: 'Vent calme',
+            1: 'Très légère brise',
+            2: 'Légère brise',
+            3: 'Petite brise',
+            4: 'Jolie brise',
+            5: 'Bonne brise',
+            6: 'Forte brise',
+            7: 'Grosse brise',
+            8: 'Coup de vent',
+            9: 'Fort coup de vent',
+            10: 'Tempête',
+            11: 'Violente tempête',
+            12: "Vent d'ouragan",
+          }.get(beaufort_force, 'Vent') %}
+        {% elif wind_speed is number %}
+          {# Other wind speed unit, no scale, just the speed #}
+          {% set beaufort_scale = 'Vent' %}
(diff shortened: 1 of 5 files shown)


jlpouffier (changes requested):
Before I even start reviewing this one more closely ☺️ :

Why no test changes?
I would be more comfortable with test changes for this PR to be honest.

Additional note: The build is failing you have an issue in one of your templates ;)

responses/fr/HassGetWeather.yaml
```diff
@@ -3,22 +3,200 @@ responses:
...
+            {% set wind_scale = 'Grosse brise' %}
+          {% elif wind_speed <= 74 %}
+            {% set wind_scale = 'Coup de vent' %}
+          {% elif wind_speed <= 88 %}
+            {% set wind_scale = 'Fort coup de vent' %}
+          {% elif wind_speed <= 102 %}
+            {% set wind_scale = 'Tempête' %}
+          {% elif wind_speed <= 117 %}
+            {% set wind_scale = 'Violente tempête' %}
+          {% else %}
```
jlpouffier, line 54: Can you transform this `else` into a `eilf wind_speed > 117`.
I know it should be the same, but I want to avoid showing `vent d'ouragan` if the wind_speed is incorrectly set to `unavailable` or `-1` for example

responses/fr/HassGetWeather.yaml
```diff
@@ -3,22 +3,200 @@ responses:
...
+            {% set wind_scale = "vent d'ouragan" %}
```
jlpouffier, line 55: Capitalize? LIke all the other ones?

responses/fr/HassGetWeather.yaml
```diff
@@ -3,22 +3,200 @@ responses:
...
+          {% elif wind_bearing < 270 %}
+            {% set wind_bearing_text = 'sud-ouest' %}
+          {% elif wind_bearing < 315 %}
+            {% set wind_bearing_text = 'ouest' %}
+          {% elif wind_bearing <= 360 %}
+            {% set wind_bearing_text = 'nord-ouest' %}
+          {% else %}
+            {% set wind_bearing_text = 'inconnue' %}
+          {% endif %}
+        {% endif %}
```
jlpouffier, line 83: I find the sentence "Dont la direction est Ouest" a bit weird.
IN itself it's not that bad but since you already have all the array to translate it, how about you translate a better-sounding sentence inside the translation array itself and use it below?

Something like
```yaml
{% set wind_bearing_text = '' %}
{% if wind_bearing %}
  {% if wind_bearing < 45 %}
    {% set wind_bearing_text = 'venant du nord' %}
...
  {% endif %}
{% endif %}
...
{{ wind_scale }} de {{ wind_speed | int }} {{ wind_speed_unit }}, {{ wind_bearing_text }}, avec des rafales à {{ wind_gust_speed | int }}
...
```

responses/fr/HassGetWeather.yaml
```diff
@@ -3,22 +3,200 @@ responses:
...
+
+        {% if wind_speed_unit == 'km/h' %}
+          {% set wind_speed_unit = 'kilomètres par heure' %}
+        {% endif %}
```
jlpouffier, line 87: The native wid speed unit on the weather entity can be
`m/s, km/h, mi/h, ft/s or kn`

I would say, let's support them all, especially for our french-speaking colleague leaving elsewhere

responses/fr/HassGetWeather.yaml
```diff
@@ -3,22 +3,200 @@ responses:
...
+        {% elif wind_speed %}
+          {{ wind_scale }} de {{ wind_speed | int }} {{ wind_speed_unit }}.
+        {% else %}
+          Aucune information météo sur le vent n'est disponible.
+        {% endif %}
+
+      uv: >
+        {% set uv_index = state.attributes.get('uv_index') %}
+
+        {% if uv_index %}
```
jlpouffier, line 104: Does this particular scale have a name?
I quite like the comment on the Beaufort scale for the wind.

sentences/fr/weather_HassGetWeather.yaml
```diff
@@ -3,12 +3,53 @@ intents:
...
+          - "(<quelest>|<donnemoi>) le détail du vent [<aujourdhui>]"
+        response: wind
+      - sentences:
+          - "(<quelest>|<donnemoi>) la vitesse du vent [<aujourdhui>] à [<le>]{name}"
+          - "(À|A) <quel> vitesse le vent souffle [<aujourdhui>] à [<le>]{name}"
+          - "(<quelest>|<donnemoi>) la météo du vent [<aujourdhui>] (pour|à) [<le>]{name}"
+          - "(<quelest>|<donnemoi>) le détail du vent (pour|à) [<le>]{name}"
+        requires_context:
+          domain: weather
+        response: wind
```
jlpouffier, line 55: We have a specific device class under the `sensor` domain called `wind_speed`.
I do not think it's a problem we are removing most (if not all) sentences targeting specific device classes under the `sensor` domain.

@piitaya thoughts?

jlpouffier:
Awesome work ❤️

Review done, some comments added

responses/fr/HassGetWeather.yaml
bors-ltd: You're right, it's a leftover of the Hungarian version through an automated translator.

responses/fr/HassGetWeather.yaml
bors-ltd: This is the scale I found on the Météo-France website. I could leave a link as the source, but I have no idea how much of a standard it is.

sentences/fr/weather_HassGetWeather.yaml
```diff
@@ -3,12 +3,53 @@ intents:
...
-          - "(<quelest>|donne[s][-moi]) (le|la|les) (temps|météo) (pour|à) [<le>]{name}"
-          - "(Donne-moi|Dis-moi) le temps qu'il fait à [<le>]{name}"
+          - "(<quelest>|<donnemoi>) (le|la|les) (temps|météo) (pour|à) [<le>]{name}"
+          - "<donnemoi> le temps qu'il fait à [<le>]{name}"
           - "Quel temps fait-il à [<le>]{name}"
         requires_context:
           domain: weather
+
+      - sentences:
+          - "<donnemoi> (un[e]|le) [bulletin|rapport] météo [détaillé[e]]"
```
bors-ltd, line 17: Just realised with so many optional words, it collides with "donne-moi la météo". Will split in two sentences.

sentences/fr/weather_HassGetWeather.yaml
piitaya: If we only target by name, it shouldn't be a issue right?

responses/fr/HassGetWeather.yaml
RSickenberg: In French, `kn` would give `nd` for `noeuds`, idk what is the best :)

sentences/fr/_common.yaml
```diff
@@ -587,6 +588,8 @@ expansion_rules:
   atil: "(ont|a)[-][ ][t][ ][-][(il[s]|elle[s])]"
   quel: "quel[le][s]"
   quelest: "<quel> (est|sont)"
+  donnemoi: "(donne[s]|dis)[(-| )moi]"
+  what_is_the_class_of_name: "<quelest> (le |la |l'|les )<class> [(indiqué[e][s]|mesuré[e][s]|renvoyé[e][s]|restant[e][s]|retourné[e][s]|utilisé[e][s]|produit[e][s]|consommé[e][s]|donné[e][s]) ][(par|<dans>|sur)] [<le>]{name} [<dans> [<le>]{area}]"
```
jlpouffier, line 592:
```suggestion

```

jlpouffier: Let's remove this one, it's unused anymore.
And then let's finally merge this pr ;) 

Thx a lot man!

jlpouffier (changes requested):

jlpouffier (approved):

PR #2708, opened 2024-12-03
[DE] HassGetState floor support

andreasbrett:
DE version of https://github.com/home-assistant/intents/pull/2706.

Also added some more aliases for "Geschoss" (Stock, Stockwerk). Not sure though why "Geschoss" was in there in the first place. Couldn't find any test sentence making use of that. If it is not needed (can you even address floors by number and not their name?), we should get rid of those 3 as this increases the sentence count quite a bit.


andreasbrett:
@mib1185 what's your opinion? I don't get the "Geschoss" part in the floor expansion rule.

mib1185 (commented):
We had a discussion about it in https://github.com/home-assistant/intents/pull/2121#discussion_r1550217615, but you're right, we missed to add proper test sentences then :see_no_evil: So I don't I think we need these additional aliases.

PR #2224, opened 2024-06-07
[EN] HassGetState optimizations

tetele:
Apart from optimizing the sentences for `HassGetState` a bit, I've also introduced a new set of sentences specifically for querying lights.

`Is the light off in the kitchen` usually means asking if *all* of the lights are off.

Inspired by #2218 

## Summary by CodeRabbit

- **New Features**
  - Enhanced querying capabilities for Home Assistant entity states, including specific areas like the bedroom and kitchen.
  - Added new sentence structures for more detailed and varied queries.

- **Refactor**
  - Improved clarity and coverage of existing sentence structures for querying entity states.

- **Tests**
  - Added and refined test queries for specific devices and areas to ensure accurate responses.

PR #2711, opened 2024-12-04
[EN] New "any" query for HassGetState for area, domain and state

ViViDboarder:
Now will match queries like "is the bedroom light off" if there is no entity specifically named "bedroom light" in the bedroom. It will instead consider if "any" bedroom light is on and respond in kind.


ViViDboarder:
Closing in favor of #2224

PR #2725, opened 2024-12-06
[PT-BR] Improve cover control

bgoncal:


sentences/pt-br/_common.yaml
```diff
@@ -364,6 +364,8 @@ expansion_rules:
   falta: "(falta|resta)"
   alarme: "(alarme[s]|cronômetro[s]|timer[s]|temporizador[es])"
   pausar: "(pausa[r]|pause)"
+  cortina: "(cortina[s]|curtinha[s]|persiana[s])"
```
luyzfernando08, line 367: Never heard of "Curtinha", is it some regional term?

luyzfernando08, line 367: I would remove the article 'as', which is already covered via 'a[s]'.
Do you think it is feasible to add 'das' and 'dos'?

sentences/pt-br/_common.yaml
```diff
@@ -364,6 +364,8 @@ expansion_rules:
...
+  artigos: "[(o[s]|a[s]|as|da|do|de)]"
```
luyzfernando08, line 368: I would remove the article 'as', which is already covered via 'a[s]'.
Do you think it is feasible to add 'dos'?

sentences/pt-br/_common.yaml
bgoncal: Nope, it's a common mistake from STT providers (including the one home assistant cloud uses) when you say "cortina", so this will cover those cases as well

sentences/pt-br/_common.yaml
bgoncal: Oh, good catch

PR #2471, opened 2024-11-06
Improve sentence structure for common.yaml

mrdarrengriffin:
Improves the standardisation of the sentence structure, ensuring spacing is kept out of variations and expressions

PR #2727, opened 2024-12-06
[en] Enable the use of possessives for some nouns

ViViDboarder:
Rather than only accepting `the`, many sentences should also accept a possessive like `my`.

This doesn't make sense for all instances of `<light>`, so I'll make another refactor commit after this one

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -364,22 +364,23 @@ lists:
     wildcard: true
 
 expansion_rules:
-  name: "[the] {name}"
-  area: "[the] {area}"
-  floor: "[the] {floor} [floor]"
+  the: "(the|my|our)"
+  name: "[<the>] {name}"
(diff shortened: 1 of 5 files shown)


frenck (commented):
There is a merge conflict @ViViDboarder, could you take a look?

Thanks! 👍 

../Frenck

ViViDboarder:
Ah. The conflict is because of #2471, which I more or less included here because the PR was listed as draft. I'll sort through and see what I come up with.

tetele:
This should probably be closed in favor of #2728

ViViDboarder:
Conflicts resolved!

tetele (approved):
There are some places still using stuff like `a|the|my` (e.g. timer intents) but we can fix that later.

This looks like a good refactoring to me

mrdarrengriffin (approved):

PR #2728, opened 2024-12-06
[En] remove `the` from `<light>`

ViViDboarder:
This doesn't always make sense in a sentence to allow `the` to prefix. It's not
that big of a deal, however once adding `<the>` in, it does increase the number
of matching sentences by even more. This patch can be considered an optimization
rather than a quality improvement.

Based on top of #2727


sentences/en/light_HassTurnOff.yaml
```diff
@@ -39,25 +33,25 @@ intents:
           - "<floor> <light> (off|out)"
           - "<light> (off|out) [<in>] <floor>"
           - "<turn> (off|out) [<all>] <floor> <light>"
-          - "<turn> (off|out) [<all>] <light> [<in>] <floor>"
+          - "<turn> (off|out) [<all>] [<the>] <light> [<in>] <floor>"
```
tetele, line 36: `<all>` is already `(all[[ of] <the>]|every[ single]|each[ and every])`. Why would you add `[<the>]` as an extra? You could also get a match for `every the light`, which you don't want

sentences/en/light_HassTurnOff.yaml
```diff
@@ -18,18 +18,12 @@ intents:
           - "<area> <light> (off|out)"
           - "<light> (off|out) [in] <area>"
           - "<turn> (off|out) [<all>] <area> <light>"
-          - "<turn> (off|out) [<all>] <light> <in> <area>"
+          - "<turn> (off|out) [<all>] [<the>] <light> in <area>"
```
tetele, line 21: <all> is already (all[[ of] <the>]|every[ single]|each[ and every]). Why would you add [<the>] as an extra? You could also get a match for every the light, which you don't want

sentences/en/light_HassLightSet.yaml
```diff
@@ -83,8 +83,8 @@ intents:
         response: "color"
 
       - sentences:
-          - "[<set>] [[the] color of] [all] <light> [to] {color}"
-          - "[<set>] [[the] color of] [all] <light> (<in_here>;[to] {color})"
+          - "[<set>] [[the] color of] [all] [<the>] <light> [to] {color}"
```
tetele, line 86: You could use `<all>` instead of `all`
```suggestion
          - "[<set>] [[the] color of] [<all>] <light> [to] {color}"
```

tetele (changes requested):
Please review all combinations of `<all> [<the>]`, as they could match `all the the`, which is unwanted.

sentences/en/light_HassTurnOff.yaml
ViViDboarder: Ah. This is because I wanted it to match in cases <all> was omitted. I'll change to `[(<all>|<the>)]`.

sentences/en/light_HassLightSet.yaml
ViViDboarder: Hmm. Lots of other refactors here too. I'll try to make some minimal adjustments and then do a bigger refactor of the sentences as a whole in another PR.

tetele (approved):
Needs a rebase, but LGTM

ViViDboarder:
Rebased anyway to make the commit history cleaner.

mrdarrengriffin (approved):

mrdarrengriffin (approved):

PR #2737, opened 2024-12-11
[LT] Create default_nevermind.yaml

braticks:
Added Lithuanian translations for HassNevermind intent.

Changes include:
- Added sentences/lt/default_nevermind.yaml with common Lithuanian phrases for cancellation
- Added corresponding test cases
- Follows the standard intent format

The translations cover common Lithuanian expressions used to cancel or dismiss actions.


balloob:

PR #2742, opened 2024-12-12
[ES] Adding sentences to cancel timers

davefx:


andreasbrett:
@davefx I'll leave the merging to you, since this is out of my scope.

PR #2744, opened 2024-12-12
[en] Tighten the matching for general turn on and off statements

ViViDboarder:
This will prevent ambiguous sentences from being recognized as an turn on or off statement.

Eg. If you currently say 'Is the unknown device on', the '<area> <name> on' sentence was matched and the user recieved a confusing error suggesting there is no area called 'Is'. This will require an explicit request to 'turn|switch|etc' something on or off.

### DIFF
--- a/sentences/en/homeassistant_HassTurnOff.yaml
+++ b/sentences/en/homeassistant_HassTurnOff.yaml
@@ -3,9 +3,10 @@ intents:
   HassTurnOff:
     data:
       - sentences:
-          - "<turn> off (<area> <name>|<name> [in <area>])"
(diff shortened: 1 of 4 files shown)


tetele:
I've seen numerous Youtube videos where people prefer the minimal way of toggling their equipment, such as "couch lights on". Merging this would prevent that behavior and I don't want to do it.

As an alternative, we could enhance the matching system so that it would rather match [this sentence](https://github.com/home-assistant/intents/blob/8aa6cf0aab1dc9fc4ff5d3cf78a40cbfa77a5b1b/sentences/en/homeassistant_HassGetState.yaml#L13) first. But that's a separate concern entirely.

PR #2749, opened 2024-12-13
[SK] cleaning in prepositions

softko:
Removed prepositions from "in" expansion rule. They are not real area prepositions, but area name parts - https://github.com/home-assistant/intents/discussions/480#discussioncomment-11485056

### DIFF
--- a/tests/sk/light_HassTurnOn.yaml
+++ b/tests/sk/light_HassTurnOn.yaml
@@ -40,10 +40,9 @@ tests:
       name: HassTurnOn
       slots:
         area:
+          - pred vchodom
+          - pred vchodom
           - vchode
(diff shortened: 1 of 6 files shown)


tests/sk/light_HassTurnOn.yaml
```diff
@@ -40,10 +40,9 @@ tests:
       name: HassTurnOn
       slots:
         area:
+          - pred vchodom
+          - pred vchodom
```
LubosKadasi, line 44: duplicita

softko: opravene, aj v on aj off

LubosKadasi (approved):

PR #2752, opened 2024-12-15
[DE] say xyz

andreasbrett:
DE version of https://github.com/home-assistant/intents/pull/2750

@synesthesiam Not sure if this is known/intentional, but the tests for the response does not work actually (I tested in both DE and EN). The slot is tested but the actual response sentence is never tested. The following validates just fine.

```yaml
tests:
  - sentences:
      - "say I'm a little teapot"
    intent:
      name: HassRespond
      slots:
        response: "I'm a little teapot"
    response: "I'm something completely different"
```

### DIFF
--- a/responses/de/HassRespond.yaml
+++ b/responses/de/HassRespond.yaml
@@ -2,6 +2,7 @@ language: de
 responses:
   intents:
     HassRespond:
+      default: ""
       hello: "Hallo vom Home Assistant."
       listening: "Nein, ich höre nur zu, wenn du das Signalwort sprichst."
       data: "Deine Daten liegen auf deinem Home Assistant server."
--- a/sentences/de/_common.yaml
+++ b/sentences/de/_common.yaml
@@ -535,6 +535,8 @@ lists:
     wildcard: true
   timer_command:
     wildcard: true
+  response:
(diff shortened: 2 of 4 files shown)


mib1185:
We should still check why the `response` seems irrelevant in the tests in this case!

andreasbrett:
> We should still check why the `response` seems irrelevant in the tests in this case!

Seems intentional. Asserts are skipped for `HassRespond` (introduced in the original PR linked above).

https://github.com/home-assistant/intents/blob/0bf0869b82f8b4b272322e04bccfca193e8ed02b/tests/test_language_sentences.py#L241-L244

mib1185:
this looks weird, since it will stop the whole test file after very first `HassRespond` with `"response" in result.entities` :thinking:

andreasbrett:
I agree, don't understand the reasoning here but since it is rather simple and the only change in the test script, chances are rather high this was intentionally put before any assert.

@synesthesiam can you help us out here? Why was this return-early approach introduced and why will it return early for ALL `HassRespond` responses and not just the newly introduced `default` one?

mib1185:
Fixed in #2753

PR #2761, opened 2024-12-18
[PT] Fixed typo on turning lights on

modem:
Fixed typo


abmantis (approved):
Thanks! These responses are not yet used in PT (they were probably copied from other language but never implemented), but I plan to eventually use them.

PR #2768, opened 2024-12-21
[SR] Add GetWeather

ddxic:
added Serbian intents for Get Weather


ddxic:
Closed as another user mixed antoher language into this PR...

PR #2779, opened 2024-12-23
[SR] Add ListAddItem

ddxic:
added Serbian intents for List Add Item


ddxic:
Closed as another user mixed antoher language into this PR...

PR #2784, opened 2024-12-23
[en] Add expansion for smart apostrophes

ViViDboarder:
Maybe this should be in hassil, but for now this will help with matching when users type on their phones.


andreasbrett:
That should definitely be resolved upstream in hassil.

ViViDboarder:
Yea, I agree, but I never really got clear answers there. I was thinking about doing some normalization, but that could also have unintended consequences if not done properly. For example, we can’t just normalize input text, but also need to normalize all slot values too. 

Some form of edit distance sorting and fuzzy matching would also alleviate the problem, but that’s even more complex. 

I figured this could be a short term patch until a decision is made and hassil is patched.

balloob:

---

PR #2785, opened 2024-12-25
[LV] Simple responses intent

gintsgints:
Tests run fine, but for some reson

`python3 -m script.intentfest parse --language lv --sentence 'saki Es esmu maza tējkanniņa'`

gives error


responses/lv/HassRespond.yaml
```diff
@@ -0,0 +1,10 @@
+language: lv
+responses:
+  intents:
+    HassRespond:
+      default: ""
+      hello: "Sveicieni no Home Assistant."
+      listening: "Nē, Es klausos tikai tad, kad tu pasaki modināšanas frāzi."
```
klejejs, line 7:
```suggestion
      listening: "Nē, Es klausos tikai tad, kad tu mani uzrunā."
```
Maybe like this?

responses/lv/HassRespond.yaml
```diff
@@ -0,0 +1,10 @@
...
+      data: "Jūsu dati tiek nosūtīti uz Home Assistant serveri."
+      commands: "Lai uzzinātu ko jūs varat man jautāt, apmeklējiet H.A. punkts I.O. slīpsvītra voice."
+      creator: "Mani izveidoja brīnišķīgā Home Assistant komūna, sastavoša no domātājiem no visas pasaules."
```
klejejs, line 10:
```suggestion
      creator: "Mani izveidoja brīnišķīgā Home Assistant kopiena, kas sastāv no domātājiem visā pasaulē."
```
What do you think of this?

sentences/lv/homeassistant_HassRespond.yaml
```diff
@@ -0,0 +1,26 @@
+language: "lv"
+intents:
+  HassRespond:
+    data:
+      - sentences:
+          - "(pasaki|saki) {response}"
```
klejejs, line 6:
```suggestion
          - "[pa]saki {response}"
```

sentences/lv/homeassistant_HassRespond.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+      - sentences:
+          - "(sveiks|sveiki|hi|čau) [home assistant]"
```
klejejs, line 9:
```suggestion
          - "(sveik(s|i)|čau) [home assistant]"
```
I don't think we should include `hi` in Latvian language.

klejejs (changes requested):
Overall looks good, but I added a few suggestions for shorter definitions and sentences that could grammatically make more sense in Latvian.
Also, regarding the test failures, I believe it is because the response slot is not present when you are running the test from CLI.

responses/lv/HassRespond.yaml
gintsgints: I would like somehow to point out wakeup phrase. Because it does not listen on any "mani uzrunā"

responses/lv/HassRespond.yaml
gintsgints: Actualy I do not like nether :). But could not get better translation. Anyways yours sound bit better.

gintsgints:
> Also, regarding the test failures, I believe it is because the response slot is not present when you are running the test from CLI.

Should I fix it somehow, or it could go as is?

klejejs:
> > Also, regarding the test failures, I believe it is because the response slot is not present when you are running the test from CLI.
> 
> Should I fix it somehow, or it could go as is?

If the tests pass as setup, it is OK.

responses/lv/HassRespond.yaml
klejejs: Yes, that sounds good as well 👍🏻

klejejs (approved):
Thanks for the PR! 🚀

PR #2790, opened 2024-12-27
[EN] Add floor support to HassClimateGetTemperature

jackwilsdon:
English equivalent of #2788. Fixes https://github.com/home-assistant/core/issues/133733.

I removed `is it (hot|cold|warm|cool) [in <area>]`, as it didn't seem to work and doesn't really make much sense considering the response ("N degrees"). Happy to add it back (and fix it) if it's still wanted.

### DIFF
--- a/sentences/en/climate_HassClimateGetTemperature.yaml
+++ b/sentences/en/climate_HassClimateGetTemperature.yaml
@@ -13,7 +13,7 @@ intents:
       # Get temperature of a climate in an area or with a name
       - sentences:
           - "<what_is> [the] [current] <temp> [of] <name>"
-          - "<what_is> [the] [current] <temp> <in> <area>"
+          - "<what_is> [the] [current] <temp> <in_area_floor>"
           - "<what_is> <area> [current] <temp>"
           - "<what_is> <name> [current] <temp>"
-          - "how (hot|cold|warm|cool) is (<name>|[it <in>] <area>)"
+          - "how (hot|cold|warm|cool) is (<name>|[it] <in_area_floor>)"
--- a/intents.yaml
+++ b/intents.yaml
@@ -260,6 +260,9 @@ HassClimateGetTemperature:
     area:
       description: "Name of an area"
       required: false
+    floor:
+      description: "Name of a floor"
+      required: false
   response_variables:
     state:
       description: "State of the entity that contains the temperature"
--- a/tests/en/climate_HassClimateGetTemperature.yaml
+++ b/tests/en/climate_HassClimateGetTemperature.yaml
@@ -29,6 +29,15 @@ tests:
     response: "68 degrees"
 
   - sentences:
+      - "what's the temperature in the basement?"
+      - "how hot is it in the basement?"
+    intent:
+      name: HassClimateGetTemperature
+      slots:
+        floor: Basement
+    response: "1 degree"
+
+  - sentences:
       - "what is the temperature of the office thermostat?"
       - "what's the office thermostat temperature?"
       - "how hot is the office thermostat?"


tetele (approved):

sentences/en/climate_HassClimateGetTemperature.yaml
```diff
@@ -2,18 +2,19 @@ language: "en"
 intents:
   HassClimateGetTemperature:
     data:
-      # What is...
+      # Get temperature of a climate in the same area as a satellite device
+      - sentences:
+          - "<what_is> [the] [current] <temp> [[in] here]"
```
ViViDboarder, line 7: Instead of `[[in] here]`, you should use `[<here>]`, which includes “in here” but also more variations.

sentences/en/climate_HassClimateGetTemperature.yaml
```diff
@@ -2,18 +2,19 @@ language: "en"
...
+          - "how (hot|cold|warm|cool) is it [[in] here]"
+        requires_context:
+          area:
+            slot: true
+
+      # Get temperature of a climate in an area or with a name
       - sentences:
           - "<what_is> [the] [current] <temp> [of] <name>"
-          - "<what_is> [the] [current] <temp> [in <area>]"
+          - "<what_is> [the] [current] <temp> in <area>"
```
ViViDboarder, line 16: Instead of `in <area>`, I think `<in_area_floor>` will be more expansive.

sentences/en/climate_HassClimateGetTemperature.yaml
```diff
@@ -2,18 +2,19 @@ language: "en"
...
           - "<what_is> <area> [current] <temp>"
           - "<what_is> <name> [current] <temp>"
-      # How...
-      - sentences:
-          - "how (hot|cold|warm|cool) is it [in <area>]"
+          - "how (hot|cold|warm|cool) is it in <area>"
           - "how (hot|cold|warm|cool) is <name>"
```
ViViDboarder, line 20: This same sentence could work for an area or floor too. Maybe `(<name>|<area>|<floor>)` would do.

ViViDboarder: Actually, this is pretty close to the sentence above. Maybe they can be combined to be `how (hot|cold|warm|cool) is (<name>|[it] <in_area_floor>)`.

ViViDboarder (changes requested):
Thanks for the PR. Some changes to support queries like “how cool is the living room” rather than strictly “how cool is it in the living room”.

sentences/en/climate_HassClimateGetTemperature.yaml
jackwilsdon: HassClimateGetTemperature doesn't support floor at the moment from what I can tell (https://github.com/home-assistant/core/blob/f8399b2c0fa6e71fbfc2ff28d96bf8f7bf133364/homeassistant/components/climate/intent.py#L18) , so I've kept this as just area.

sentences/en/climate_HassClimateGetTemperature.yaml
jackwilsdon: Combined: [9177e6..74a7e7](https://github.com/home-assistant/intents/compare/9177e6b793e0cd84f02c16fe43986b47ac04bd42..00c49802397480507cf5d5d65f81d1ce14ffcebf)

I've made `[it in]` optional to allow "how hot is the living room?". Not sure if this would introduce some ambiguity with `<name>` though.

sentences/en/climate_HassClimateGetTemperature.yaml
jackwilsdon: Done: [9177e6..b887e9](https://github.com/home-assistant/intents/compare/9177e6b793e0cd84f02c16fe43986b47ac04bd42..b887e91e2ddd86c03be98158ac6f327f9a8a1953)

sentences/en/climate_HassClimateGetTemperature.yaml
ViViDboarder: Would be good, but I’m ok merging the PR supporting what’s in Core today.

ViViDboarder: Could still use `<in>` rather than the literal to cover other relevant prepositions. That way it covers “what is the temperature of the living room” as well.

ViViDboarder (changes requested):
Just one change unless you decide to add floor support. Floor support would actually be really useful here as many places only have a single thermostat per floor.

ViViDboarder (approved):
Looks good! Did you get a core PR for floors? Mind linking to it so we can make sure that gets merged?

mib1185:
Maybe split out the part about floor support, so we can get this merged with area support now.

jackwilsdon:
Split out area support into #2793 :+1:

mib1185:
#2793 is merged, you need to rebase this PR now to solve the conflict.

mib1185 (commented):
- needs https://github.com/home-assistant/core/pull/134098

PR #2798, opened 2024-12-29
[Various languages] Fix AM/PM for times with hour of 12

robtpaton:
**What**
Change condition for when to use AM vs PM

**Why**
Currently a time of 12:xx is given suffix of AM which is incorrect. 00:xx is AM, 12:xx is PM.

### DIFF
--- a/responses/ar/HassGetCurrentTime.yaml
+++ b/responses/ar/HassGetCurrentTime.yaml
@@ -5,8 +5,12 @@ responses:
       default: >
         {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
 
-        {% if slots.time.hour < 12: %}
+        {% if slots.time.hour == 0: %}
+        12:{{ minute_str }} صباحا
+        {% elif slots.time.hour < 12: %}
         {{ slots.time.hour }}:{{ minute_str }} صباحا
+        {% elif slots.time.hour == 12: %}
+        12:{{ minute_str }} مساءا
         {% else: %}
         {{ slots.time.hour - 12 }}:{{ minute_str }} مساءا
         {% endif %}
--- a/responses/ar/HassGetCurrentTime.yaml
+++ b/responses/ar/HassGetCurrentTime.yaml
@@ -5,7 +5,7 @@ responses:
       default: >
         {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
 
-        {% if slots.time.hour <= 12: %}
+        {% if slots.time.hour < 12: %}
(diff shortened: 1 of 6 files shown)


tetele:
Are you really familiar with all those languages you modified, @robtpaton ? Are you sure they all use the same logic?

Also, your modification would output "00:15 AM", "06:15 AM" and "00:15 PM". If we're gonna fix it, I think we should fix it properly.

robtpaton:
> Are you really familiar with all those languages you modified, @robtpaton ? Are you sure they all use the same logic?

No, only English! Seems like a good thing to fix this in as many languages as possible at the same time. I see labels have automatically been added to the PR for the other languages, I'll look at how to tag language leaders of each of those for reviewing.

> Also, your modification would output "00:15 AM", "06:15 AM" and "00:15 PM". If we're gonna fix it, I think we should fix it properly.

Good point 🤦

tetele:
> I'll look at how to tag language leaders of each of those for reviewing.

They've already been tagged. My point was that it's not a good idea to have a multi-language PR.

robtpaton:
> > I'll look at how to tag language leaders of each of those for reviewing.
> 
> They've already been tagged. My point was that it's not a good idea to have a multi-language PR.

OK, I took an educated guess that the setup is similar for these languages, since the existing logic was similar, but I see your point and accept the contributing guide does say to prefer single language PRs.

My follow-up question would be: what is the best mechanism for raising "I think this might be wrong across multiple languages?" type issues? Maybe we could add some more nuance/details in the contributing guidelines.

Have taken another look at the logic also 👍

tetele (approved):
OK for English

tetele:
> what is the best mechanism for raising "I think this might be wrong across multiple languages?"

Usually language leaders keep an eye out for English changes. To be sure, you could add an issue and tag it with other language tags. The language leaders will then be assigned. I can't remember if automatically or not.

synesthesiam:
Related issue: https://github.com/home-assistant/core/issues/134224

synesthesiam:
Looks good to mean. I'd like to hear from the other language leaders, so let's give it a little time. This fix needs to be in the intents bump for the next release though.

ViViDboarder (approved):

ViViDboarder:
Also here to chime in that English looks good. Something worth flagging (but I don’t think we can address yet) is that the way localization of times work is not necessarily directly tied to language, though I think it’s often enough. 

It seems to me like it would be better for Home Assistant, which knows the localization format of the user, to pass a localized string back to the assistant rather than unlocalized numbers.

balloob:

---

PR #2806, opened 2024-12-30
[DA] chnage wording for timer in Danish in Update _common.yaml

kimvonmullen:
"Timer" is not pronounce correctly in Danish, it should be "nedtælling" as the default.


MTrab (approved):
Looks like a good solution to me.
The pronunciation was never good in Dansih, using timer.

PR #2807, opened 2024-12-30
[DA] added cancel all timer

kimvonmullen:


MTrab (changes requested):
Looks good to me, but missing the tests

PR #2810, opened 2024-12-30
[FR] Set temperature with area context

galaad:
Split sentences due to following limitation see [here](https://community.home-assistant.io/t/set-default-list-slot-value-in-intent-or-intent-script/761568/2) and add area context.


jlpouffier (approved):
I approve the changes, however please be aware that `HassClimateSetTemperature` is an intent that does not yet exists.

So the sentences are there, we (Language leaders and contributors) are getting ready to support it, but it's not possible to set a temperature by voice for now.

However, the changes are approved!

Thx

PR #2815, opened 2024-12-31
[CA] Add alternatives for HassGetCurrentDate

jmontane:
Add variants


duhow (changes requested):
Pots afegir exemples de les noves frases al test?

duhow (approved):

PR #2821, opened 2025-01-02
Enable support for `HassClimateSetTemperature` intent

mib1185:
- ~~needs: https://github.com/home-assistant/core/pull/134316~~
- needs: https://github.com/home-assistant/core/pull/136484

At the moment the action `climate.set_temperature` does not support the `temperature_unit` argument -> an [architectural discussion](https://github.com/home-assistant/architecture/discussions/1177) has been started, but it most properly results in a more broader and generic solution, than just add the argument to this single action call.

So even the intent supports the `temperature_unit` argument, the underlying action call doesn't.
Can we keep `temperature_unit` argument in the existing sentences or should we remove it for now?

### DIFF
--- a/intents.yaml
+++ b/intents.yaml
@@ -232,7 +232,7 @@ HassLightSet:
 # -----------------------------------------------------------------------------
 
 HassClimateSetTemperature:
-  supported: false
+  supported: true
   domain: climate
   description: "Sets the desired temperature of a climate device or entity"
   slots:


tetele:
I think this is [discussion](https://github.com/home-assistant/intents/discussions) material, but here's my 2c until you open one: we should not remove the optional temperature unit, but we should add context to the intent. Something like:

```yaml
# sentences/en/_common.yaml
expansion_rules:
  degrees_c: ([ ]°[[ ](c[elsius])])|( degrees c[elsius])
  degrees_c: ([ ]°[[ ](f[ahrenheit])])|( degrees f[ahrenheit])
  temperature_c: "{temperature}[<degrees_c>]"
  temperature_f: "{temperature}[<degrees_f>]"
```

```yaml
# sentences/en/climate_HassClimateSetTemperature.yaml
intents:
  HassClimateSetTemperature:
    data:
      - sentences:
          - "(<numeric_value_set>|adjust) <temp> [in <area>] to <temperature_c>"
          - "(<numeric_value_set>|adjust) <area> <temp> to <temperature_c>"
        requires_context:
          temperature_unit: "celsius"
      - sentences:
          - "(<numeric_value_set>|adjust) <temp> [in <area>] to <temperature_f>"
          - "(<numeric_value_set>|adjust) <area> <temp> to <temperature_f>"
        requires_context:
          temperature_unit: "fahrenheit"
```

Then the conversation agent would include the system temperature unit in the context.

PR #2824, opened 2025-01-03
[EN] Remove "i want" as skip word

### LINKED ISSUE #2795: [en] skip word "i want" limitation when creating custom sentences
When creating custom sentences, there is no indication that a skip word is preventing the custom sentence from matching.  Example provided below.

custom_sentences/en/WatchTV.yaml:
```yaml
language: en
intents:
  WatchTv:
    data:
      - sentences:
          - i want to watch tv                # this doesn't work
          - (i'm | i am) going to watch tv    # this does work
```

Debug showing that the sentence is identical but does not match.
image

Would be helpful if any or all of the following could be done:

- provide an option to override/ignore skip words from intents/sentences/xx/_common.yaml 
- show additional information in debug window when the text is altered by the following for example
    - a skip word is used/triggered/ignored
    - a conversion is performed due to a list in/out value

tetele:
Fixes #2795 

More context https://discord.com/channels/330944238910963714/1322275840150470758/1322275840150470758

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -426,4 +426,3 @@ skip_words:
   - "for me"
   - "i'd like"
   - "i'd like to"
-  - "i want"
--- a/tests/en/light_HassTurnOff.yaml
+++ b/tests/en/light_HassTurnOff.yaml
@@ -85,7 +85,6 @@ tests:
       - deactivate all the lights here
       - deactivate the lights
       - deactivate the lights in this room
-      - i want all the lights off in this room
       - lights off
       - lights off in here please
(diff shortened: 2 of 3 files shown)


ViViDboarder:
I’m still not fully convinced this should be removed. I think maybe changing it to “I want to” would make more sense though. 

The takeaway from that Discord thread for me is that there should be some indication when words are removed in the debug logs. This could just have easily happened with “please” or any other word or phrase used and is not unique to “I want”.

synesthesiam:
This is being fixed upstream: https://github.com/OHF-Voice/hassil/pull/275

PR #2836, opened 2025-01-05
Fix incorrect French responses in HassGetWeather.yaml

willdurand:
These two sentences are not quite correct in French, there should only be one "pas".


jlpouffier (approved):

jlpouffier:
@willdurand Fix the tests and you'll be good to go ;) 
Check the CI results

jlpouffier (approved):

PR #2841, opened 2025-01-06
[SV] Added `binary_sensor_HassGetState.yaml` + "som" as skip word

slungs:


sentences/sv/_common.yaml
```diff
@@ -349,7 +349,11 @@ expansion_rules:
   area: "{area}[e|me][t|n][s]"
   vad: "(vad är | vad är det | vad är det för | vilken)"
   var: "(var | var är | vart | vart är)"
+  vilka: "(vilka | vilken)"
+  någon: "(någon | något | några)"
+  ingen: "(ingen | inga | inget)"
```
larsdunemark, line 354: Should we use this expansion_rules for something? Feels like it could be used in the new sentences.

slungs: I added in a sentance for vilka and removed the other two, will add in another pr

tests/sv/binary_sensor_HassGetState.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  - sentences:
+      - "Vilka batterier är låga?"
+    intent:
+      name: HassGetState
+      slots:
+        domain: binary_sensor
+        device_class: battery
+        state: "on"
+    response: "Inte någon"
```
slungs, line 53: Not sure how "genetiv" should be handled here. Could we do some logic within the scope of the `which` response?

slungs:
The test is failing for code I have not touched in this pr. It is skipped when run locally.

larsdunemark:
@slungs I think the problem could be that we added "som" as skip word and it is part of the testcase for timers. "som heter ...."

slungs:
> @slungs I think the problem could be that we added "som" as skip word and it is part of the testcase for timers. "som heter ...."

Yeah you are right. I removed it again and will add it back in another PR if used. I think I got it right now.

tests/sv/binary_sensor_HassGetState.yaml
larsdunemark: I guess it would be possible to have 2 version of the "which" respond depending to sentence.
It might also be that the "which" respond should be simplifyed to "Ingen/Inget"

larsdunemark (approved):

PR #2840, opened 2025-01-06
[DE] harmonize todo and shopping lists

andreasbrett:
SSIA - fixes https://github.com/home-assistant/intents/issues/2808

@mib1185 @easterapps could you please review this? don't want to rush this although tests and (my late night) logic behind the PR work out.

Issue #2808, opened 2024-12-30
[DE] add element to shoppinglist


jabdoa2:
"Füge XYZ zur Einkaufsliste hinzu" -> will add "XYZ zur" to the list. 
"Füge XYZ zum Einkauf hinzu" -> will add "XYZ zum" to the list. 

Interestingly this works:
"Füge XYZ zu der Einkaufsliste hinzu" -> will add "XYZ" to the list. 

![image](https://github.com/user-attachments/assets/89ad8485-296c-4842-9786-c0acaf01da7d)
![image](https://github.com/user-attachments/assets/75db9cd8-141e-4712-9188-4b02097ed5c2)

Probably related to:
```
          - "füge <item>[ zu] <meine_liste> hinzu"
          - "füge[ zu] <meine_liste> <item> hinzu"
```

Maybe better would be:
```
          - "füge <item>[ zu| zur| zum] <meine_liste> hinzu"
          - "füge[ zu| zur| zum] <meine_liste> <item> hinzu"
```

Not sure if those pronoms should be supported more generically? There seem to be some specifics here: https://github.com/home-assistant/intents/blob/main/sentences/de/_common.yaml#L554. 

```
  zu: "(zu|[he]runter|nach unten)"
  von_dem: "(von[ dem]|vom)"
  zu_dem: "(zu[ dem]|zum)"
```

PR #2842, opened 2025-01-06
[DE] Add more german shopping list examples

jabdoa2:
some more examples for #2808 and  #2840

### DIFF
--- a/sentences/de/shopping_list_HassShoppingListAddItem.yaml
+++ b/sentences/de/shopping_list_HassShoppingListAddItem.yaml
@@ -3,14 +3,18 @@ intents:
   HassShoppingListAddItem:
     data:
       - sentences:
-          - "füge <item>[ (zu|zur|zum)] <meine_liste_dativ> hinzu"
+          - "füge <item>[ (zu|zur)] <meine_liste_dativ> hinzu"
           - "füge[ (zu|zur|zum)] <meine_liste_dativ> <item> hinzu"
           - "(setz[e]|schreib[e]|nehme|nimm) <item> (auf|in) <meine_liste_akkusativ>"
           - "<item> (auf|in) <meine_liste_akkusativ>[ (setzen|schreiben|nehmen)]"
-          - "<item>[ (zu|zur|zum)] <meine_liste_dativ> hinzufügen"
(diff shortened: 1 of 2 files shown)


jabdoa2:
I am not sure why the build failed. The Intent test seems to have passed:
```
 [gw0] [ 43%] PASSED tests/test_language_intents.py::test_shopping_list_HassShoppingListAddItem[de]

mib1185:
> AssertionError: Recognition failed for 'füge dem Einkauf Milch hinzu'

Didn't check the intent sentence right now, but this test sentence seems not to be covered by the intents

andreasbrett:
I don't think we should add "dem Einkauf" as fixed name for this intent. You can't add a shopping item to "dem Einkauf" but only to your shopping list. I might be nitpicking here but adding something to your Einkauf is more like adding something to your shopping cart, not your shopping list. If you want to use "dem Einkauf" this is already possible by naming your shopping list "Einkauf" or adding an alias.

jabdoa2:
> I don't think we should add "dem Einkauf" as fixed name for this intent. You can't add a shopping item to "dem Einkauf" but only to your shopping list. I might be nitpicking here but adding something to your Einkauf is more like adding something to your shopping cart, not your shopping list. If you want to use "dem Einkauf" this is already possible by naming your shopping list "Einkauf" or adding an alias.

That is how other members of my household try to use it. I know that you probably would not get any points in your German grammar classes for this buts its a more of a practical intent. Other assistants (Alexa and co) would understand that out of the box. At least here adding stuff to the shopping list and timers are the main usecases for voice.

andreasbrett:
Well I don't want to block this rigorously. I'm not certain it will be used by a lot of people but who knows.

Please do fix the sentences though. You haven't defined a sentence for your test "Milch dem Einkauf hinzufügen" yet which is why the tests are failing.

andreasbrett (approved):
I fixed a minor issue* and re-arranged the sentences to silence my inner monk. Tests should run fine now and I hereby approve.

* when defining multiple options these are in(optionA|optionB). if you want to make that optional you use [(optionA|optionB)]. this here won't work: [optionA|optionB].

PR #2861, opened 2025-01-11
[SV] Translate percent into Swedish

zyberzero:
During discussions on a Swedish home automation discord we figured out that controlling light levels didn't work and it turned out to be that the word for percent wasn't translated into Swedish.


danielholm:
Sorry to steal the PR but what Discord is this? I would like to join. Unable to find it, however.

larsdunemark (approved):

PR #2865, opened 2025-01-12
[EL] Update _common

apo-mak:


apo-mak:
Can we merge this ? 
@chatziko 
@frenck

PR #2872, opened 2025-01-15
[DE] Add some shorter sentences for "set timer"

AlCalzone:
With Alexa and Siri, I'm able to set a timer just by saying "Timer X Minuten". This PR adds this possibility along with some variations for named timers.


andreasbrett:
Great addition. Should help those that migrate from Alexa to Assist. Thanks for the contribution!

PR #2873, opened 2025-01-15
[DE] Add some sentences for timer status

AlCalzone:
I used to be able to query timers in Alexa using "Was macht mein/der Timer?"
This PR adds some similar sentences to make this kind of querying possible.


easterapps (approved):

andreasbrett:
Thank you! Again very clean PR that leaves no room for improvement.

KonstBr:
Is it also possible to add "was **sagt** der Timer"?

PR #2876, opened 2025-01-16
[EN] Remove say from English

synesthesiam:
Remove "say {message}" from English sentences. This interacts poorly with skip words and LLMs, for example "can you say that again" will respond with "that again" instead of being processed by the LLM.


ViViDboarder (approved):

tetele:
> "can you say that again" will respond with "that again"

Makes for proper dad jokes, though 😅

tetele (approved):

PR #2877, opened 2025-01-16
[EN] Add HassBroadcast

synesthesiam:
Adds support for `HassBroadcast` intent: https://github.com/home-assistant/core/pull/135337

This intent announces a message on every other Assist satellite (or all satellites if used through the web UI).

### DIFF
--- a/sentences/en/assist_satellite_HassBroadcast.yaml
+++ b/sentences/en/assist_satellite_HassBroadcast.yaml
@@ -1,0 +1,6 @@
+language: "en"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "broadcast [that] {message}"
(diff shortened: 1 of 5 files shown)


ViViDboarder (approved):
It would be great if we could target broadcasts to areas, but that is another core change.

sentences/en/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,6 @@
+language: "en"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "broadcast [that] {message}"
```
tetele, line 6:
```suggestion
          - "(broadcast|announce) [everywhere] [that] {message}"
```

PR #2882, opened 2025-01-17
[DE] Add HassBroadcast

andreasbrett:
DE version of https://github.com/home-assistant/intents/pull/2877

@mib1185 @easterapps this needs fine tuning. suggestions very welcome. "verkünde, dass das abendessen fertig ist" would need to require a transformed response "das abendessen ist fertig" to be perfectly correct. unfortunately it's not as easy in DE as it is in EN. I could not find phrases that are as smooth as in EN.


responses/de/HassBroadcast.yaml
```diff
@@ -0,0 +1,6 @@
+---
```
mib1185, line 1:
```suggestion
```

sentences/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,8 @@
+language: "de"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören)) [dass ]{message}"
+          - "verkünde [überall ][dass ]{message}"
+          - "sag[e] überall [dass ]{message}"
```
mib1185, line 8:
```suggestion
          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören))[, dass] {message}"
          - "verkünde[ überall][, dass] {message}"
          - "sag[e] überall[, dass] {message}"
          - "gib[ überall] Bescheid[, dass] {message}"
```

tests/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - "gib bekannt das abendessen ist fertig"
+      - "gib durch das abendessen ist fertig"
+      - "gebe bekannt das abendessen ist fertig"
+      - "gebe durch das abendessen ist fertig"
+      - "lass alle wissen das abendessen ist fertig"
+      - "lass alle hören das abendessen ist fertig"
+      - "verkünde das abendessen ist fertig"
+      - "verkünde überall das abendessen ist fertig"
+      - "sag überall das abendessen ist fertig"
+      - "sage überall das abendessen ist fertig"
```
mib1185, line 13:
```suggestion
      - "sage überall das abendessen ist fertig"
      - "sage überall, dass das Abendessen fertig ist"
      - "verkünde überall, dass das Abendessen fertig ist"
      - "gib Bescheid, dass das Abendessen fertig ist"
      - "gib überall Bescheid, dass das Abendessen fertig ist"
```

mib1185 (changes requested):
what do you think about these ones?

sentences/de/assist_satellite_HassBroadcast.yaml
andreasbrett: I like the "gib überall Bescheid" addition but why add the comma? isn't punctuation from STT stripped before being handed over to intent matching?

mib1185: > isn't punctuation from STT stripped before being handed over to intent matching?

mhhh ... good question :thinking: not sure :see_no_evil:

andreasbrett: just verified that it is indeed stripped. then again one could argue that's why it's okay to have the punctuation in there :) 

image

mib1185:
>  Sentences should only contain words and matching syntax. They should not contain punctuation. 

you're right, we should not add the punctuation :grimacing:

sentences/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,9 @@
+language: "de"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören))[, dass] {message}"
+          - "verkünde[ überall][, dass] {message}"
+          - "sag[e] überall[, dass] {message}"
+          - "gib[ überall] Bescheid[, dass] {message}"
```
mib1185, line 9:
```suggestion
          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören)) [dass ]{message}"
          - "verkünde[ überall] [dass ]{message}"
          - "sag[e] überall [Bescheid ][dass ]{message}"
          - "gib[ überall] Bescheid [dass ]{message}"
```

tests/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,22 @@
...
```
mib1185, line 13:
```suggestion
      - "sage überall das abendessen ist fertig"
      - "sage überall Bescheid das abendessen ist fertig"
```

tests/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - "sage überall Bescheid das abendessen ist fertig"
+      - "sage überall, dass das Abendessen fertig ist"
+      - "verkünde überall, dass das Abendessen fertig ist"
+      - "gib Bescheid, dass das Abendessen fertig ist"
+      - "gib überall Bescheid, dass das Abendessen fertig ist"
```
andreasbrett, line 18: these will produce a different response: "das Abendessen fertig ist" instead of "das Abendessen ist fertig". that's the transformation issue I mentioned. I see no way to work around that unless we want to introduce a long list of verbs and conjugations of "sein", "haben", "werden"....

mib1185: ahhh ... with response you mean the broadcasted message?

andreasbrett: Sorry, yes. That was unnecessarily ambiguous

mib1185: than we should remove the `dass` at all - so only allow things like "gib überall Bescheid das Abendessen ist fertig"

andreasbrett: This issue must be the same for at least some other languages (IIRC dutch should have the same issue). If we don't come up with a solution on our own, we can improve later on, when HassBroadcast is implemented in those languages.

andreasbrett: > than we should remove the `dass` at all - so only allow things like "gib überall Bescheid das Abendessen ist fertig"

but then asking "gib überall bescheid dass das abendessen fertig ist" will result in a broadcast "dass das abendessen fertig ist". we can only chose pest ("das abendessen fertig ist") or cholera ("dass das abendessen fertig ist") I guess. both sound weird.

andreasbrett: tbh, I don't think the original approach in EN to simply speak out {message} is that good. I'd rather have a broadcast like "Hiermit wird verkündet: {message}".

mib1185: > tbh, I don't think the original approach in EN to simply speak out {message} is that good. I'd rather have a broadcast like "Hiermit wird verkündet: {message}".

this would need a change i the core I think. So for now i would just remove the `dass` and all related test sentences. The user will recognize, that the message is replayed as is, there is no transformation or similar in between.

andreasbrett: yes, also a chime before the announcement like in https://github.com/nimroddolev/chime_tts would be great

while the message is simply repeated, I agree with your idea of removing `dass` as it's the less weird behaviour.

mib1185: > yes, also a chime before the announcement like in https://github.com/nimroddolev/chime_tts would be great

even this is a change in the core, and maybe also in the firmware of the VPE itself :thinking: but nothing for this PR for now :wink:

sentences/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,9 @@
+language: "de"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören)) {message}"
+          - "verkünde[ überall] {message}"
+          - "sag[e] überall[ Bescheid]{message}"
```
mib1185, line 8:
```suggestion
          - "sag[e] überall[ Bescheid] {message}"
```

tests/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - "gib durch das abendessen ist fertig"
+      - "gebe bekannt das abendessen ist fertig"
+      - "gebe durch das abendessen ist fertig"
+      - "lass alle wissen das abendessen ist fertig"
+      - "lass alle hören das abendessen ist fertig"
+      - "verkünde das abendessen ist fertig"
+      - "verkünde überall das abendessen ist fertig"
+      - "sag überall das abendessen ist fertig"
+      - "sage überall das abendessen ist fertig"
+      - "sage überall Bescheid das abendessen ist fertig"
```
mib1185, line 14:
```suggestion
      - "sage überall Bescheid das abendessen ist fertig"
      - "gib Bescheid das abendessen ist fertig"
      - "gib überall Bescheid das abendessen ist fertig"
```

mib1185 (changes requested):

sentences/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,9 @@
+language: "de"
+intents:
+  HassBroadcast:
+    data:
+      - sentences:
+          - "((gib|gebe) (bekannt|durch)|lass alle (wissen|hören)) {message}"
+          - "verkünde[ überall] {message}"
+          - "sag[e] überall[ Bescheid] {message}"
+          - "gib[ überall] Bescheid {message}"
```
mib1185, line 9:
```suggestion
          - "(gib|gebe)[ überall] Bescheid {message}"
```

tests/de/assist_satellite_HassBroadcast.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - "gebe durch das Abendessen ist fertig"
+      - "lass alle wissen das Abendessen ist fertig"
+      - "lass alle hören das Abendessen ist fertig"
+      - "verkünde das Abendessen ist fertig"
+      - "verkünde überall das Abendessen ist fertig"
+      - "sag überall das Abendessen ist fertig"
+      - "sage überall das Abendessen ist fertig"
+      - "sage überall Bescheid das Abendessen ist fertig"
+      - "gib Bescheid das Abendessen ist fertig"
+      - "gib überall Bescheid das Abendessen ist fertig"
```
mib1185, line 16:
```suggestion
      - "gib Bescheid das Abendessen ist fertig"
      - "gebe Bescheid das Abendessen ist fertig"
      - "gib überall Bescheid das Abendessen ist fertig"
      - "gebe überall Bescheid das Abendessen ist fertig"
```

mib1185 (changes requested):

mib1185 (approved):

PR #2893, opened 2025-01-20
[NL] Add assist_satellite_HassBroadcast

TheFes:
SSIA

### DIFF
--- a/responses/nl/HassBroadcast.yaml
+++ b/responses/nl/HassBroadcast.yaml
@@ -1,0 +1,5 @@
+language: en
+responses:
+  intents:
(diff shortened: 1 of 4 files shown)


responses/nl/HassBroadcast.yaml
```diff
@@ -0,0 +1,5 @@
+language: nl
+responses:
+  intents:
+    HassBroadcast:
+      default: "Klaar"
```
frenck, line 5: Wouldn't "Gedaan" be better in this case?

TheFes: Gedaan

PR #2901, opened 2025-01-22
[NL] Fix HassGetWeather

TheFes:
After getting some reports that HassGetWeather didn't return the current state, I did some testing on my local instance.

Apparantly ` state.state`  already returns the translated weather state for the weather entity, so the template using the mapping didn't return anything (as it expects the original untranslated entity state).

This way it will return a weather condition, and not only the temperature.

### DIFF
--- a/responses/nl/HassGetWeather.yaml
+++ b/responses/nl/HassGetWeather.yaml
@@ -3,22 +3,4 @@ responses:
   intents:
     HassGetWeather:
       default: >
-        {% set weather_condition = {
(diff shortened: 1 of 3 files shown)


frenck (approved):
Thanks, @TheFes 👍 

../Frenck

TheFes:
@andreasbrett so, if I understand correctly, I should revert this change?

andreasbrett:
Yes, the linked PR will bring us untranslated states finally.

PR #2906, opened 2025-01-23
[DA] Add HassVacuum support

kbhsn4:
Add HassVacuumStart/ReturnToBase support in danish

### DIFF
--- a/sentences/da/_common.yaml
+++ b/sentences/da/_common.yaml
@@ -347,7 +347,7 @@ lists:
 expansion_rules:
   al: "(alle | alt | samtlige)"
   blæser: "blæser[(e|en|ne)]"
-  etage: "{area}[(n)]"
+  etage: "{floor}[(n|en)]"
   garagedør: "garage(dør|port)[(en|e|ene)]"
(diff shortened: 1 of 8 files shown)


sentences/da/_common.yaml
```diff
@@ -347,7 +347,7 @@ lists:
 expansion_rules:
   al: "(alle | alt | samtlige)"
   blæser: "blæser[(e|en|ne)]"
-  etage: "{area}[(n)]"
+  etage: "{floor}[(n|en)]"
```
MTrab, line 350: On purpose this is still english?

MTrab (changes requested):

sentences/da/_common.yaml
kbhsn4: Yes. It's in curly brackets, meaning it's a Slots List - in this case it used for matching against the internal Home Assistant defined list of floors. It's not for matching directly against user supplied input.

kbhsn4: Do you require anything else from my side?

MTrab (approved):

PR #2910, opened 2025-01-24
[FR] weather state is now exposed raw and untranslated

### LINKED ISSUE #2899: [FR] Weather State not translated
@bors-ltd Hey 👋🏻

I am creating an issue to check if you want to fix this yourself, as you did a really great job with the weather intent in French (Credits where it's due ;) )

The weather state (On both our default and detailed sentences) does not seem to be translated.

Image

Image

A small mapping of all weather states and their corresponding French translation would be worth it.

The available states can be found here:
https://developers.home-assistant.io/docs/core/entity/weather/

Image

Important note:
Feel free to tell me if you do not have the time to fix this.
In any case, I will assign this ticket to myself in a few days and fix it in case you are unavailable.

Have a great day.
JLo

bors-ltd:
As it goes for every exposed entity of any domain from now on.

This concludes uncertainty about the content of "state.state", and will have the nice side effect of putting the development environment on par with the production behaviour.

Fixes #2899

### DIFF
--- a/responses/fr/HassGetWeather.yaml
+++ b/responses/fr/HassGetWeather.yaml
@@ -102,11 +102,11 @@ responses:
           'kn': 'nœuds',
         }.get(wind_speed_unit, wind_speed_unit) %}
 
-        {% if beaufort_scale and wind_bearing_text and wind_gust_speed is number %}
+        {% if beaufort_scale and wind_bearing_text and wind_gust_speed %}
           {{ beaufort_scale }} de {{ wind_speed | int }} {{ wind_speed_unit }} {{ wind_bearing_text }}, avec des rafales à {{ wind_gust_speed | int }} {{ wind_speed_unit }}.
         {% elif beaufort_scale and wind_bearing_text %}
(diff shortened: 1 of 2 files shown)


responses/fr/HassGetWeather.yaml
```diff
@@ -3,9 +3,24 @@ responses:
...
+          'partlycloudy': 'Éclaircies',
+          'pouring': 'Averses',
+          'rainy': 'Pluvieux',
+          'snowy': 'Neigeux',
+          'snowy-rainy': 'Neigeux avec pluie mêlée',
+          'sunny': 'Ensoleillé',
+          'windy': 'Venteux',
+          'windy-variant': 'Venteux et nuageux',
+        } %}
+        {{ weather_condition.get(state.state | string, "") }}
```
bors-ltd, line 23: I assumed the future behaviour already, so I don't "lower" like most weather intents do today.

jlpouffier:
Sorry for the delay man! This completely slipped my mind!!!

PR #2941, opened 2025-01-31
[DA] Add intents for GetState for Person domain

kbhsn4:
Is there a better way to match and translate certain states but leave unmatched states untranslated?
```
"hjemme" -> "home"
"ikke hjemme" -> "not_home"
"arbejde" -> "arbejde"
"skole" -> "skole"
```
"home"/"not_home" is a special case related to the default home zone, which is in English internally. Other user defined zones just have the name the user have given (in their local language)


MTrab (approved):
I think the published solution is the best - I cannot see how it can be done different

kbhsn4:
> I think the published solution is the best - I cannot see how it can be done different

Thank you for looking into it @MTrab 👍

MTrab:
And thank you for your numorous contributions :)

PR #2955, opened 2025-02-03
[SV] Fix #2944 - Response is None 

### LINKED ISSUE #2944: Response entity is None
Image

On HA 2025.2.b2. 
Swedish language. 

Using Voice assistant in HA to, for example, turn a light on, the response is missing entity name and simply response with

Turned on None

However turn off, shows the name

danielholm:
Excluded light domain from default and added separate response for light. I believe this could fix #2944


sentences/sv/homeassistant_HassTurnOff.yaml
```diff
@@ -3,6 +3,17 @@ intents:
...
+            - light
+            - cover
+      - sentences:
+          - <stäng_gardiner> <name> #close curtains
+        requires_context:
+          domain:
+            - cover
         response: cover
+      - sentences:
+          - <släck> <name> # lights off
```
tetele, line 18: if this is just for `light`, then it should `requires_context: {domain: light}`

sentences/sv/homeassistant_HassTurnOn.yaml
```diff
@@ -8,6 +8,11 @@ intents:
           domain:
             - vacuum
             - script
+            - light
+            - cover
       - sentences:
           - <öppna_gardiner> <name>
         response: cover
+      - sentences:
+          - <tänd> <name>
```
tetele, line 17: if this is just for light, then it should `requires_context: {domain: light}`

sentences/sv/cover_HassTurnOff.yaml
```diff
@@ -4,7 +4,7 @@ intents:
     data:
       - sentences:
           - (regla | <stäng_gardiner>) garagedörren
-        response: cover_device_class
+        response: cover
```
larsdunemark, line 7: This should still response `Öppande garaget` (depending on the state of translation it might give the device_class in that case we might need to handle the translation also.

responses/sv/HassTurnOn.yaml
```diff
@@ -22,16 +22,16 @@ responses:
...
+      light: "Tände {{ slots.name }}"
       lights_area: "Tände lamporna i {{ slots.area }}"
       lights_all: "Tände alla lamporna"
       lights_floor: "Tände lamporna på {{ slots.floor }}"
       fans_area: "Startade fläktarna i {{ slots.area }}"
       fans_all: "Slog på alla fläktar"
-      cover: "Öppnade {{ slots.name }}"
+      cover: "Öppnade"
       cover_area: "Öppnade i {{ slots.area }}"
-      cover_device_class: "Öppnade {{ slots.device_class }}"
```
larsdunemark, line 33: This is not the correct change, we don't want the response for a device_class to not include the class.
The intention it that this will say
`Öppnande garaget` if a command was issued for a device that will open that device_class.

sentences/sv/cover_HassTurnOn.yaml
```diff
@@ -4,7 +4,7 @@ intents:
     data:
       - sentences:
           - <öppna_gardiner> garagedörren
-        response: cover_device_class
+        response: cover
```
larsdunemark, line 7: Same here.

sentences/sv/homeassistant_HassTurnOff.yaml
```diff
@@ -3,6 +3,20 @@ intents:
...
+          - <slå_av> <name> # turn off
+        excludes_context:
+          domain:
+            - light
+            - cover
+      - sentences:
+          - <stäng_gardiner> <name> #close curtains
+        requires_context:
+          domain:
+            - cover
```
larsdunemark, line 15: Should probably be moved to the cover_HassTurnOff file

sentences/sv/homeassistant_HassTurnOff.yaml
```diff
@@ -3,6 +3,20 @@ intents:
...
         response: cover
+      - sentences:
+          - <slå_av> <name> # lights off
+        requires_context:
+          domain:
+            - light
+        response: light
```
larsdunemark, line 22: Moving to light_HassTurnOff file.

This will have the same sentence as the first but with a respond.

larsdunemark (changes requested):
I think that you need to base your commits on the upstream main branch to avoid the problem where you get this merge problem with the README.md, in this PR it would remove the README.md from the main project if merged.

Please also try to make the changes smaller to avoid changing multiple domains as once if possible. This change both the way we handles lights and covers.

sentences/sv/homeassistant_HassTurnOff.yaml
```diff
@@ -3,6 +3,20 @@ intents:
...
+            - light
+            - cover
+      - sentences:
+          - <stäng_gardiner> <name> #close curtains
+        requires_context:
+          domain:
+            - cover
         response: cover
+      - sentences:
+          - <slå_av> <name> # lights off
```
larsdunemark, line 18: I think that one issue can be that the expansion rule for `<slå_av>` is a bit to wide, in contains both the work turn off, stopp, shutoff etc in swedish and for lights it might make more since to split it up in multiple sentences.
like: 
"släck <name>"
looking at the DA language.

balloob:

PR #2956, opened 2025-02-03
[NL] Add context sentences for HassClimateGetTemperature

synesthesiam:
Added context sentences for getting the current temperature from [wyoming-rhasspy-speech](https://github.com/rhasspy/wyoming-rhasspy-speech/blob/master/wyoming_rhasspy_speech/sentences/nl.yaml#L374)


TheFes:
Thanks, will have a second look myself :)

PR #2984, opened 2025-02-12
[DE] more options to cancel a timer

Dr4kn:
The Imperative of stopp and lösch has two viable options, with and without an e.
 
The new word option lösch(e) isn't already used in any other common option, and is German for delete. So there shouldn't be any conflicts created by this change. 

It is also a variant to cancel a timer that is allowed on Google Home and possibly Alexa (I don't have one). Allowing it makes switching away from them easier, because you do not have to learn a now command word for the same action.

### DIFF
--- a/sentences/de/_common.yaml
+++ b/sentences/de/_common.yaml
@@ -598,7 +598,7 @@ expansion_rules:
 
   # Timers
   timer_set: "(starte|setze|stelle|erstelle)"
-  timer_cancel: "(beende|stoppe)"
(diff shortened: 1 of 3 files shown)


andreasbrett:
Could you please also provide some tests for these additions? You'll find them in the tests folder.

andreasbrett (changes requested):
Could you please also provide some tests for these additions? You'll find them in the tests folder.

mib1185 (approved):
LGTM, thx @Dr4kn :+1:

PR #3002, opened 2025-02-19
[HR] Start timer: fix logic for count and cases

mdimjasevic:
The logic for determining the case was simplistic, hence incorrect for example when a timer is set to 3 minutes. This PR fixes it and reuses the logic when determining the case for hours and seconds as well.


mdimjasevic:
This took more time than I was expecting, but it's ready. I've added a bunch of tests.

PR #3003, opened 2025-02-19
Allow storing TTS and STT support

balloob:
The goal is to store all data that backs the widget at https://www.home-assistant.io/voice_control/#supported-languages-and-sentences in this repository. The first step is to allow documenting the support for speech-to-phrase, whisper and cloud for each language.

Since we note that support based on region, we organize support by region.

### DIFF
--- a/languages.yaml
+++ b/languages.yaml
@@ -46,6 +46,14 @@ en:
     - mrdarrengriffin
     - tetele
     - vividboarder
+  support:
+    US:
+      speech-to-text:
+        speech-to-phrase: true
+        whisper-wer: 8.9
+      text-to-speech:
+        piper: true
+        cloud: true
 es:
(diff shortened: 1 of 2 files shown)


synesthesiam (changes requested):
Let's change `whisper-wer` to a boolean that indicates if we think the recommended model is good enough for users.

tetele:
What do you mean by "recommended model"?

balloob:
We expect to offer fine tuned Whisper versions for different languages in the future.

synesthesiam:
> What do you mean by "recommended model"?

For x86_64 systems, we recommend the `base` model and for ARM-based systems we recommend `tiny`.

synesthesiam (approved):

PR #3004, opened 2025-02-19
feat: Add HassListCompleteItem

michael-genson:
Adds ~~"remove"~~ "complete" intents for todo lists. See core PR:
- https://github.com/home-assistant/core/pull/127806

Previously this also had the shopping list complete item intents, but I've moved those to [another PR](https://github.com/home-assistant/intents/pull/3035)


tetele:
Thank you, but such new functionality requires a corresponding PR in core. Do you know of any such PR and can you link it here? It's great to see you have marked the intents as not supported, at least

michael-genson:
Ah yep, makes sense. I wasn't sure if I needed to get this in first (as not supported) then get it in core, or vice versa, or simultaneously. I can get a PR open there in the next day or two and link it here.

tetele:
Sounds great! Please cross link the PRs once you do.

michael-genson:
Turns out someone is already working on this in core:
- https://github.com/home-assistant/core/pull/128565
- https://github.com/home-assistant/core/pull/127806

I've renamed to "complete" (rather than remove) since that's the verbiage in the core PR (and it works better than remove anyway since other internals use the same terminology)

Lesekater:
> Turns out someone is already working on this in core:
> 
> * [Add complete intent function for shopping list component core#128565](https://github.com/home-assistant/core/pull/128565)
> * [Add complete item intent function for todo component core#127806](https://github.com/home-assistant/core/pull/127806)
> 
> I've renamed to "complete" (rather than remove) since that's the verbiage in the core PR (and it works better than remove anyway since other internals use the same terminology)

FIY - https://github.com/home-assistant/core/pull/127806 just got merged. You might wanna split up this PR, so HassListCompleteItem can get merged without having to wait on https://github.com/home-assistant/core/pull/128565.

michael-genson:
Thanks! Will do

synesthesiam:
Can we close this PR then since the shopping list has been split out?

michael-genson:
No they are both required. This PR adds intents for the todo list, while the other PR adds intents for the shopping list. Both are required, but since core was two separate PRs I split this into two as well (so we can get this one merged before the shopping list core PR is merged)

sentences/en/todo_HassListCompleteItem.yaml
```diff
@@ -0,0 +1,12 @@
+language: "en"
+intents:
+  HassListCompleteItem:
+    data:
+      - sentences:
+          - (check|complete|delete|remove) [off] <item> [off] (off|from|on|in) <my_list>
```
tetele, line 6: Same comment as for shopping lists: This would match "check off item off off my todo list". I'd split it in multiple sentences to make it readable.

- `(check|...) (off;<item>) (from|on|in) <my_list>`
- `(check|...) <item> (off|from|on|in) <my_list>`

sentences/en/todo_HassListCompleteItem.yaml
```diff
@@ -0,0 +1,12 @@
...
+        response: item_completed
+        requires_context:
+          domain: todo
+        expansion_rules:
+          my_list: "[my|the] {name} [list]"
+          item: "{shopping_list_item:item}"
```
tetele, line 12: This is not a shopping list. Names should not refer to it as such.

tests/en/todo_HassListCompleteItem.yaml
```diff
@@ -0,0 +1,15 @@
+language: en
+tests:
+  - sentences:
+      - check off apples from my trader joes list
```
tetele, line 4: These examples are better suited for shopping lists, not TODO lists. Let's use more relevant tests please.

intents.yaml
```diff
@@ -568,6 +568,25 @@ HassListAddItem:
...
+      required: true
+    name:
+      description: "Name of the list"
+      required: true
+  slot_combinations:
+    - slots:
+      - "item"
+      - "name"
+      importance: "complete"
+      example: "check off apples from my shopping list"
```
tetele, line 588: This example is for shopping lists. Let's add one which is more suited for TODO lists.

tetele (changes requested):

tests/en/todo_HassListCompleteItem.yaml
michael-genson: Sure thing, did you want me to update the existing "add item" examples as well? These were copied from there (but changed to "check off" etc

sentences/en/todo_HassListCompleteItem.yaml
michael-genson: This would miss "Check item off from my todo list" but I can add that to the second sentence pattern

tests/en/todo_HassListCompleteItem.yaml
tetele: Oh, really? Those should be changed too, but not in this PR. Thanks for the heads up!

michael-genson: No problem! If I find time I can open a PR to update those as well. But I'll get the new ones changed for this PR

sentences/en/todo_HassListCompleteItem.yaml
michael-genson: Ahh you're right, my mistake, thanks!

michael-genson:
As discussed in a review comment, currently `HassListAddItem` uses shopping list verbiage (examples, expansion rules, and tests) so those should be updated in a separate PR. For this PR I've updated the appropriate verbiage for `HassListCompleteItem`.

tetele (approved):
@ViViDboarder @mrdarrengriffin please confirm

tetele:
@michael-genson run `script/lint` locally and push the changes it suggests

mrdarrengriffin (approved):

PR #3006, opened 2025-02-21
[CS] Add intents for Date and Time

adamjez:
Add built-in intents for `HassGetCurrentDate` and `HassGetCurrentTime`.

### DIFF
--- a/sentences/cs/homeassistant_HassGetCurrentDate.yaml
+++ b/sentences/cs/homeassistant_HassGetCurrentDate.yaml
@@ -1,0 +1,8 @@
+language: cs
+intents:
+  HassGetCurrentDate:
+    data:
+      - sentences:
+          - "jaké je [dnešní|dnes] datum"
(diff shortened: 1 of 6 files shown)


sentences/cs/homeassistant_HassGetCurrentDate.yaml
```diff
@@ -0,0 +1,8 @@
+language: cs
+intents:
+  HassGetCurrentDate:
+    data:
+      - sentences:
+          - "jaké je [dnešní|dnes] datum"
+          - "kolikátého (je;[dnes])"
```
halecivo, line 7: is semicolon here intentional?

adamjez: Yes, it is here to support both sentences:
* "kolikátého dnes je"
* "kolikátého je dnes"

halecivo (approved):

PR #3014, opened 2025-02-25
[EN] Match articles "a" and "an" for <the>

tannisroot:
Whisper (for me) seems to always put article `a` after vacuum commands `start` or `return`.
Someone might also actually say it this way so let's handle that.

### DIFF
--- a/sentences/en/_common.yaml
+++ b/sentences/en/_common.yaml
@@ -357,7 +357,7 @@ lists:
     wildcard: true
 
 expansion_rules:
-  the: "(the|my|our)"
+  the: "(a[n]|the|my|our)"
   name: "[<the>] {name}"
   area: "[<the>] {area}"
   floor: "[<the>] {floor} [floor]"


tetele:
First of all, I don't think it's a good idea to prefix `<name>` (which can contain `<the>`) with an indefinite article. You would be able to say `start a the roborock`, which doesn't make grammatical sense.

Second of all, if we're doing this, I don't see why there wouldn't be the `an` form as well in there.

Third, I don't see why this would apply strictly to vacuums and not every other entity.

Fourth, to counter the the issues above and create new ones, why not add `a[n]` to [`<the>`](https://github.com/home-assistant/intents/blob/f70372ebc4adaf045ee842147ede2d73ffd9b870/sentences/en/_common.yaml#L360)?

Finally, like I said numerous times before, I don't think it's wise to add incorrect sentences just to please Whisper or any other STT. The proper solution here would be to fix Whisper.

I'd like to hear the other language leaders' comments on this.

tannisroot:
> First of all, I don't think it's a good idea to prefix `<name>` (which can contain `<the>`) with an indefinite article. You would be able to say `start a the roborock`, which doesn't make grammatical sense.
> 
> Second of all, if we're doing this, I don't see why there wouldn't be the `an` form as well in there.
> 
> Third, I don't see why this would apply strictly to vacuums and not every other entity.
> 
> Fourth, to counter the the issues above and create new ones, why not add `a[n]` to [`<the>`](https://github.com/home-assistant/intents/blob/f70372ebc4adaf045ee842147ede2d73ffd9b870/sentences/en/_common.yaml#L360)?
> 
> Finally, like I said numerous times before, I don't think it's wise to add incorrect sentences just to please Whisper or any other STT. The proper solution here would be to fix Whisper.
> 
> I'd like to hear the other language leaders' comments on this.

Oh I agree on most of this, and I can change the PR to have a[n] in <the> instead of just for the vacuum, I just thought it would an issue to have a change that would affect other commands. I would be more than happy to add the change directly to <the> if that's what would be preferred.
And yes in my case it is fixing Whisper, but since Whisper is, like LLMs, a statistical model, it means in the data it was trained on, people would often say it with "a", or at least say it in a way that would make it sound like an "a". I am certainly guilty of mushing the "the" in such sentences in a way that makes it sound almost like "a".
It's not grammatically correct, sure, but the point of intents is to understand all people, not just people with good grammar knowledge, and it seems innocent enough?

tannisroot:
Also, as much as I would love to fix Whisper to follow grammar, but there is so much you can do with it to influence the output. OpenAI probably trained it on mega powerful datacenters with all the speech data they sucked off the internet, it's not really realistic to be able to somehow fix all the edge cases like this without the expertise and resources they had.
And yes, there is `speech-to-phrase` as an alternative, but in my testing bigger Whisper models are far, far better at understanding noisy, imperfect audio that you typically get out of assist satellites, as well as telling apart "turn off" and "turn on", so I am afraid it is here to stay for those who need local STT.

tannisroot:
I went ahead and added "a[n]" directly to <the>, since adding it to the intents themselves leads to wonky matching.

ViViDboarder (changes requested):
Yea, I'm not sure I like adding `a[n]` here either. That implies multiple options that are being picked from. I'm personally not opposed to adding improper grammar that is actually something people use, but since this is more about coercing what Whisper transcribed, I'm not so sure.

I'd prefer we close this in favor of either improving the voice models or using phonetic matching, which might help match `a` to `the`.

synesthesiam:
@tannisroot Thanks for the idea, but I agree that this needs to be fixed either upstream (Whisper/STT) or downstream as a better fallback system. I'm working on the fallback part, so this will hopefully become a non-issue this year.

Issue #3016, opened 2025-02-25
[DE] "Rollo" / "Beschattung" mapped to two cover device classes


mithomas:
In https://github.com/home-assistant/intents/blob/f70372ebc4adaf045ee842147ede2d73ffd9b870/sentences/de/_common.yaml, "Rollo" and "Beschattung" (and their variations) are mapped to both the _blind_ and the _shutter_ device class:

https://github.com/home-assistant/intents/blob/f70372ebc4adaf045ee842147ede2d73ffd9b870/sentences/de/_common.yaml#L357-L376

Living in an attic flat with both roof windows with indoor blinds and regular windows with exterior shutters (both configured as "shutters") I'd like to be able to query "Welche Rollos sind geschlossen?" - currently this is answered as "None" where I would expect both to show up. 

(Arguably, mapping "Rollo" to _blind_ might actually be the right choice and I should refer to them as "Rollläden". In that case I'd really like know how to refer to the whole _cover_ domain for a given area, ignoring any _device_class_ setting...)

andreasbrett:
Thanks for bringing this up. I think we should translate correctly and make Rollos only blinds and not shutters as well. These are the official translations though I think in German Rollo and Rollladen is used in an interchangeable fashion.

To your question how to query all covers regardless of device class: I think this is best solved in a custom sentence as the cover domain is quite large and contains doors, windows etc pp.

andreasbrett:
> Another way could be to locally overwrite the mapping via a custom configuration - not sure if this is possible (or even if it _should be_), but I could well image that to be relevant for other things as well; language after all is not always unambiguous and it would be nice to resolve these ambiguities in the way I as a user see fit :-).

You can add to those maps/lists. I created the following file in `custom_sentences/de/_common.yaml` and could query my doors using `Schildkröte` (don't forget to restart HA or reload conversation in the HA dev tools).

```yaml
language: de
lists:
  cover_classes:
    values:
      - in: Schildkröte[n]
      - out: door
```

PR #3023, opened 2025-02-27
[DE] fix ambiguities for covers

andreasbrett:
Fixes some ambiguities for the `cover` device classes. As mentioned in https://github.com/home-assistant/intents/issues/3016, `Rollo` and `Beschattung` map to multiple device classes resulting in incorrect responses. Although these words are commonly used interchangeably, I propose to match those EXACTLY to their english variants. If we have precise device classes, then we should also expect users to provide those exact class names.

`Beschattung` -> `awning`
`Rollo` -> `blind`

Additionally, responses for device class `shade` were harmonized. If `Schirm` is used for querying this class, `Schirm` should be used for the corresponding responses.

### DIFF
--- a/sentences/de/_common.yaml
+++ b/sentences/de/_common.yaml
@@ -107,7 +107,7 @@ responses:
         "door": "keine Tür",
         "garage": "kein Garagentor",
         "gate": "kein Tor",
-        "shade": "kein Rollo",
+        "shade": "kein Schirm",
         "shutter": "kein Rollladen",
         "window": "kein Fenster"
         } %}
@@ -124,7 +124,7 @@ responses:
         "door": "keine Tür",
         "garage": "kein Garagentor",
         "gate": "kein Tor",
-        "shade": "kein Rollo",
+        "shade": "kein Schirm",
         "shutter": "kein Rollladen",
         "window": "kein Fenster"
         } %}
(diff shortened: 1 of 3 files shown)


andreasbrett:
adding @ascha191 to this discussion since `Beschattungen` was introduced by him in https://github.com/home-assistant/intents/pull/2463

andreasbrett:
> > `Rollo` and `Beschattung` map to multiple device classes resulting in incorrect responses
> 
> We should also check the `device_class` in the tests to avoid such things in future :thinking: 

Yes. Let me modify the tests and add some more fixtures to have more meaningful tests.

andreasbrett:
> > `Rollo` and `Beschattung` map to multiple device classes resulting in incorrect responses
> 
> We should also check the `device_class` in the tests to avoid such things in future 🤔

I fixed some fixtures (...) and added some more to represent ALL device classes. Furthermore, I "only" added tests for `welches` and `wie viele` since it is a rather tedious task. My brain hurt before I could start to also add tests for `irgendeins` or `alle`. I think this is fine to merge though. IMHO that many test sentences effectively hurt the process, since it starts to become too large of a set of sentences to see any inconsistencies. For these kinds of (test) cases a different testing methodology would really help.

PR #3024, opened 2025-02-28
[CS] Fix typo in test for Czech intents.

jose1711:


schizza (approved):

jose1711:
Nepresli testy - jejak sa to bije s triedou. Nic rozumnejsie nez premenovat senzor mi nenapadlo

halecivo:
I don't think the change in test with specifying sensor name is wanted - it should be possible to just ask about area, which is what failing test was covering.

PR #3059, opened 2025-03-14
Upgrade to hassil 3

synesthesiam:
Upgrade code to use hassil 3: https://github.com/home-assistant/hassil/compare/v2.2.3...v3.0.0


dotlambda:
I'm a little confused as to how this is supposed to go together with Home Assistant staying on hassil 2.2.3 but also requiring a version of home-assistant-intents that needs hassil 3.0.0.

PR #3096, opened 2025-03-19
[EN] Add area aware HassSetPosition for covers

abmantis:


sentences/en/cover_HassSetPosition.yaml
```diff
@@ -14,3 +14,14 @@ intents:
           - "(<numeric_value_set>|<open>|<close>) <area> {cover_classes:device_class} [position] to <position>"
         slots:
           domain: cover
+
+      - sentences:
+          - "(<numeric_value_set>|<open>|<close>) [the] {cover_classes:device_class} [<in_here>] [position] to <position>"
```
tetele, line 19: These should be split out. As it is, you can say "close the curtains in here position to 50%", which is not natural.

First, i'd move `[position]` before `[<in_here>]` and i'd only leave `[position]` after `<numeric_value_set>`, not in combination with open and close.

tetele (changes requested):

tetele (approved):
@mrdarrengriffin @ViViDboarder please confirm

mrdarrengriffin (approved):

PR #3114, opened 2025-03-27
[CA] Add HassBroadcast

edurenye:
CA version of https://github.com/home-assistant/intents/pull/2877

Hi @duhow I propose this sentence: "[Fes un] (broadcast|anunci|anuncia) [a tot arreu] [(que digui)|(dient que)] {message}"

Is it too big? Should I break it down?

Also, I would use "broadcast" as "anglicisme", but maybe it's just me. What do you think? Can we keep it?

### DIFF
--- a/responses/ca/HassBroadcast.yaml
+++ b/responses/ca/HassBroadcast.yaml
@@ -1,0 +1,5 @@
+language: ca
+responses:
+  intents:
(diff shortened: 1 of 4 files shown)


duhow (approved):
Gracies! 👍🏻

PR #3128, opened 2025-04-09
Move language scores script to intents repo

synesthesiam:
Moves `script/write_language_scores.py` from the packaging repo to here


language_scores.json
```diff
@@ -0,0 +1,312 @@
+{
```
balloob, line 1: Since we can generate this on the fly, no need to include it. 

Also, when generating website, make sure it’s generated in the folder that will be public.

synesthesiam: Writing the JSON file to `website/dist` now in Github workflow

balloob (approved):

PR #3140, opened 2025-04-19
[EN] Extended light set sentences

onedashone:
Extends the list of sentences that can trigger light_HassLightSet intent to include the brightness synonyms "light" and "lights".

all uses of the word "brightness" are now mapped to the expansion rule brightness_synonyms. Results in some minor odd sounding acceptable sentences such as "set the lights of the bedroom to 50% lights" but with the added benefit of the following sentences:
"set the lights in the bedroom to 50%"
"set the bedroom lights to the max"
"set the bedroom lamp to minimum light"

Change to expansion rule also allows for trivial addition of other brightness synonyms such as "illumination" or "luminosity" for those with a finer word palette. Anything unique enough for the system to infer that the intent must be light based.

Added tests duplicate brightness cases but with intermittent changes with either the word "light" or "lights" to avoid a complete duplication of tests for each synonym.


tetele (approved):

sentences/en/_common.yaml
```diff
@@ -376,6 +376,7 @@ expansion_rules:
   are_any: "[<are>] <any>"
   how_many: "how many [of <the>]"
   brightness: "{brightness}[([ ]%)| percent]"
+  brightness_synonyms: "(brightness|light[s])"
```
mrdarrengriffin, line 379: Wouldn't it be better to use the expansion rule for lights here?

onedashone: This excellent question led me down a much longer path than I originally anticipated. In short, swapping out "light[s]" with "<light>" better aligns with the current setup for the light_HassTurnOff and light_HassTurnOn sentence structures. However, once I started adding some additional tests to cover the additions "<light>" would create, I quickly realized that simply swapping out "brightness" with "<brightness_synonyms>" wasn't a good long term solution.

My new commit tries to clearly differentiate between words that define general light entities and words used as a measurement of light like brightness.

onedashone:
@mrdarrengriffin @ViViDboarder @tetele I did a pretty big overhaul to the branch that probably warrants a review. If my changes cause too many concerns, I'll revert to my last commit and simply implement mrdarrengriffin's suggestion. Here's a quick rundown of the changes:

1. Further cleaned up tests as well as added some new tests to go along with other options that can be introduced by words in <light>.
2. Added "here" and "in here" to the common's list of skip_words to simplify light_HassLightSet's sentence structure.
3. Attempted to clean up the sentence structures in light_HassLightSet for brightness changes.

The changes to sentence structure comes down to this:
1. Sentences either deal with modifying an entity, a general area (area/floor), or the implied current environment.
2. Sentences for each scenario are now permutations structured as (what's being effected);(what value is being set);(an optional unit of measurement).
3. The setting of lights to specific brightness-es or max/min values have been combined into a localized expansion rule.

onedashone:
Just realized the update to skip_words caused a few tests outside of light_HassLightSet to fail.  Should be pretty easy to fix if it's anything like the change I had to make for light_HassLightSet. I'll look into that soon.

sentences/en/light_HassLightSet.yaml
```diff
@@ -4,72 +4,30 @@ intents:
     data:
       # brightness
       - sentences:
-          - "[<numeric_value_set>] <name> brightness [to] <brightness>"
-          - "[<numeric_value_set>] [the] brightness [of] <name> [to] <brightness>"
-          - "[<numeric_value_set>] <name> [to] <brightness> [brightness]"
+          - "[<numeric_value_set>] ([of] <name>;<light_value>;[[the] <brightness_synonyms>])"
```
tetele, line 7: Although i am a fan of contraction, this template is excessive. I'll underline the issues.

1. In a permutation, having a completely optional member is a no-no. The reason for that is that there are too many generated variants which would match.

As an example, `(a;b;[c])` expands into 6 variants: `a b [c]`, `a [c] b`, `b a [c]`, `b [c] a`, `[c] a b` and `[c] b a`. Of those, when `c` is missing, 3 variants will match `a b`: `a b [c]`, `a [c] b` and `[c] a b`, which is way too many.

Now, if `c` is another expanding template, that will break into many more other variants which will match, and hassil will check them all.

TL;DR: as a general rule of thumb, don't add completely optional parts to a permutation.

2. Your proposed sentence would match stuff like `set of living room lights to 43 percent the brightness` or, worse still, `set 43% of the living room lights to light`, which doesn't make sense.

All templates you add should NOT match any sentence you don't want to match. That's the reason they were split into 3 templates originally.

sentences/en/light_HassLightSet.yaml
```diff
@@ -4,72 +4,30 @@ intents:
...
-        requires_context:
-          area:
-            slot: true
-
-      # Max/Min brightness
-      - sentences:
-          - "[<numeric_value_set>] <name> brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness of <name> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <name> [to] [the] {brightness_level:brightness} brightness"
+          light_value: "[to] (<brightness>|[the] {brightness_level:brightness})"
```
tetele, line 10: I'd extract `[to]` into the encompassing sentence and leave just the light value in this expansion rule, to make it clearer.

sentences/en/light_HassLightSet.yaml
```diff
@@ -4,72 +4,30 @@ intents:
...
         requires_context:
           domain: light
-        response: "brightness"
 
       - sentences:
-          - "[<numeric_value_set>] [the] brightness in <area> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness of <area> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <area> brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <area> [to] [the] {brightness_level:brightness} brightness"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;<light_value>;[[the] <brightness_synonyms>])"
```
tetele, line 15: This is getting very hard to read. I'll happily review it after you'll have fixed the split i mentioned earlier and apply the same to these.

sentences/en/light_HassLightSet.yaml
```diff
@@ -4,72 +4,30 @@ intents:
...
+        expansion_rules:
+          location: "([<light>];[<in>] (<area>|<floor>))"
```
tetele, line 17: There's an `<in_area_floor>` expansion rule you can use.

sentences/en/light_HassLightSet.yaml
```diff
@@ -4,72 +4,30 @@ intents:
...
         response: "brightness"
 
       - sentences:
-          - "[<numeric_value_set>] [the] brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness (<in_here>;to [the] {brightness_level:brightness})"
+          - "[<numeric_value_set>] [(<all>|<the>)] <light> (<light_value>;[<brightness_synonyms>])"
+          - "[<numeric_value_set>] [[(<all>|<the>)] <light>] (<light_value>;[the] <brightness_synonyms>)"
         expansion_rules:
-          in_here: "[in] here"
+          light_value: "[to] (<brightness>|[the] {brightness_level:brightness})"
```
tetele, line 25: Maybe this should be a global expansion rule, if it keeps repeating?

tetele (changes requested):
I'm really glad you want to make this work well, covering all use cases. There are some things you may have not been aware about, i've underlined them in comments.

sentences/en/light_HassLightSet.yaml
onedashone: Ah, I see what you mean about the permutation issues. I made sure to split up sentences further so optional permutations don't exist as well as tried to prevent sentences like you used in your second bullet point from occurring.

sentences/en/light_HassLightSet.yaml
onedashone: Done!

sentences/en/light_HassLightSet.yaml
onedashone: Nice! Swapped out `[<in>] (<area>|<floor>)` with `<in_area_floor>`.

sentences/en/light_HassLightSet.yaml
onedashone: Went ahead and moved `light_value` to the global expansion rule set.
Also, I went ahead and just removed "here" and "in here" from the global skip_words and further updated the templates to still account for those words showing up in sentences.

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,40 @@ intents:
   HassLightSet:
     data:
       # brightness
+      # specific entity control
       - sentences:
-          - "[<numeric_value_set>] <name> brightness [to] <brightness>"
-          - "[<numeric_value_set>] [the] brightness [of] <name> [to] <brightness>"
-          - "[<numeric_value_set>] <name> [to] <brightness> [brightness]"
+          - "[<numeric_value_set>] ([of] <name>;[to] <light_value>)"
+          - "[<numeric_value_set>] ([of] <name>;[to] <light_value>;[the] <brightness_synonyms>)"
```
tetele, line 9: 2. Your proposed sentence would match stuff like `set of living room lights to 43 percent the brightness` or, worse still, `set 43% of the living room lights to light`, which doesn't make sense.

All templates you add should NOT match any sentence you don't want to match. That's the reason they were split into 3 templates originally.

tetele (changes requested):

sentences/en/light_HassLightSet.yaml
onedashone: New push prevents the second command example from occurring.

The idea of permutation stops making sense to me if the first example isn't acceptable. It might also be worth considering simply making all articles and prepositions part of the skip_words list if the goal is to significantly reduce sentence combinations across the board. One upgrade at a time though.

sentences/en/_common.yaml
```diff
@@ -376,7 +376,9 @@ expansion_rules:
   are_any: "[<are>] <any>"
   how_many: "how many [of <the>]"
   brightness: "{brightness}[([ ]%)| percent]"
+  brightness_synonyms: "(brightness|light)"
   light: "(light|lights|lighting|lamp|lamps)"
+  light_value: "(<brightness>|[the] {brightness_level:brightness})"
```
tetele, line 381:
```suggestion
  brightness_value: "(<brightness>|[the] {brightness_level:brightness})"
```

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,41 @@ intents:
...
-
-      - sentences:
-          - "[<numeric_value_set>] [the] brightness to <brightness>"
-          - "[<numeric_value_set>] [the] brightness (<in_here>;to <brightness>)"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;[to] <light_value>)"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;[the] <brightness_synonyms>) [to] <light_value>"
+          - "[<numeric_value_set>] [(<all>|<the>)] <location> ([to] <light_value>;[the] <brightness_synonyms>)"
         expansion_rules:
-          in_here: "[in] here"
+          location: "(<in_area_floor>|([<light>];<in_area_floor>))"
```
tetele, line 21:
```suggestion
          location: "(<in_area_floor>|(<light>;<in_area_floor>))"
```
Not only are completely optional permutation members not recommended for the reasons I've mentioned, but having the other member listed alone as an alternative simply makes no sense.

Your version would yield `(<in_area_floor>|<in_area_floor>|<light> <in_area_floor>|<in_area_floor>|<in_area_floor> <light>)` whereas the suggestion would yield `(<in_area_floor>|<light> <in_area_floor>|<in_area_floor> <light>)`

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,41 @@ intents:
...
-
-      - sentences:
-          - "<numeric_value_set> <area> to <brightness>"
-        response: "brightness"
-
-      - sentences:
-          - "[<numeric_value_set>] [the] brightness to <brightness>"
-          - "[<numeric_value_set>] [the] brightness (<in_here>;to <brightness>)"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;[to] <light_value>)"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;[the] <brightness_synonyms>) [to] <light_value>"
```
tetele, line 18: This would match `set the upstairs light light to 50%`. I don't really think that `<location>` should contain `<light>` 🤔

tetele (changes requested):
Sorry for the delay, I've been on a small vacation.

sentences/en/light_HassLightSet.yaml
onedashone: Yep, this one was just a slip up on my part. Fixed now.

sentences/en/light_HassLightSet.yaml
onedashone: Yeah, this is one of those sentence types that are grammatically correct but sound pretty weird since "light" works both as a way of defining objects that produce luminescence and a unit of measurement for luminescence; similar to saying something like `the orange orange`. Sure, `set the upstairs light light` sounds weird; but, `set the upstairs upstairs lamp light` still makes sense to say despite probably fitting better in the Victorian era.

While there are some ways to clean this up, I don't think removing `<light>` from `<location>` is a good idea. It's the whole reason I created this pull request in the first place since there wasn't a way before to say a command like `set the lights to 50%` which we can do when strictly turning lights on and off.

Personally I'd like to leave this as is; but if that would prevent this from getting accepted, what I'd recommend doing instead is removing "light" as a brightness synonym instead and removing any test sentences that depended on that. Something like `light light` couldn't happen anymore but we still would maintain backwards compatibility with the voice commands for turning lights on and off while only losing sentences that are less likely to be missed such as `set bedroom to 50% light`.

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,41 @@ intents:
...
       - sentences:
-          - "[<numeric_value_set>] <name> brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness of <name> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <name> [to] [the] {brightness_level:brightness} brightness"
-        requires_context:
-          domain: light
-        response: "brightness"
+          - "[<numeric_value_set>] [<the>] <light> [to] <brightness_value>"
+          - "[<numeric_value_set>] [<the>] <light> ([to] <brightness_value>;<here>)"
+          - "[<numeric_value_set>] [(<all>|<the>)] <light> ([to] <brightness_value>;<here>)"
```
tetele, line 28: These can be contracted, right?
```suggestion
          - "[<numeric_value_set>] [(<all>|<the>)] <light> ([to] <brightness_value>;<here>)"
```

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,41 @@ intents:
...
 
-      - sentences:
-          - "[<numeric_value_set>] [the] brightness in <area> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness of <area> to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <area> brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] <area> [to] [the] {brightness_level:brightness} brightness"
-        response: "brightness"
+          - "[<numeric_value_set>] [<the>] <light> ([to] <brightness_value>;<brightness_synonyms>)"
+          - "[<numeric_value_set>] [<the>] <light> (([to] <brightness_value>;<brightness_synonyms>);<here>)"
+          - "[<numeric_value_set>] [(<all>|<the>)] <light> (([to] <brightness_value>;<brightness_synonyms>);<here>)"
```
tetele, line 32:
```suggestion
          - "[<numeric_value_set>] [(<all>|<the>)] <light> (([to] <brightness_value>;<brightness_synonyms>);<here>)"
```

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,41 @@ intents:
...
 
-      - sentences:
-          - "[<numeric_value_set>] [the] brightness to [the] {brightness_level:brightness}"
-          - "[<numeric_value_set>] [the] brightness (<in_here>;to [the] {brightness_level:brightness})"
-        expansion_rules:
-          in_here: "[in] here"
+          - "[<numeric_value_set>] ([to] <brightness_value>;[the] <brightness_synonyms>)"
+          - "[<numeric_value_set>] ([to] <brightness_value>;[the] <brightness_synonyms>;<here>)"
```
tetele, line 35: This is too permissive, I think
`set to 50% here the brightness`
`set here to 50% the brightness`

Maybe this is better?
```suggestion
          - "[<numeric_value_set>] ([to] <brightness_value>;[the] <brightness_synonyms> [<here>])"
```

tetele (changes requested):

sentences/en/light_HassLightSet.yaml
onedashone: Yep, another good catch. I went ahead and made this change as well as the similar suggestion below.

sentences/en/light_HassLightSet.yaml
onedashone: Makes sense to me. Went ahead and made this suggested change as well and updated a few of the tests accordingly.

tetele (approved):
@mrdarrengriffin @ViViDboarder please take a look

tetele:
@mrdarrengriffin @ViViDboarder there has been a lot of work put into this PR. Please take a look when you can.

mrdarrengriffin (approved):

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,39 @@ intents:
   HassLightSet:
     data:
       # brightness
+      # specific entity control
       - sentences:
-          - "[<numeric_value_set>] <name> brightness [to] <brightness>"
-          - "[<numeric_value_set>] [the] brightness [of] <name> [to] <brightness>"
-          - "[<numeric_value_set>] <name> [to] <brightness> [brightness]"
+          - "[<numeric_value_set>] ([of] <name>;[to] <brightness_value>)"
```
ViViDboarder, line 8: This expands to a few sentences that don't make sense to me, but I guess aren't ambiguous:
- Change the of Living Room Light to 50%
- Living Room Light 50%
- Of Living Room Light 50%

I also don't think you need the `;`, because I can't think of a sentence where I'd want the percent to come before the name.
- To 50% of Living Room Light

Might be good to use less permutations so you can get the prepositions in the right spot too.

- `[<numeric_value_set>] [[the] <brightness_synonyms> of] <name> [to] <brightness_value>`

Maybe give this another pass removing the permutations (`;`) so we can strike the ones that don't make sense.

sentences/en/light_HassLightSet.yaml
```diff
@@ -3,73 +3,39 @@ intents:
...
-        response: "brightness"
-
-      - sentences:
-          - "<numeric_value_set> <area> to <brightness>"
-        response: "brightness"
-
-      - sentences:
-          - "[<numeric_value_set>] [the] brightness to <brightness>"
-          - "[<numeric_value_set>] [the] brightness (<in_here>;to <brightness>)"
+          - "[<numeric_value_set>] ([of] [(<all>|<the>)] <location>;[to] <brightness_value>)"
```
ViViDboarder, line 17: Similar to above. Don't think sentences with percent first make sense and the `of` doesn't make sense if not following `<brightness_synonyms>`.

ViViDboarder (changes requested):
Thanks for the PR. I think splitting up a few of the permutations will make this fairly easy to merge.

sentences/en/light_HassLightSet.yaml
onedashone: Good catch with the percentage happening before entity declaration. Latest commit fixes that.

sentences/en/light_HassLightSet.yaml
onedashone: Likewise, I fixed the issue with the percentage happening before the location being defined.

sentences/en/light_HassLightSet.yaml
onedashone: As for the issues with permutation and odd sounding sentences, I'm of the opinion that an unambiguous sentence should result in the right intent being triggered regardless of how it sounds grammatically. Y'all are welcome to modify this however you want; but, I think this pull request has gotten long enough for me.

ViViDboarder: I agree. Thanks! I’ll try to get this merged next week.

onedashone: @ViViDboarder was there something else that needed to be done here before this is ready to merge?

onedashone:
@tetele @ViViDboarder  @mrdarrengriffin Is there something else that needs to be done for this merge request? It's been 3 months since it sounds like this was ready to merge.

balloob:

---

PR #3142, opened 2025-04-21
Being transition of intents repo to slot combinations

synesthesiam:
**DRAFT**

See `docs/slot_combinations.md` file for explanation


docs/slot_combinations.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Sentences
+
+For each intent and slot combination found in `intents.yaml`, a corresponding file named `sentences/<language>/<intent>/<slot_combination>.yaml` may be found. When the slot combination has an `importance` of "required", this file must be present for validation to succeed. If the `importance` is "usable", a warning will be issued when the file is missing.
+
+All sentence templates in a slot combination YAML file must match the exact slots specified in `intents.yaml`. Additional validation of sentence templates has been added to ensure that:
+
+**list references cannot be present in alternatives and optionals**
+
+In other words, `(text|{list_name})` and `[{list_name}]` are no longer allowed in sentence templates. This ensures that a sentence template always matches the same slots.
```
tetele, line 16: We need to be careful with lists nested under expansion rules here. for instance `<name>: [the] {name}` should not be used un constructs such as `(text|<name>)`.

tetele: Oh, nevermind, I didn't read everything :D

tetele:
What's the purpose of all these rule changes? I get the wish to transition to slot combinations, but I don't get the extra rigidity introduced by rules not being able to contain other rules or lists.

There will be languages where these new restrictions will render the current translations useless and it just might be that there is no longer an active language leader to make the transition, thus introducing breaking changes.

docs/slot_combinations.md
```diff
@@ -0,0 +1,53 @@
...
+
+Importantly, all of the information needed to correctly generate the `requires_context` portion of the sentence templates is present in `intents.yaml`. Template authors are no longer responsible this part, making it much easier to ensure consistency across languages.
```
abmantis, line 18:
```suggestion
Importantly, all of the information needed to correctly generate the `requires_context` portion of the sentence templates is present in `intents.yaml`. Template authors are no longer responsible for this part, making it much easier to ensure consistency across languages.
```

staticdev:
@synesthesiam I find the new structure cleaner and will make possible to have clear the slots needed to fulfill the intent. The only stress you already found out is to transition all languages. To make it a bit easier I would propose a slightly different structure where we favor comparison on how to do in different languages. Since I have been working recently in the repo I suffer a bit to compare for example sentence, response and tests of same intent in different languages.

For me it would make easier if they were structured like:
`<intent>/<slot_combination>/<language>-common.yaml`
`<intent>/<slot_combination>/<language>-sentences.yaml`
`<intent>/<slot_combination>/<language>-responses.yaml`

And then the tests following the same idea something like:
`tests/<intent>/<slot_combination>/<language>.yaml`

I would like to know also what other people working here think about the structure if it would also help them.

synesthesiam:
Another option would be to group everything by language first, so sentences and everything would be together for each language.

Something to consider is the assignment of codeowners in GitHub. We do this by directory, and each person leads a specific language. So the final structure needs to have language high up in the structure to avoid having to add more codeowner rules over time.

staticdev:
> Another option would be to group everything by language first, so sentences and everything would be together for each language.
> 
> Something to consider is the assignment of codeowners in GitHub. We do this by directory, and each person leads a specific language. So the final structure needs to have language high up in the structure to avoid having to add more codeowner rules over time.

Then to keep the folder assignment of codeowners by language maybe: 
For me it would make easier if they were structured like:
<intent>/<slot_combination>/<language>/common.yaml
<intent>/<slot_combination>/<language>/sentences.yaml
<intent>/<slot_combination>/<language>/responses.yaml

And then the tests following the same idea something like:
tests/<intent>/<slot_combination>/<language>/tests.yaml
tests/fixtures/<language>/*.yaml (could modularize the fixtures in diferent names but everything here is imported for the language tests)

tetele:
I think the rules can be grouped into concern-based files, like `verbs.yaml` (where you have `set`, `open` and `turn` expansion rules), `questions.yaml` (for stuff like `what_is`, `which_is`, `are_any` etc.).

The way it is, I find it very hard to maintain and I don't see a reason for one file per rule. OK, having them all in `_common.yaml` was too much, but so is having one file per rule.

synesthesiam:
Based on feedback and testing, I've made a few changes:

* Restrictions on expansion rules have been removed; they are recommendations
* Lists and expansion rules can be grouped in a file (e.g., `lights.yaml` and `verbs.yaml`)
* The importance level is per domain group when `{name}` or `{domain}` are present, otherwise at the slot combination level

See `docs/slot_combinations.md` for more details

PR #3143, opened 2025-04-21
French answer

Slim1399:
correct the answer in French instead of having On/Off

### DIFF
--- a/responses/fr/HassGetState.yaml
+++ b/responses/fr/HassGetState.yaml
@@ -8,7 +8,13 @@ responses:
         {{ state.state_with_unit }}
 
       one: |
-        {{ slots.name | capitalize }} est {{ state.state_with_unit }}
+        {% if state.state == "on" %}
+          {{ slots.name | capitalize }} est allumée
+        {% elif state.state == "off" %}
+          {{ slots.name | capitalize }} est éteinte
+        {% else %}
+          {{ slots.name | capitalize }} est {{ state.state_with_unit }}
+        {% endif %}  
 
       one_yesno: |
         {% if query.matched %}


piitaya:
This PR assumes that all name are feminine which is not the case, right? It will work for "lampe", "lumière" but not for "spot" or "ventilateur".

Slim1399:
> This PR assumes that all name are feminine which is not the case, right? It will work for "lampe", "lumière" but not for "spot" or "ventilateur".

When it's a list, there are no more personal pronouns, so no !

jlpouffier:
List are not supported 
As we cannot detect gender of name today, we are assuming everything is masculine.

Slim1399:
Mais vous n'avez pas besoin de savoir le genre dans ce cas là !

piitaya:
You were modifying a wrong part of the code. The "which" answer should be changed not the "one".
Also, you added feminine version of the verb. It must be masculine because we do not know the gender.

Slim1399:
> You were modifying a wrong part of the code. The "which" answer should be changed not the "one". Also, you added feminine version of the verb. It must be masculine because we do not know the gender.

I did my best, if you would be so kind as to correct it.

Slim1399:
Hi,


    `HassGetState:
      which: |
        {% set etat = {
          "on": "allumé",
          "off": "éteint",
          "open": "ouvert",
          "closed": "fermé",          
          "unavailable": "indisponible",
          "unknown": "inconnu"
        } %}
        {% if not query.matched %}
          Aucun appareil
        {% else %}
          {% set match = query.matched | map(attribute="name") | sort | list %}
          {% set count_match = match | length %}
          {% set state_text = etat.get(state.state, state.state_with_unit) %}

          {% if count_match > 4 %}
            {{ match[:3] | join(", ") }} et {{ count_match - 3 }} autres sont {{ state_text }}s
          {% elif count_match == 1 %}
            {{ match[0] }} est {{ state_text }}
          {% else %}
            {% for name in match %}
              {% if not loop.first and not loop.last %}, {% elif loop.last %} et {% endif %}
              {{ name }}
            {% endfor %}
            sont {{ state_text }}s
          {% endif %}
        {% endif %}  `


it works perfectly !

PR #3164, opened 2025-05-09
[NL] cover: add verbs to control blind|shutter|shade

kritsel:
Add more verbs to control blinds, shutters and shades.

* HassTurnOn: `optrekken`, `ophalen`
* HassTurnOff: `laten zakken`, `neer laten`

### DIFF
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -673,7 +673,7 @@ expansion_rules:
       |[<in> [<area>]] [<by>] <name>
       |[<by>] <name> [<in>] <area>
     )
-  change: "(zet|mag|mogen|doe|verander|maak|schakel|haal|trek|laat)"
+  change: "(zet|mag|mogen|doe|verander|maak|schakel)"
   would: "(kan|kun|wil) (je|jij)"
   to: "(naar|op)"
   is: "(zijn|is|staa(n|t)|zit[ten]|word[t|en]|lig(t|gen))"
@@ -721,6 +721,9 @@ expansion_rules:
   # cover states
   open: "(open|omhoog|naar boven|opwaarts|op|optrekken|ophalen)"
   closed: "(dicht|omlaag|naar beneden|neerwaarts)"
+  # cover actions:
+  open_action: "(openen|open doen|optrekken|ophalen)"
+  open_command: "(open|haal|trek)"
   # binary_sensor specific
   detect: "(detecteert|registreert|detecteren|registreren|gedetecteerd|geregistreerd|waar[ ]genomen)"
   sensor: "[een|de] (apparaat|apparaten|sensor[s|en]|iets)"
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -673,7 +673,7 @@ expansion_rules:
       |[<in> [<area>]] [<by>] <name>
       |[<by>] <name> [<in>] <area>
     )
-  change: "(zet|mag|mogen|doe|verander|maak|schakel)"
+  change: "(zet|mag|mogen|doe|verander|maak|schakel|haal|trek|laat)"
   would: "(kan|kun|wil) (je|jij)"
   to: "(naar|op)"
   is: "(zijn|is|staa(n|t)|zit[ten]|word[t|en]|lig(t|gen))"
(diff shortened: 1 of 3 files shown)


kritsel:
Hi @TheFes, note that this PR is still a draft and only covers cover_HassTurnOn

I'm not yet entirely happy with the points below, and I'm interested in hearing your thoughts before I add similar patterns to cover_HassTurnOff.

1. some verb variations are explicitly mentioned in the sentences, while others are expressed via expansion_rules in _common.yaml (`open` and `closed')
2.  the expansion_rules `open` and `closed` now contain a mix of words which denote a direction (e.g. `omhoog`) as well as verbs (e.g. `optrekken')

sentences/nl/_common.yaml
```diff
@@ -673,7 +673,7 @@ expansion_rules:
       |[<in> [<area>]] [<by>] <name>
       |[<by>] <name> [<in>] <area>
     )
-  change: "(zet|mag|mogen|doe|verander|maak|schakel)"
+  change: "(zet|mag|mogen|doe|verander|maak|schakel|haal|trek|laat)"
```
TheFes, line 676: The cover specific commands (`haal`, `trek`, `laat`) should be in a separate expansion rule (eg `<change_cover>`), otherwise you can say thinks link `haal de lampen in de woonkamer uit`

sentences/nl/cover_HassTurnOn.yaml
```diff
@@ -72,10 +72,10 @@ intents:
...
       - sentences:
-          - "open [de|het] (<blind>|<shutter>|<shade>)"
-          - "open ([de|het] (<blind>|<shutter>|<shade>);<here>)"
-          - "[<would>] [de|het] (<blind>|<shutter>|<shade>) (openen|open doen)"
-          - "[<would>] ([de|het] (<blind>|<shutter>|<shade>);<here>) (openen|open doen)"
+          - "[open|haal|trek] [de|het] (<blind>|<shutter>|<shade>) [op]"
+          - "[open|haal|trek] ([de|het] (<blind>|<shutter>|<shade>);<here>) [op]"
+          - "[<would>] [de|het] (<blind>|<shutter>|<shade>) (openen|open doen|optrekken|ophalen)"
+          - "[<would>] ([de|het] (<blind>|<shutter>|<shade>);<here>) (openen|open doen|optrekken|ophalen)"
           - "[<change>] [de|het] (<blind>|<shutter>|<shade>) [<here>] <open>"
```
TheFes, line 79: With the other comment taken into account, this should then be something like `[<change>|<change_cover>]`

TheFes (changes requested):
In general we should avoid to add wrong sentences.
With the current approach you can now use sentences like `verander de gordijnen op`, where `op` should only be used in combination with words like `trek` and `haal`.
Can you try to split the usage of `op` and other similar words into separate sentences so they can not be used incorrectly?

sentences/nl/_common.yaml
```diff
@@ -719,8 +719,11 @@ expansion_rules:
   temperature: "{temperature}[<degrees>]"
   warm: "(warm|heet|koud|koel)"
   # cover states
-  open: "(open|omhoog|naar boven|opwaarts)"
+  open: "(open|omhoog|naar boven|opwaarts|op|optrekken|ophalen)"
```
kritsel, line 722: I see now that I forgot to undo this change. The new values should be removed.

sentences/nl/cover_HassTurnOn.yaml
kritsel: Decided to introduce `<open_command>` (to support commanding sentences with `haal op`) and `<open_action`> (to support requestings sentences with `ophalen`) instead of your suggested generic `<change_cover>`. Makes it easier to target the close counterparts in their own expressions as well. Does that make sense or would you opt for a more generic approach?

sentences/nl/_common.yaml
kritsel: Undid this change. See comment below for more explanation.

kritsel:
@TheFes Take two, I believe this is now heading in the right direction.

kritsel:
@TheFes, could you take a look at the second commit when you have time?

TheFes:
It's still marked as draft, can you make it ready for review again

kritsel:
@TheFes PR is open for review

TheFes (approved):
Looks good like this! Thanks!

PR #3183, opened 2025-05-25
[NL] scene and script: add sentence structures

kritsel:
* add sentence structure more suitable for scenes/scripts named after an activity (the existing sentence structure are suitable for scenes/scripts named after a mode/state)
* move several activation verbs from scene/script specific files to common

### DIFF
--- a/sentences/nl/_common.yaml
+++ b/sentences/nl/_common.yaml
@@ -659,6 +659,7 @@ expansion_rules:
   area: "[de|het] {area}"
   floor: "[de|het] {floor}"
   in: "[in|op|van|bij]"
+  aan: "(in|aan)"
   by: "(door|met|bij)"
   here: "(hier|in deze (kamer|ruimte))"
   name_area: >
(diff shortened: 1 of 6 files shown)


sentences/nl/_common.yaml
```diff
@@ -659,6 +659,7 @@ expansion_rules:
   area: "[de|het] {area}"
   floor: "[de|het] {floor}"
   in: "[in|op|van|bij]"
+  aan: "(in|aan)"
```
TheFes, line 662: all expansion rules are in English, so this should be `on` then.

TheFes: Not sure if we should add this anyway, the only verb which uses `in` is `schakelen`. But this way you can say `zet feestmodus in`.
I must admit that the sentences in the previous version also used incorrect combinations, but I'd rather not add more incorrect phrases.
In https://github.com/home-assistant/intents/blob/main/sentences/nl/homeassistant_HassTurnOn.yaml the sentences with `schakel in` have been split off to avoid wrong sentences with `in`

TheFes (changes requested):
Do you think sentences for areas and floors are really needed? If people use scripts with the same name in different areas, it could also be solved with aliases.

sentences/nl/_common.yaml
kritsel: About `on`: I actually started out with `on` , but it resulted in an error about a missing expansion_rule. I could only resolve it by changing `on` to something else in both `_common.yaml` and the sentence file that refers to it. Really strange.

kritsel: About the construct itself: Agree, this construct can match incorrect sentences. Trade off between a generic construct on one side and strictly matching only correct sentences on the other side. 
I slightly favour the generic option (as the implementation shows :-), but happy to follow your request to use the more explicit/strict pattern to match the other intents files.

kritsel:
PS, I'll be on holiday until next Monday so the next iteration will have to wait a while.

balloob:

Dutch has just been migrated to a brand new sentence format in #3882. The new slot-combination format makes it much easier to **write, validate, and maintain** sentences going forward. Unfortunately, this PR is written against the old format, so it can no longer be merged as-is.

You can follow the progress and coverage for Dutch here: https://ohf-voice.github.io/intents/language_nl.html

---

PR #3186, opened 2025-05-27
[DE] Shopping list complete item

Se7enair:
it was not possible to add items to another list then the shopping list.
Now you can add items to your shopping list, delete them and also to other lists.
If you are using as list title only `Liste` then the shopping list will be the default.

### DIFF
--- a/sentences/it/_common.yaml
+++ b/sentences/it/_common.yaml
@@ -401,7 +401,7 @@ expansion_rules:
   which: "(qual[e|i] sono | che)"
   how_many: "quant(i|e) [sono]"
   hello: "(ciao|salve)"
-  my: "(il mio|i miei)"
+  my: "([il|i] (mi[o|a|e|ei]|nostr[o|a|e|i]))"
 
   # Dispositivi
   all: "(tutt(o|e|i|a) [quant(i|e)] | ogni)"
--- a/sentences/it/shopping_list_HassShoppingListCompleteItem.yaml
+++ b/sentences/it/shopping_list_HassShoppingListCompleteItem.yaml
@@ -6,5 +6,5 @@ intents:
           - (controlla[re] | completa[re] | elimina[re] | rimuov(i|ere) | spunta[re] ) [<the>] <item> (dalla | <on> | <in>) <my_list>
         response: item_completed
         expansion_rules:
-          my_list: "[(mio|mia|mie|miei|nostra|nostre)] (lista [della spesa] | shopping list)"
+          my_list: "[<my>] (lista [della spesa] | shopping list)"
           item: "{shopping_list_item:item}"


Se7enair:
Oh man, test failed. I thought I tested before requesting review. Nevertheless, now everything should be fine. And next time less commits ;)
Figured out how to completely work with codespace.

Se7enair:
To be honest, I dont know how to split a PR. I deleted the changes and will create a new PR for "packen".

responses/de/HassShoppingListCompleteItem.yaml
```diff
@@ -0,0 +1,5 @@
+language: de
+responses:
+  intents:
+    HassShoppingListCompleteItem:
+      item_completed: "{{ slots.item }} gelöscht"
```
Lesekater, line 5: In HassListCompleteltem, the Response is
"{{ slots.item }} abgehakt". I think we should use the same response here.

Lesekater (changes requested):

responses/de/HassShoppingListCompleteItem.yaml
Se7enair: this is fine for me

Lesekater (approved):

sentences/de/_common.yaml
```diff
@@ -486,6 +486,9 @@ lists:
   shopping_list_item:
     wildcard: true
 
+  todo_list_item:
+    wildcard: true
+
```
mib1185: this addition of `todo_list_item` seems to be unrelated to the shopping list topic?

tests/de/shopping_list_HassShoppingListCompleteItem.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - Lösche Milch von der Einkaufsliste
+      - Streiche Milch von meiner Einkaufsliste
+      #- (entferne|lösche|streiche) <item> (von|auf|aus) <mein_einkauf_dativ>
+      - Entferne Milch von meinem Einkauf
+      - Entferne Milch von dem Einkauf
+      - Lösche Milch von meinem Einkauf
+      - Lösche Milch von dem Einkauf
+      - Streiche Milch von meinem Einkauf
+      - Streiche Milch von dem Einkauf
+      #- (schließe|hake) <item> (von|auf|aus) <meine_liste_dativ> ab
```
mib1185, line 20: we need also tests for `hake ... ab`

mib1185 (changes requested):
Hi @Se7enair 
overall this looks good :+1: just two small comments
and yes, there is no defined "split a PR action" it is just removing unrelated parts and put them into a new PR/branch :wink:

sentences/de/_common.yaml
mib1185: This is still unresolved

mib1185 (changes requested):

Se7enair:
Now everything which is not directly related to the main thing in this PR should be cleaned. From my point of view ready to merge.

sentences/de/_common.yaml
andreasbrett: it is meanwhile

PR #3218, opened 2025-06-10
[GA] Added GetCurrentTime intents

GGamerzz:
- Added 12 variations of user queries asking for the current time in Irish
- Updated expected response to use full phrasing: "Tá sé 1:02 ar maidin"
- Avoided time abbreviations like "r.n." and "i.n." to align with natural Irish expressions
- Follows Home Assistant voice intent style and formatting guidelines


GGamerzz:
> > require the user to read the sentence instead of just seeing the numbers in the text
> > …
> > stand out more to my eyes
> 
> Since this repo is related to voice control I wouldn't place too much emphasis on how it appears in text. To me the more important aspect is how it's spoken. If we go with `1:02` etc., how would that be read aloud to the user?

Good point on the appearance in text. Yeah it's definitly something that I will look at. I would be more inclined upon further thinking to go with the "tar éis/chun" route as in Irish you wouldn't really say "dáréag tríocha" (twelve thirty)" when asked the time, you would say "leathuair tár eis a dó" for example. 

It makes sense in my head :). I think I will end up going the "tar é/chun" route

synesthesiam:
I would recommend focusing on how it's spoken. Most HA voice assistants don't have a screen, but all have a speaker. 
In the future, I think it would be worth adding both spoken and written forms (for display) to some responses.

GGamerzz:
Not happy with how my code was, doing more work on it from feedback received, thank you everyone

PR #3219, opened 2025-06-12
Add new response format for HassShoppingListCompleteItem

Lesekater:
Add new response format like in the [media search and play intent](https://github.com/home-assistant/intents/blob/main/responses/en/HassMediaSearchAndPlay.yaml).
[Discussed](https://github.com/home-assistant/core/pull/128565#discussion_r2091770222) in the connected, merged [core pr](https://github.com/home-assistant/core/pull/128565).


ViViDboarder (commented):
I've never reviewed any of the test scripts before, but this at least looks correct to me on glance. Do tests pass? I would think changing the response might break some tests.

Lesekater:
> I've never reviewed any of the test scripts before, but this at least looks correct to me on glance. Do tests pass? I would think changing the response might break some tests.

The positive response stays the same, so the [test for that](https://github.com/OHF-Voice/intents/blob/441e5102681cd86c921d02c11159ddcdf86752a8/tests/en/shopping_list_HassShoppingListCompleteItem.yaml#L3-L14) does still pass without changes. I adjusted the tests for the negative response and for mocking the behavior of the intent in the core repo.
The old test and the new one pass.

balloob:

PR #3222, opened 2025-06-15
Fix "ligar" command to work with other devices

mhalano:
Remove domain vacuum from turnOn intent, since it already has its own intent.

This was done according the discussion https://community.home-assistant.io/t/problem-with-multipart-aliases-for-a-device-in-portuguese/901819/3


mhalano:
You can see [here](https://github.com/OHF-Voice/intents/blob/main/sentences/en/homeassistant_HassTurnOn.yaml) the en file doesn't have `vacuum` as domain.

mhalano:
It seems these lines are necessary in Portuguese, even they are not in English.

PR #3223, opened 2025-06-18
[FR] Fix HassGetTemperatureResponse

jlpouffier:
A small fix for HassGetTemperatureResponse.

It was not compatible with sensors in the area slot.

Before:
CleanShot 2025-06-18 at 15 38 35

After:
CleanShot 2025-06-18 at 15 37 46


responses/fr/HassClimateGetTemperature.yaml
```diff
@@ -2,4 +2,11 @@ language: fr
 responses:
   intents:
     HassClimateGetTemperature:
-      default: "{{ state_attr(state.entity_id, 'current_temperature') }} degrés"
+      default: >
+        {% set current_temperature = state_attr(state.entity_id, 'current_temperature') %}
+        {% set temperature = state.state if current_temperature is none else current_temperature %}
+        {% if temperature == 1: %}
```
balloob, line 8: `state.state` is a string I think. So it would need to be matched against `"1"` I think?

jlpouffier: I copied what was on every language. But I can double check tomorrow. Easy test

PR #3226, opened 2025-06-18
English sentences media search play

balloob:
Requires #3225

Add some English sentences for the new SearchAndPlay


ViViDboarder (approved):
Sentences all look good. I can probably expand on them by adding synonyms in a new PR.

synesthesiam (approved):

PR #3228, opened 2025-06-20
[GA] Add Irish language support for HassGetCurrentTime intent

GGamerzz:
- Implemented intent sentences in Irish for querying the current time
- Added templated response to generate natural time phrases in Irish
- Ensured compatibility with live slotless queries (e.g., “cén t-am é”)
- Included unit tests for common time-related phrases

### DIFF
--- a/responses/ga/HassGetCurrentTime.yaml
+++ b/responses/ga/HassGetCurrentTime.yaml
@@ -14,6 +14,8 @@ responses:
           {% set minute = 2 %}
         {% endif %}
         {% set next_hour = (hour + 1) % 24 %}
+        {% set is_default = (hour == 1 and minute == 2) %}
+
 
         {% set hours_map = {
           0: "a dó dhéag",
@@ -46,7 +48,7 @@ responses:
           1: "nóiméad amháin",
           2: "dhá nóiméad",
           3: "trí nóiméad",
-          4: "ceithre nóiméad",
+          4: "ceathair nóiméad",
           5: "cúig nóiméad",
           6: "sé nóiméad",
           7: "seacht nóiméad",
@@ -56,17 +58,17 @@ responses:
           11: "aon nóiméad déag",
           12: "dhá nóiméad déag",
           13: "trí nóiméad déag",
-          14: "ceithre nóiméad déag",
+          14: "ceathair nóiméad déag",
           15: "cúig nóiméad déag",
           16: "sé nóiméad déag",
           17: "seacht nóiméad déag",
           18: "ocht nóiméad déag",
           19: "naoi nóiméad déag",
           20: "fiche nóiméad",
-          21: "fiche nóiméad agus nóiméad amháin",
-          22: "fiche dhá nóiméad",
+          21: "fiche haon nóiméad",
+          22: "fiche dó nóiméad",
           23: "fiche trí nóiméad",
-          24: "fiche ceithre nóiméad",
+          24: "fiche ceathair nóiméad",
           25: "fiche cúig nóiméad",
           26: "fiche sé nóiméad",
           27: "fiche seacht nóiméad",
@@ -74,19 +76,31 @@ responses:
           29: "fiche naoi nóiméad"
         } %}
 
+        {% set time_suffix = "" %}
+        {% if is_default %}
+           {% set time_suffix = " san oíche" %}
+        {% elif hour < 12 %}
+           {% set time_suffix = " ar maidin" %}
+        {% elif hour < 22 %}
+           {% set time_suffix = " sa tráthnóna" %}
+        {% else %}
+           {% set time_suffix = " san oíche" %}
+        {% endif %}
+
+
         {% if hour == 0 and minute == 0 %}
           Tá sé meán oíche.
         {% elif hour == 12 and minute == 0 %}
           Tá sé meán lae.
         {% elif minute == 15 %}
-          Tá sé ceathrú tar éis {{ hours_map[hour] }}.
+          Tá sé ceathrú tar éis {{ hours_map[hour] }}{{ time_suffix }}.
         {% elif minute == 30 %}
-          Tá sé leathuair tar éis {{ hours_map[hour] }}.
+          Tá sé leathuair tar éis {{ hours_map[hour] }}{{ time_suffix }}.
         {% elif minute == 45 %}
-          Tá sé ceathrú chun {{ hours_map[next_hour] }}.
+          Tá sé ceathrú chun {{ hours_map[next_hour] }}{{ time_suffix }}.
         {% elif minute < 30 %}
-          Tá sé {{ minutes_map.get(minute, minute ~ ' nóiméad') }} tar éis {{ hours_map[hour] }}.
+          Tá sé {{ minutes_map.get(minute, minute ~ ' nóiméad') }} tar éis {{ hours_map[hour] }}{{ time_suffix }}.
         {% else %}
           {% set remaining = 60 - minute %}
-          Tá sé {{ minutes_map.get(remaining, remaining ~ ' nóiméad') }} chun {{ hours_map[next_hour] }}.
+          Tá sé {{ minutes_map.get(remaining, remaining ~ ' nóiméad') }} chun {{ hours_map[next_hour] }}{{ time_suffix }}.
         {% endif %}
--- a/responses/ga/HassGetCurrentTime.yaml
+++ b/responses/ga/HassGetCurrentTime.yaml
@@ -2,11 +2,91 @@ language: ga
 responses:
   intents:
     HassGetCurrentTime:
-      default: "TODO: {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
+      default: >
+        {% if slots.time is defined %}
+          {% set hour = slots.time.hour %}
+          {% set minute = slots.time.minute %}
+        {% elif context['time.hour'] is defined %}
+          {% set hour = context['time.hour'] | int %}
+          {% set minute = context['time.minute'] | int %}
+        {% else %}
+          {% set hour = 1 %}
+          {% set minute = 2 %}
+        {% endif %}
+        {% set next_hour = (hour + 1) % 24 %}
 
-        {% if slots.time.hour == 0: %} 12:{{ minute_str }} AM {% elif slots.time.hour
-        < 12: %} {{ slots.time.hour }}:{{ minute_str }} AM {% elif slots.time.hour
-        == 12: %} 12:{{ minute_str }} PM {% else: %} {{ slots.time.hour - 12 }}:{{
-        minute_str }} PM {% endif %}
+        {% set hours_map = {
+          0: "a dó dhéag",
+          1: "a haon",
+          2: "a dó",
+          3: "a trí",
+          4: "a ceathair",
+          5: "a cúig",
+          6: "a sé",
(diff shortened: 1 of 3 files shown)


responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,91 @@ language: ga
 responses:
   intents:
     HassGetCurrentTime:
-      default: "TODO: {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
+      default: >
+        {% if slots.time is defined %}
```
AlanGreene: Should we add suffixes depending on the hour? e.g. `ar maidin`, `um thráthnóna`, `istoíche`

These were present in [the previous PR](https://github.com/OHF-Voice/intents/pull/3218), but perhaps there was a reason for omitting them here.

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,91 @@ language: ga
...
+          18: "a sé",
+          19: "a seacht",
+          20: "a hocht",
+          21: "a naoi",
+          22: "a deich",
+          23: "a haon déag"
+        } %}
 
-        "
+        {% set minutes_map = {
```
AlanGreene, line 26: General question, would it be more natural to omit `nóiméad` in many cases?

e.g. keep `nóiméad amháin tar éis`, but then `cúig tar éis`, `fiche chun`, etc.

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,91 @@ language: ga
...
+          12: "dhá nóiméad déag",
+          13: "trí nóiméad déag",
+          14: "ceithre nóiméad déag",
+          15: "cúig nóiméad déag",
+          16: "sé nóiméad déag",
+          17: "seacht nóiméad déag",
+          18: "ocht nóiméad déag",
+          19: "naoi nóiméad déag",
+          20: "fiche nóiméad",
+          21: "fiche nóiméad agus nóiméad amháin",
```
AlanGreene: Would this not be `aon nóiméad is fiche` / `aon nóiméad fichead`, `dhá nóiméad is fiche`, etc.?

Alternatively `fiche a haon`, `fiche a dó`, etc. if dropping `nóiméad`

AlanGreene (commented):
Mise arís (it's me again…) with a few questions 👋

I almost missed this as it doesn't have the `lang: ga` label 🤷

responses/ga/HassGetCurrentTime.yaml
GGamerzz: I think this is the biggest hurdle/challenge we will come across, asking my family we agree that adding `nóiméad` into every case would be better, personally I find it like its taking a shortcut when you leave it out. 
So I think I will leave the `nóiméad` in.

responses/ga/HassGetCurrentTime.yaml
GGamerzz: `20: "fiche nóiméad",
 21: "fiche a haon nóiméad",
 22: "fiche a dó nóiméad",   
    23: "fiche a trí nóiméad",  
          24: "fiche a ceithre nóiméad",  
          25: "fiche a cúig nóiméad",     
          26: "fiche a sé nóiméad",       
          27: "fiche a seacht nóiméad",   
          28: "fiche a hocht nóiméad",    
          29: "fiche a naoi nóiméad"    

Adding the "a" in the middle is the issue, i wouldn't say "fiche a dó", I would say "fiche dó". Let me know what you think of that

AlanGreene: I would be the same and not actually say the `a`, I think it's just a matter of pronunciation / spoken vs. written form, since the `fichE A …` is essentially combined…

Shouldn't `fiche a ceithre` be `fiche a ceathair` though?

I also wonder how much dialect plays into this, I opened a related discussion last night and would be interested in your thoughts: https://github.com/OHF-Voice/intents/discussions/3238

responses/ga/HassGetCurrentTime.yaml
AlanGreene: Fair enough. I suspect it may be a regional / dialect thing anyway (I have fairly mixed Irish…), so at least being self-consistent is good 👍

responses/ga/HassGetCurrentTime.yaml
GGamerzz: > I would be the same and not actually say the `a`, I think it's just a matter of pronunciation / spoken vs. written form, since the `fichE A …` is essentially combined…
> 
I believe that is one of the issues, there is the official way to say it but then there is the unofficial way that is more widely adopted but then you have the different dialects with often their own official ways and unofficial ways! If we got a group discussion on this going, I don't think anyone would agree on much!!!

> Shouldn't `fiche a ceithre` be `fiche a ceathair` though?
> 
Missed that, was using it for counting and forgot this was for telling time

> I also wonder how much dialect plays into this, I opened a related discussion last night and would be interested in your thoughts: #3238
>
I'll look now!

responses/ga/HassGetCurrentTime.yaml
AlanGreene: Just a point of reference: https://www.focloir.ie/en/dictionary/ei/one#one__4:~:text=twenty%20to%20one-,fiche%20chun%20a%20haon,-%2C%20fiche%20go

Foras na Gaeilge omit `nóiméad` in the examples given.

GGamerzz: I think I will stick with `nóiméad`, it does/would sound better in a conversation from my thinking and use of it.

responses/ga/HassGetCurrentTime.yaml
GGamerzz: Good spot! No clue how I missed that! Add that back in now
Does this look good for the template?
`"Tá sé dhá nóiméad tar éis a haon san oíche.",`

GGamerzz: Here is another conflict, there are 4 time ranges,
```
`Morning (Maidin)
00:00–11:59 → **ar maidin**

Afternoon (Iarnóin)
12:00–17:59 → **san iarnóin**

Evening (Tráthnóna)
18:00–21:59 → **sa tráthnóna**

Night (Oíche)
22:00–23:59 → **san oíche**`
```

I don't really use Iarnóin in my day to day usage

GGamerzz:
I am going to update my current code so far, what i've changed and updated for you to have a look, any other issues that are outstanding or come up, we can work on that too

responses/ga/HassGetCurrentTime.yaml
AlanGreene: Yeah I'd use `tráthnóna` to cover the whole period between noon and night. Rarely if ever use `iarnóin`.

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,105 @@ language: ga
...
+          21: "a naoi",
+          22: "a deich",
+          23: "a haon déag"
+        } %}
+
+        {% set minutes_map = {
+          1: "nóiméad amháin",
+          2: "dhá nóiméad",
+          3: "trí nóiméad",
+          4: "ceathair nóiméad",
```
AlanGreene:
```suggestion
          4: "ceithre nóiméad",
```

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,105 @@ language: ga
...
+          5: "cúig nóiméad",
+          6: "sé nóiméad",
+          7: "seacht nóiméad",
+          8: "ocht nóiméad",
+          9: "naoi nóiméad",
+          10: "deich nóiméad",
+          11: "aon nóiméad déag",
+          12: "dhá nóiméad déag",
+          13: "trí nóiméad déag",
+          14: "ceathair nóiméad déag",
```
AlanGreene:
```suggestion
          14: "ceithre nóiméad déag",
```

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,105 @@ language: ga
...
   intents:
     HassGetCurrentTime:
-      default: "TODO: {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
+      default: >
+        {% if slots.time is defined %}
+          {% set hour = slots.time.hour %}
+          {% set minute = slots.time.minute %}
+        {% elif context['time.hour'] is defined %}
+          {% set hour = context['time.hour'] | int %}
+          {% set minute = context['time.minute'] | int %}
```
AlanGreene: Looking at the files for other languages it appears safe to assume that both `slots.time.hour` and `slots.time.minute` are defined. This can be reduced to just declaring the `hour` and `minutes` variables directly unless there was a particular reason for the `context` fallback here.

```suggestion
        {% set hour = slots.time.hour %}
        {% set minute = slots.time.minute %}
```

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,105 @@ language: ga
...
+        {% else %}
+          {% set hour = 1 %}
+          {% set minute = 2 %}
```
AlanGreene: I'm not sure I understand what this is for. It seems odd to set an arbitrary time, and given my above comment about the hour and minute appearing to always be defined this shouldn't be necessary.

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,105 @@ language: ga
...
+        {% endif %}
+        {% set next_hour = (hour + 1) % 24 %}
+        {% set is_default = (hour == 1 and minute == 2) %}
```
AlanGreene: same as above, not sure why this is needed

AlanGreene (changes requested):
Hey @GGamerzz, not sure if this is ready for another review but I left a few comments / questions anyway. Apologies if I jumped the gun.

responses/ga/HassGetCurrentTime.yaml
GGamerzz: I believe this was me trying to get the tests working and I should've removed before pull request

responses/ga/HassGetCurrentTime.yaml
GGamerzz: same as above for me, shoulda been removed after testing

GGamerzz:
Not sure why it is not building, will have a look during the week, if you spot the issue let me know and ill get to it

responses/ga/HassGetCurrentTime.yaml
```diff
@@ -2,11 +2,95 @@ language: ga
...
+          25: "fiche cúig nóiméad",
+          26: "fiche sé nóiméad",
+          27: "fiche seacht nóiméad",
+          28: "fiche ocht nóiméad",
+          29: "fiche naoi nóiméad"
+        } %}
+
+        {% set time_suffix = "" %}
+        {% if is_default %}
+           {% set time_suffix = " san oíche" %}
```
AlanGreene: This needs to be removed too

> Error:  responses/ga/HassGetCurrentTime.yaml: 'is_default' is undefined in response 'default' (template='{% set hour = slots.time.hour %} {% set minute = slots.time.minute %}

PR #3241, opened 2025-07-03
[MN] Create weather_HassGetWeather.yaml

chuk-a:

PR #3243, opened 2025-07-03
[MN] Update climate_HassClimateGetTemperature.yaml

chuk-a:


staticdev:
@chuk-a thanks for the pull requests. If possible alwaya identify the language you are working on with [MN] in the beginning of the name of them. We do this so the right people will look at it, since we have many languages.

chuk-a (commented):
please check

synesthesiam:
Thanks for the effort here, @chuk-a. Unfortunately this PR has several issues that make it unmergeable as-is, so I'm closing it — please feel free to open a fresh, focused PR:

1. **Mixed-in Dutch text.** Several of the new `climate_HassClimateGetTemperature` sentences contain untranslated Dutch copied from the `nl` intents, e.g. `"Wat is [de] [huidige] temperatuur [van|in] [de] <name>"`, `"Хэр <дулаан> is het [<in>] <area>"`, and `"(hoe [өндөр|бага|<warm>]|хэдэн хэм) ... [de] <name>"`. These need to be fully Mongolian.
2. **Destroys `tests/mn/_fixtures.yaml`.** The change replaces the existing mn areas/entities (kitchen, bedroom, lights, fans, etc.) with only weather entities, which breaks every other mn test that relies on them. Fixtures should be added to, not replaced.
3. **Stray files.** `tests/mn/responses_mn_HassGetWeather_Version2.yaml` and `tests/mn/tests_mn__fixtures_Version2.yaml` look like accidental web-upload artifacts and aren't valid test files.

The Mongolian `HassGetWeather` weather intent is being added separately in #3241. For a clean climate `HassClimateGetTemperature` contribution, please add Mongolian-only sentences plus matching tests, and append (not replace) any needed fixtures. Thanks!

PR #3251, opened 2025-07-08
[DE] Added some fill-words for timer status requests

hecktech27:


andreasbrett:
Could you please add test for these new variations?

PR #3253, opened 2025-07-09
[DE] Improve HassClimateSetTemperature intent

freeDom-:
Improve HassClimateSetTemperature intent for de by
- adding the option to say either "Temperatur" or "Heizung"
- adding a new sentence that is area context sensitive and gives the option to use phrases such as "Wohnzimmerheizung" or "Heizung im Wohnzimmer"


sentences/de/climate_HassClimateSetTemperature.yaml
```diff
@@ -3,6 +3,7 @@ intents:
   HassClimateSetTemperature:
     data:
       - sentences:
-          - "<setzen>[ <area>] [die ]Temperatur auf <temperature>"
-          - "<setzen> [die ]Temperatur <von_dem> <name> auf <temperature>"
-          - "<setzen> [die ]Temperatur <name>[s] auf <temperature>"
+          - "<setzen>[ <area>] [die ](Heizung|Temperatur) auf <temperature>"
+          - "<setzen> [die ](Heizung|Temperatur) <von_dem> <name> auf <temperature>"
+          - "<setzen> [die ](Heizung|Temperatur) <name>[s] auf <temperature>"
+          - "[die] (<heizung> <area>|{area}<heizung>) auf <temperature> <setzen>"
```
andreasbrett: You reference a non-existing expansion rule `<heizung>`. What was the idea behind that? I assume you simply meant the word `Heizung`?

Furthermore, I think allowing addressing a device through `{area}heizung` can introduce issues with TTS, although it is common to combine words in German. Let me test that with the Nabu Casa TTS (based on Azure) as that is a good reference point as base TTS system for HA.

mib1185: `<heizung>` is not a defined expansion rule, so this sentence wouldn't work at all

mib1185 (changes requested):
we should pack `(Heizung|Temperatur)` into an expansion rule and maybe also extend with `Klima[anlage|tisierung]` as a `climate` entity can also represent an air condition system. Further please also add test sentences so ensure that these additions works as intended

sentences/de/climate_HassClimateSetTemperature.yaml
mib1185: you were faster :grimacing:

freeDom-: Ahh sorry. I am already using these sentences locally and indeed already created an expansion rule for heizung, which is (Heizung|Temperatur) and did not properly replaced it in my PR... I assume the  right place for this expansion would be the _common.yaml file, right? In my local version, I just added the expansion into the climate_HassClimateSetTemperature.yaml.
I can add this expansion rule.

I tested the phrases Wohnzimmerheizung, Kinderzimmerheizung and Schlafzimmerheizung on a Home Assistant Voice PE using Nabu Case Cloud. I will also add test sentences.

mib1185: > I just added the expansion into the climate_HassClimateSetTemperature.yaml.

if the expansion rule is only used here, then this is the correct place for them, else (_when used in multiple sentence files_) it should go into the common.yaml

andreasbrett: > I tested the phrases Wohnzimmerheizung, Kinderzimmerheizung and Schlafzimmerheizung on a Home Assistant Voice PE using Nabu Case Cloud. I will also add test sentences.

For me it is a hit and miss of 60/40 if I get "Wohnzimmerheizung" or "Wohnzimmer. Heizung."

mib1185: I think this should be solved by adding proper aliases to the entities :thinking:

freeDom-: > I think this should be solved by adding proper aliases to the entities 🤔

Using aliases works only if you have one thermostat in an area, but if you have multiple heaters with individual thermostats aliases are not working here. The goal is to set the temperature for an area, meaning all thermostat entites in that area.

freeDom-: > > I just added the expansion into the climate_HassClimateSetTemperature.yaml.
> 
> if the expansion rule is only used here, then this is the correct place for them, else (_when used in multiple sentence files_) it should go into the common.yaml

~~It looks like HA is not allowing expansion rules within the sentence file. Should I add it to the _common expansion rules then? When the yaml is put into a config/custom_sentences/de/set_temperature.yaml file it is working though  🤔~~

Nevermind, had to fix the indentation to make it work.

andreasbrett: I tested using Nabu Casa TTS and it tends to vary between `{area}heizung` and `{area} Heizung` -> we should reflect that in the sentences. Good idea, @freeDom-!

freeDom-: OK, I will also do more testing and can add both sentences for better performance!

sentences/de/climate_HassClimateSetTemperature.yaml
```diff
@@ -3,8 +3,13 @@ intents:
...
+          - "<setzen>[ <area>] <heizung> auf <temperature>"
+          - "<setzen> <heizung> <von_dem> <name> auf <temperature>"
+          - "<setzen> <heizung> <name>[s] auf <temperature>"
+          - "<heizung>[ (<area>|[<von_dem> ]<name>)] auf <temperature>[ <setzen>]"
           - "<area> auf <temperature> [be]heizen"
+          - "<setzen> <area_heizung> auf <temperature>"
+          - "<area_heizung> auf <temperature>[ <setzen>]"
+        expansion_rules:
+          heizung: ([(die|den|das) ](temperatur|heizung|thermostat|klima[anlage|tisierung]))
+          area_heizung: ([(die|den|das) ]<area><heizung>)
```
andreasbrett: To not introduce unnecessary permutations (like `den temperatur` or `das klimatisierung`), let's use the following.

```suggestion
          heizung: ([die ]temperatur|[die ]heizung|[(das|den) ]thermostat|[die ]klima(anlage|tisierung)|[das ]klima)
          area_heizung: ([die ]<area>temperatur|[die ]<area>heizung|[(das|den) ]<area>thermostat|[die ]<area>klima(anlage|tisierung)|[das ]<area>klima)
```

andreasbrett (changes requested):
Apart from the mentioned improvements on permutations, this looks good.

sentences/de/climate_HassClimateSetTemperature.yaml
andreasbrett: Please also add some (not all) tests for the added space (`das wohnzimmer thermostat...`)

sentences/de/climate_HassClimateSetTemperature.yaml
freeDom-: Added sentences with white space  between area and <heizung> and replaced expansion_rule for <area> with {area} to further remove unneccessary permutations

andreasbrett (approved):
This looks good! Thank you very much for helping us out and improving Assist.

PR #3254, opened 2025-07-09
Added some more Timer-controls

hecktech27:
some more timer status / control variations


andreasbrett:
As mentioned in https://github.com/OHF-Voice/intents/pull/3256 I don't particularly like omitting the "timer" part in prompts (e.g. ``wie lange dauert es noch` in HassTimerStatus). The other additions though are good in my book.

I like the added verbs for canceling 1 or all timers but they should a) be turned into an expansion rule (see `sentences/de/_common.yaml`) and b) all verbs should be used everywhere (HassCancelAllTimers has more verbs than HassCancelTimer) to improve user experience. It shouldn't matter if I want to cancel 1 or all timers - "deaktivieren" should work for both scenarios.

hecktech27:
I also removed the variations without the "Timer" word

mib1185 (approved):

PR #3256, opened 2025-07-09
Add possibility to start Timer only by telling assist the required duration

hecktech27:
I added the possibility to start Timers by only telling assist the required duration of the timer.
Since I was not sure whether something like "5 Minuten" (which I use on a regular basis) would be enough to assign the command globally to the timer-controls-category I put this one in an extra pull request.


hecktech27:
would this also be the case for "5 Minuten für Pizza" or would this be specific enough?

hecktech27:
I can absolutely see the possibility of this one interfering with other (current or future) intents - thats exactly why I wanted to discuss this separately. Thanks all!

PR #3261, opened 2025-07-13
[DE] added some more date requests

hecktech27:


andreasbrett (approved):
Solid as well. Please consider though for future PRs to add the `[DE]` prefix to the title. That way the release notes look much better.

hecktech27:
> Solid as well. Please consider though for future PRs to add the `[DE]` prefix to the title. That way the release notes look much better. 

Well that's an easy one ;)

PR #3263, opened 2025-07-13
added some more person location request variations

hecktech27:


andreasbrett (changes requested):
Makes the sentences much more natural! Could you please add `gerade` and `momentan` and `im Moment` as further variations?

andreasbrett:
> would you like to have it added anyways

No, you're right. No need to have it in there. Even if it might get removed from skip words, if we now add it, it could cause confusion why it is not working.

PR #3264, opened 2025-07-13
[DE] added variations for cover state requests

hecktech27:


andreasbrett (changes requested):
By defining derzeit/aktuell twice as optional components, you create multiple unnecessary permutations. We should avoid these to not introduce unnecessary load to intent matching performance (we had to reduce permutations a while ago because it impacted slower HA machines)

Could you please use separate sentences for derzeit/aktuell before and after the cover name?

hecktech27:
maybe wait with this one - I would modify the code once again to also use the expansion rule for `<aktuell>` from #3263 once it is merged

andreasbrett (approved):
I'm running out of variations of "thank you" 😀

mib1185:
> I'm running out of variations of "thank you" 😀

https://giphy.com/search/thank-you :wink:

PR #3266, opened 2025-07-14
[DE] Added even more Timer controls

hecktech27:
i added variations for increasing, decreasing pausing and unpausing timers.
for decreasing timers i moved(and extended) the verbs to an expansion rule.
From PauseTimer i removed "pausiere meinen timer für x minuten" since this would imply that the timer gets resumed after x minutes which i don´t think is correct in this context.


andreasbrett (approved):
Thank you!

PR #3267, opened 2025-07-16
[DE] additions to lock state requests

hecktech27:


andreasbrett (approved):
👍

sentences/de/lock_HassGetState.yaml
```diff
@@ -3,43 +3,67 @@ intents:
   HassGetState:
     data:
       - sentences:
-          - "ist <name> {lock_states:state}[ <area>]"
-          - "ist <name>[ <area>] {lock_states:state}"
-          - "ist[ <area>] <name> {lock_states:state}"
+          - "ist[ <aktuell>] <name> {lock_states:state}[ <area>]"
+          - "ist <name>[ <aktuell>] {lock_states:state}[ <area>]"
```
mib1185, line 7:
```suggestion
          - "ist <name> <aktuell> {lock_states:state}[ <area>]"
```
in the second sentence, when the former optional words just change the position, it should not be optiional anymore, as it covers some variations from that one above

eq. "ist name geschlossen" would be matched by both sentences

hecktech27: you´re absolutely right! I will go through my changes of the last days and remove other occurences of these combinations - I´m afraid I might have implemented one or two of these..
Since this PR is already closed I will open another one where your suggestion is resolved as well as possible others!

hecktech27: @mib1185 this should be covered by #3268

PR #3276, opened 2025-07-21
[DE] Additional variations for MediaPlayer pause / unpause

hecktech27:


sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
   HassMediaPause:
     data:
       - sentences:
-          - "(pausiere|pause) <name>"
-          - "halte <name> an"
-          - "<name> (pausieren|pause|anhalten|stopp|halt)"
+          - "(pausiere|pause|stop[p][e]) <name>"
```
andreasbrett: This would allow stop, stopp, stoppe but also __stope__.

`stop[p[e]]` is hard to read, so I propose to go stupidly simple with just `stop|stopp|stoppe`.

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
+          - "halt[e] <name> an"
+          - "<name> (pausieren|pause|anhalten|stop[p][en]|halt)"
```
andreasbrett: Same here. We should not allow stopen.

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
         requires_context:
           domain: media_player
       - sentences:
-          - "(pausiere[n]|pause|anhalten|stopp|halt[e an])"
+          - "(pausiere[n]|pause|anhalten|stop[p][en]|halt[(e an| an)])"
```
andreasbrett: Same here with stopen.

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
+          - "pausiere[ (die|das|mein|meine)] <media_type>"
+          - "(stopp[e]|halt[e])[ (die|das|mein|meine)] <media_type>[ an]"
+          - "[(die|das|mein|meine) ]<media_type> (pausieren|pause|anhalten|stop[p][en]|aus[schalten])"
```
andreasbrett: Stopen once more

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
         requires_context:
           area:
             slot: true
       - sentences:
           - "pausiere[ (die|das|mein|meine)] <media_type> <area>"
+          - "(stopp[e]|halt[e])[ (die|das|mein|meine)] <media_type> <area>[ an]"
+          - "[(die|das|mein|meine) ]<media_type> <area> (pausieren|pause|anhalten|stop[p][en]|aus[schalten])"
```
andreasbrett: Stopen once more

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
+          - "[(die|das|mein|meine) ]<media_type> (pausieren|pause|anhalten|stop[p][en]|aus[schalten]) <area>"
+          - "<area> [(die|das|mein|meine) ]<media_type> (pausieren|pause|anhalten|stop[p][en]|aus[schalten])"
```
andreasbrett: Stopen once more

sentences/de/media_player_HassMediaPause.yaml
```diff
@@ -3,15 +3,25 @@ intents:
...
+      - sentences:
+          - "<area> (pausieren|pause|anhalten|stop[p][en]|halt)"
```
andreasbrett: Stopen once more

sentences/de/media_player_HassMediaUnpause.yaml
```diff
@@ -3,14 +3,25 @@ intents:
   HassMediaUnpause:
     data:
       - sentences:
-          - "<name> weiter[ (<machen>|laufen lassen)]"
+          - "[(am|auf) ]<name> weiter[ ][(<machen>|laufen lassen|hören|[an]schauen)|[ab]spielen]"
```
andreasbrett: Why the `[ ]` and why did you not put `[ab]spielen` into the first parenthesis group?

sentences/de/media_player_HassMediaUnpause.yaml
```diff
@@ -3,14 +3,25 @@ intents:
...
           - "<name> fortsetzen"
+          - "<name> wieder (starten|[ab]spielen|laufen lassen)"
         requires_context:
           domain: media_player
       - sentences:
-          - "fortsetzen|weiter[ (<machen>|laufen lassen)]"
+          - "[[(die|das|mein|meine) ]<media_type> ](fortsetzen|weiter[ ][(<machen>|laufen lassen|[an]hören|[ab]spielen|[an]schauen)]|wieder [(laufen lassen|[an]hören|[ab]spielen|[an]schauen)])"
```
andreasbrett: That is very hard to read now. Can this be made easier? E.g. put all the options into an expansion list (local if it is only used here).

andreasbrett (changes requested):

sentences/de/media_player_HassMediaPause.yaml
mib1185: or `stop[p]|stoppe` :grimacing:

hecktech27: well - you´re obiously right with this...

sentences/de/media_player_HassMediaUnpause.yaml
hecktech27: abspielen was by mistake.
`weiter[ ][(<machen>|laufen lassen|hören|[an]schauen)|[ab]spielen]`
the `[ ]` is there to e.g. allow for "weitermachen" as well as "weiter machen" - depending on what the ouput of the StT-Engine is so both get recognized. But maybe this is just some edge case - i don´t know... what do you think?

andreasbrett: Got it, thank you. Unfortunately that will also allow `weiteranschauen` and `weiterabspielen`. I am more a fan of splitting it into multiple sentences when complexity of the regex passes a certain threshold. I think this is the case here with so many verbs. Sorry for being so picky but I think we should also consider making this more readable and maintainable. I have a hard time figuring out all the options and seeing if everything is covered by tests.

hecktech27: absolutely no problem! i will adjust it accordingly - i get your point ;)
by the way... is there a possibility to define an expansion rule for the whole file and not only per 
`- sentences:` without moving it to the commons.yml?

andreasbrett: I don't think so but I haven't tried yet.

hecktech27: i tried - and failed. the result is in MediaUnpause of this PR...

tests/de/media_player_HassMediaUnpause.yaml
```diff
@@ -26,6 +58,24 @@ tests:
...
+      - "Meine Musik im Wohnzimmer weiter abspielen"
+      - "Lass meine Musik im Wohnzimmer weiter laufen"
+      - "Lass meine Musik im Wohnzimmer wieder starten"
+      - "Lass die Fernsehsendung im Wohnzimmer weiter laufen"
+      - "Lass die Fernsehsendung im Wohnzimmer wieder laufen"
+      - "Lass meine Musik im Wohnzimmer weiterlaufen"
+      - "Meine Musik im Wohnzimmer weiterlaufen lassen"
       - "Meine Fernsehsendung im Wohnzimmer fortsetzen"
       - "Die Fernsehsendung im Wohnzimmer weiter machen"
       - "Das Gerät im Wohnzimmer fortsetzen"
```
hecktech27, line 81: while making changes to this file i thought about this one...
Any idea in which context "Gerät" for <media_type> would be correct?

andreasbrett: While it is a bit weird for `fortsetzen` it does sound way more natural when using it for the other verbs. Tbh I don't think I've ever used or thought about `unpausing` in German. For me unpause is just play.

andreasbrett: To answer your question: `Mein Gerät im Wohnzimmer weiterlaufen lassen` is a context I see myself using (rarely).

PR #3285, opened 2025-07-24
[DE] Reduce unnecessary variations to lower overall sentence count

hecktech27:
After adding quite a few sentence variations during the last Pull Requests this one reduces overall sentence count to ~164 Million Sentences from previously ~247 million.
I did not edit any test sentence files - all previously covered test sentences are still covered by my optimizations.
This is mainly achieved by
- splitting expansion rules into beginning and end_of_sentence rules tu reduce unnecessary permutations
- removing "Momentan" and "im Moment" from the `<aktuell>` expansion rule since these words are on the `skip_words`-List... my bad. Maybe we should move "aktuell" and "derzeit" to the skip_words-list as well...
- removing expansion rules from sentences where only one option of the expansion rule would make sense


sentences/de/light_HassTurnOff.yaml
andreasbrett, line 1: We should combine `ausschalten` and `ausmachen`

sentences/de/light_HassTurnOn.yaml
andreasbrett, line 1: We should combine `anschalten` and `anmachen`

andreasbrett (approved):
Awesome, great work! Simple yet effective. This will be very helpful for lower power machines.

Since the impact can be rather high, I'd like to have approval from @mib1185 and/or @easterapps as well. To me this looks good.

andreasbrett:
"Maybe we should move "aktuell" and "derzeit" to the skip_words-list as well..."

Yes, I think that makes sense. And BTW sorry for leading you onto that path of adding `momentan` and `im Moment`.

sentences/de/light_HassTurnOff.yaml
hecktech27: done...

sentences/de/light_HassTurnOn.yaml
hecktech27: done as well... :)

PR #3289, opened 2025-07-25
[DE] Reduce overall sentence count by optimizing <floor> expansion rule

hecktech27:
This PR reduces overall sentence count by ~ 6 million by optimizing the expansion rule without losing tested sentences.
In general unnecessary permutations like

- auf vom FLOOR
- in im FLOOR

etc. are removed.


andreasbrett (approved):
LGTM

PR #3293, opened 2025-07-27
[DE] improve responses

andreasbrett:
Make responses more natural and less robot-like. For `HassTurnOn`/`HassTurnOff` describe what exactly has been turned on/off where possible (I always disliked when starting a script or scene that only the name would be given `Gemütlich aktiviert` vs `Szene Gemütlich aktiviert`).

I further propose to (in another PR) make the responses even more natural (while keeping them short for acceptable TTS response times) by:
- adding dynamic responses (see https://github.com/OHF-Voice/intents/pull/3290)
- using richer sentences (`Ventil geschlossen` -> `Das Ventil wurde geschlossen` or `Das Ventil wird geschlossen`)


responses/de/HassMediaPause.yaml
```diff
@@ -2,4 +2,4 @@ language: de
 responses:
   intents:
     HassMediaPause:
-      default: "Pausiert"
+      default: "Medien werden pausiert"
```
hecktech27: or maybe `Wiedergabe wird pausiert`?

andreasbrett: Or `Wiedergabe wird angehalten`?

hecktech27: good as well - i would really appreciate a possibility to use your dynamic responses. Would make the whole thing more natural (and nobody would have to decide what the best answer is as there could be multiple best answers 😉)

andreasbrett: Well then let's start with the first dynamic responses in this PR already. It's not that complicated that it requires a dedicated PR.

hecktech27: thats indeed way less complicated than i thought

andreasbrett:
@mib1185 @easterapps don't want to rush but if you have time, I'd love to see this in the 2028.8 release.

PR #3294, opened 2025-07-29
[DE] Major rework of HassLightSet

hecktech27:
oh well - this one escalated a little bit... to be honest i just wanted to optimize some sentences in HassLightSet but one optimization led to the next and here we are...
While I really tried to make sure not to lose any previously working sentences please really take your time to review this - since it ended up to be a complete rework of the file!

**What´s in this PR:**
- optimizations of expansion rules in the "light"-context
- a new expansion rule for `<hier>` which I´d like to apply to other files in the future for a consistent user experience
- I introduced `<floor>` to every place needed in HassLightSet so you can now set lights by floor
- I applied `{brightness_level:brightness}` to all variations of brightness controls so you can now control lights by "xy auf maximale Helligkeit" - context-aware so either name, satellite area, area or floor as a target
- I introduced "hell" as a comand( also context-aware) so you can now quickly turn on lights by a command "Schlafzimmer hell" and all lights in Schlafzimmer will be on 100% regardless of the previously configured brightness
- I tried to create a user-experience as consistent as possible so all sentences working for an area should be working for a floor as well as for the satellite´s area
- I removed previously possible but not working slot combinations so hopefully no more "Es ist ein Fehler aufgetreten".

- I **>>removed<<** `<alle_lichter>` from all sentences where a slot-area is used since this allowed for sentences like

> Alle Lichter auf 10 Prozent.

I decided to remove this option to be consistent with LightTurnOn/LightTurnOff where `<alle_lichter>` turns off all lights everywhere(which is the reason for the only deleted test sentences). As far as I know there is no possibility to change all lights brightness/colors(please correct me if wrong!).
While i removed `<alle_lichter>` from "satellite-area / slot: area"-sentences there is still `<lichter>` since this does not imply that all lights everywhere are changed. So "Lichter auf 10% dimmen" still works.
Happy to discuss this decision! 😊

to make sure i do not lose any previously working sentences i made extensive use of test-sentences for this file.

And the cherry on top:
all the added functionality led to a drop in overall sentence count by another 5 million 😇


sentences/de/_common.yaml
```diff
@@ -574,11 +578,17 @@ expansion_rules:
...
-  alle_lichter: "(<alle> (<lichter>|Beleuchtung)|von allen[ (Lichtern|Lampen|Leuchten|Beleuchtungen)])"
+  licht: "([das ]Licht|[die ]Lampe|[die ]Beleuchtung)"
+  licht_ohne_artikel: "(Licht|Lampe|Beleuchtung)"
+  lichtes: "([des ](Licht[s]|Lichtes)|[der ](Lampe|Beleuchtung))"
+  lichtes_mit_artikel: "(des (Licht[s]|Lichtes)|der (Lampe|Beleuchtung))"
+  lichter: "([(die|der) ]Lichter|[von den ]Lichtern|[(die|der|von den) ](Lampen|Leuchten|Beleuchtungen))"
+  #lichter: "[(die|der|von den) ](Lichter|Lichtern|Lampen|Leuchten|Beleuchtungen)"
+  #alle: "((alle|sämtliche)[r]|jede[r|s|n]|(die|der) (ganze|komplette|sämtliche)[n])[ der]"
+  alle_lichter: "(<alle> ((Lichter|Lampen|Leuchten|Beleuchtungen)|Beleuchtung)|von allen[ (Lichtern|Lampen|Leuchten|Beleuchtungen)])"
+  #alle_lichter: "(<alle> (<lichter>|Beleuchtung)|von allen[ (Lichtern|Lampen|Leuchten|Beleuchtungen)])"
```
andreasbrett: Why keep the commented lines?

hecktech27: ...not on purpose 😉
I usually keep lines i change as a comment and remove them before the PR. Just forgot the commons file

hecktech27:
> I decided to remove this option to be consistent with LightTurnOn/LightTurnOff where `<alle_lichter>` turns off all lights everywhere(which is the reason for the only deleted test sentences). As far as I know there is no possibility to change all lights brightness/colors(please correct me if wrong!).
> While i removed `<alle_lichter>` from "satellite-area / slot: area"-sentences there is still `<lichter>` since this does not imply that all lights everywhere are changed. So "Lichter auf 10% dimmen" still works.
> Happy to discuss this decision! 😊

I "discussed" this with myself again and i think maybe support for `<alle_lichter>` in combination with a non-optional `<hier>` would be best, since this would allow sentences like
> Alle Lichter in diesem Raum auf 10%

to be recognized and the result would be exactly as expected as well as consistent with HassTurnOn/Off behavior.
An optional `hier` would allow 
> Alle Lichter auf 10 Prozent

and(when a slot context area is applied) result in the same as above whereas the same with `...ausschalten` results in all lights everywhere turned off.

sentences/de/_common.yaml
```diff
@@ -340,10 +340,14 @@ lists:
...
     values:
-      - in: (max[imal[e]|imum]|höchste|volle)
+      - in: (max[imal[e]|imum]|höchste|volle|grö(ss|ß)te|hell[ste])
         out: 100
-      - in: (min[imal[e]|imum]|niedrigste)
+      - in: (min[imal[e]|imum]|niedrigste|kleinste)
         out: 1
+  brightness_level_max:
+    values:
+      - in: hell[ste]
```
andreasbrett, line 349: Why only these 2 options and not the full set from <brightness_level>?

andreasbrett (commented):
I'll leave this one for @mib1185 and @easterapps. My motivation for reviewing this monster of a PR is very low. I never change a light via Assist directly and only ever use scenes.

mib1185 (commented):
Hi @hecktech27 
Many thanks for all your contributions in the last few weeks, that's awesome ❤️ 
Can we split this PR into smaller, independent chunks? This would make it much better reviewable.

mib1185 (changes requested):
Accidentally clicked on comment instead of request change 🙈

hecktech27:
I will have a Look at it tomorrow if i have any good ideas on how to split this up - currently only on my mobile... I might need some input from you on what would work for you @mib1185

hecktech27:
> > I decided to remove this option to be consistent with LightTurnOn/LightTurnOff where `<alle_lichter>` turns off all lights everywhere(which is the reason for the only deleted test sentences). As far as I know there is no possibility to change all lights brightness/colors(please correct me if wrong!).
> 
> > While i removed `<alle_lichter>` from "satellite-area / slot: area"-sentences there is still `<lichter>` since this does not imply that all lights everywhere are changed. So "Lichter auf 10% dimmen" still works.
> 
> > Happy to discuss this decision! 😊
> 
> 
> 
> I "discussed" this with myself again and i think maybe support for `<alle_lichter>` in combination with a non-optional `<hier>` would be best, since this would allow sentences like
> 
> > Alle Lichter in diesem Raum auf 10%
> 
> 
> 
> to be recognized and the result would be exactly as expected as well as consistent with HassTurnOn/Off behavior.
> 
> An optional `hier` would allow 
> 
> > Alle Lichter auf 10 Prozent
> 
> 
> 
> and(when a slot context area is applied) result in the same as above whereas the same with `...ausschalten` results in all lights everywhere turned off.


@mib1185 I would really appreciate your opinion on this one so when I split this PR maybe we can skip one iteration of corrections

Thanks and sorry for creating such a big PR - as i wrote above it was not planend for it to become that big but one thing led to the next.... I absolutely see why this is difficult to review!

sentences/de/_common.yaml
hecktech27: my intention for this one was to use it in a context like 
> Wohnzimmer hell

and since brightness_level would then allow 
> Wohnzimmer volle
> Wohnzimmer höchste
> Wohnzimmer größte
> ...

i decided to use a smaller set of words for this special purpose...
But maybe i already have a better idea - i will add "Stufe" to the list from 'brightness_level' where needed. This will allow us to use all variations from brightness_level but without the unnecessary permutations.

hecktech27:
@mib1185 (@andreasbrett)
i created #3314 as a PR and #3315 as a draft.
would it work for you to iterate through the changes of this PR like i started in the other two PRs?

sentences/de/_common.yaml
hecktech27: #3314 
the new version got a little longer but covers all useful variations from `{brightness_level}` in the intended context.

PR #3296, opened 2025-07-29
[DE] further optimization of expansion rules

hecktech27:
reduces overall sentence count by another ~7.7 million down to currently ~81 million


sentences/de/_common.yaml
```diff
@@ -580,16 +580,16 @@ expansion_rules:
   alle_lichter: "(<alle> (<lichter>|Beleuchtung)|von allen[ (Lichtern|Lampen|Leuchten|Beleuchtungen)])"
   leuchten_lassen: "([er]leuchten lassen|[ein]färben)"
   tuer: "[die ]Tür[e|en]"
-  schloss: "[(das|die) ](Schloss|Schlösser)"
+  schloss: "([das ]Schloss|[die ]Schlösser)"
   sperren: "(sperr|schlie(ss|ß))[e|en]"
-  entsperren: "((ent|auf)<sperren>|öffne|entrieg(el|le))[n]"
+  entsperren: "(entsperr[e|en]|aufschlie(ss|ß)[e|en]|öffne[n]|entrieg(el[n]|le))"
```
andreasbrett, line 585: Let's get rid of `aufschließe` (also with `ss`) since no one will say that IMHO. Or is that in there because it is often misinterpreted by STT?

hecktech27: i had deleted it in my first iteration and implemented it again exactly for that reason. but i´m fine with both - up to you

andreasbrett: Let's keep it then.

PR #3300, opened 2025-07-30
[DE] Add more variations to delayed conversation commands

hecktech27:
This PR extends the possible commands for {timer_command:conversation_command}.
It is now possible to use sentences like 
> schalte in einer Stunde das Licht im Wohnzimmer aus

instead of 
>In einer Stunde das Licht im Wohnzimmer ausschalten

This is currently limited to commands from the `(schalten|machen|fahren)`-group since for example `öffne` would allow sentences like
> Öffne in einer Stunde die Fenster

which would result in `die Fenster` as the conversation command.


andreasbrett (approved):
Good one. I really how this makes the prompt more natural!

hecktech27:
would you see a possibility to handle something like "split commands"?
for example
> öffne in einer Stunde das Fenster

currently is (more or less - not real working code)
`öffne {timer_duration} {conversation_command}`
is thare a possibility to add `öffne ` in the beginning of `{conversation_command}`?
i would still have to think about whether(or probably what kind of) this introduces any issues...

andreasbrett:
No, unfortunately. The implementation of delayed commands is not that sophisticated and very simplistic.

hecktech27:
While i had hoped for the opposite i had a feeling this was (currently) not possible.
Thanks for the info 😉

PR #3306, opened 2025-07-31
[DE] Addition to weather requests

hecktech27:


andreasbrett (approved):
Well spotted and nice addition.

PR #3309, opened 2025-08-01
[DE] Fan controls: add support for floors and satellite area while reducing overall sentence count

hecktech27:
This PR adds support for turning on/off fans per floor as well as in the corresponding satellite´s area.
I removed test sentences like
> stoppe den Ventilator im Wohnzimmer ab
> starte den Ventilator im Wohnzimmer an
> mach den Ventilator im Wohnzimmer ein

since these sentences make no sense and during a sentence structure optimization were no longer covered.
Support for satellite area and floors results in overall sentence count - ~300.000 😉


andreasbrett (approved):
👍

PR #3310, opened 2025-08-03
[EN] allow hyphens in timer duration intents

heydabop:
Occasionally my speech to text engine (wyoming-faster-whisper) will output a hyphen between the number and unit of a timer duration, i.e. "start a 5-minute timer" vs "start a 5 minute timer" which causes the sentence to be unrecognized. This commit should ideally fix that.

Not sure if this is the best way to go about this (or if I should chase down the issue in wyoming-faster-whisper instead). I saw a similar pattern in Norwegian timer intents. I also wasn't sure how many tests to duplicate, but didn't want to overdo it.


heydabop:
I guess I'd also expect to see this failure on other PRs if it truly started with the aforementioned commit, so I must be missing something...

PR #3314, opened 2025-08-04
[DE] optimization of light-related expansion rules

hecktech27:
This is the first part of multiptle iterations through the HassLightSet file (goal is to achieve all changes from PR #3294 )
This PR introduces the following to the commons file:

- new list `{brightness_level_max}`
- additions to `{brightness_level}`
- additions to `<floor>`
- optimization of light-related expansion rules
- now already covered by #3319:  new expansion rule `<hier>`

All these changes will be needed for the following PRs.
This temporarily reduces overall sentence count by ~24 million, but this reduction will be needed for the coming additional sentences/functions. The final reduction will be ~5 million.


mib1185 (changes requested):
We highly appreciate all your contributions, but we shouldn't do changes just in preparation to any other upcoming PR, when these changes itself doesn't "work" alone. So we should also limit changes in one PR to only handle "one thing". This helps to keep the PRs revieweble, as each PR can be merged on its own.

For instance in this PR here, you're extending some expansion rules, but without adding proper tests as evidence that the additions works as expected, further there are expansion rules added, which are unused.

hecktech27:
As mentioned in the other PR i currently do not have a computer with me - i will look at it again this weekend.
That being said: I totally get what and why you are saying - makes perfect sense. 
I'm only struggling with this one a little bit since i had working code (with admittedly quite complex changes) and now Need to split all those changes again while Not forgetting to move reqired parts into the corresponding PR...
And while i do enjoy it and have been coding for a Long time now this is not something i do as my dayjob so i'm still getting used to how things work in the Open Source world. 
but anyways - I'm pretty sure we'll find a Solution that works for everyone!
(And thanks for dropping the discord link - i wasn't really aware of that!)

hecktech27:
i changed this pr to draft to keep it as a reference for me while implementing these changes made here in new, smaller PRs and will close it once i´m done.

hecktech27:
all implemented in other PRs

PR #3317, opened 2025-08-04
[DE] more responses for timer controls

hecktech27:
thanks to @andreasbrett ´s latest introduction of dynamic responses(#3293) i was able to start expanding these a little bit.
more to come... :)


andreasbrett (approved):
Let's wait for @mib1185 or @easterapps to review since I one-man-rushed the PR with the dynamic responses.

I think it's a great way to make Assist feel more natural/personal while at the same time not introducing a lot of maintenance overhead but this might not be the way others want to go.

hecktech27:
While I fully agree with you on this making Assist much more natural you might be right with your concerns. Lets just keep this PR until this is got little more discussion

hecktech27:
Thanks - I already thought so after reading a little bit on the (lately discovered) discord

PR #3319, opened 2025-08-05
[DE] ClimateGetTemperature: add support for floors and introduce <hier> expansion rule

hecktech27:
This PR adds support for temperature requests based on floors and extends requests in the satellite´s area from only `hier` to
- Hier im Raum
- Hier in diesem Raum
- im Raum
- in diesem Raum

I moved the local expansion rule `<local_hier>` from (for example) ClimateSetTemperature to the commons file and replaced other occurences of the same expansion rule.


hecktech27:
I added `Bereich` as another option to `Raum` in the `<hier>` expansion rule since areas might not only cover single rooms

PR #3321, opened 2025-08-07
[DE] move area-aware sentences to "<hier> expansion rule

hecktech27:
This PR extends area-aware sentences that used `hier` to the `<hier>` expansion rule to cover variations like
- im Raum
- in diesem Raum
- hier im Raum
- in diesem Bereich
- ...


andreasbrett (approved):
Thank you once more!

PR #3324, opened 2025-08-11
Add volume up/down intents

synesthesiam:
Adds English sentences for `HassVolumeUp` and `HassVolumeDown` intents: https://github.com/home-assistant/core/pull/150443

These intents increase or decrease the volume of a media player by a pre-determined amount (see `volume_up` and `volume_down` actions from [`media_player`](https://www.home-assistant.io/integrations/media_player/)


ViViDboarder:
Does the intent also work with floors? It would be good to have that, if so.

sentences/en/media_player_HassVolumeDown.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - sentences:
+          - "[turn [the]] volume down"
+          - "turn down [the] volume"
+          - "decrease [the] volume"
+        requires_context:
+          area:
+            slot: true
+      - sentences:
+          - "(<name>;volume down)"
+          - "decrease [the] volume [on|for] <name>"
```
ViViDboarder: Could also support “turn down” here as well. Probably for all places we use decrease or increase. I think there may be an expansion already.

staticdev (approved):
Very cool, looking forward to testing this one also =)

ViViDboarder (changes requested):

synesthesiam:
Please don't merge yet since I need to refactor due to requested changes in HA

PR #3332, opened 2025-08-12
[DE] revert improved responses

andreasbrett:
reverts https://github.com/OHF-Voice/intents/pull/3293 as discussed in https://discordapp.com/channels/330944238910963714/1381655109083074772


andreasbrett:
It's fine @mib1185. I like the direction this is heading towards and in the end I get what I wanted. Just requires more patience but then again this is implemented professionally and not just a hack.

PR #3334, opened 2025-08-13
Add sentences for media player mute/unmute Intents

anishsane:
This change adds new intents for media player mute/unmute operations.

### DIFF
--- a/intents.yaml
+++ b/intents.yaml
@@ -1223,6 +1223,88 @@ HassSetVolume:
 
 # -----------------------------------------------------------------------------
 
+HassMediaPlayerMute:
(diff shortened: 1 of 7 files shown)


anishsane:
Tagging @arturpragacz and @marcelveldt since they added the corresponding HA core PR to Voice project/next.
This PR is kept in the draft state because I think the handling of the intent should be submitted before this change.

arturpragacz:
Please do not tag people like that.

anishsane:
My apologies.
I intended to tag only for information - since the PR was already out, it might save some duplication of work.
I did not intend to nag.

PR #3356, opened 2025-08-16
[DE] HassSetVolumeRelative: more variations

hecktech27:
This PR adds "Musik" as a command


andreasbrett (approved):
I'm a bit on the edge with this one as prompts become quite long with this addition and while this is quite a natural way of asking for this intent, I feel this is too long for users to ask their assistant.

Might just be a "me thing" though, as I keep prompts short to get to my intended result ASAP.

PR #3361, opened 2025-08-17
document count_sentences in README

andreasbrett:
I always forget how to use `count_sentences` and think this should also be promoted better in `README` as not every language leader is aware that keeping an eye on permutations is beneficial.


synesthesiam (approved):

synesthesiam:
Thanks!

PR #3366, opened 2025-08-17
CI: validate only changed language

andreasbrett:
Speed up CI checks by only validating those languages that changed.

I am not sure whether this is on purpose but currently when only 1 language was changed in a PR, still all languages are run through intentfest validate while only the tests for the changed languages are run. This is because `--language XY` is only passed to `pytest` but not to `python3 -m script.intentfest validate`.

https://github.com/OHF-Voice/intents/blob/cb64a6455f0d036c37fa7498a6628a67d8f49157/.github/workflows/ci.yaml#L68-L76


andreasbrett:
@tetele got 3366... 😎

PR #3374, opened 2025-08-18
[DE] reduce overall sentence count by optimizing the "<alle>" expansion rule

hecktech27:
this PR reduces the OSC by another ~14.2 million down to currently 48.8 million

### DIFF
--- a/sentences/de/_common.yaml
+++ b/sentences/de/_common.yaml
@@ -577,7 +577,6 @@ expansion_rules:
   von_dem: "(von[ dem]|vom)"
   zu_dem: "(zu[ dem]|zum)"
   alle: "((alle|sämtliche)[r| der]|jede[r|s|n][ der]|die (ganze|komplette)|die (ganzen|kompletten)[ der]|der (ganzen|kompletten|sämtlichen)[ der])"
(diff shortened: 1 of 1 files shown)


sentences/de/_common.yaml
```diff
@@ -576,7 +576,7 @@ expansion_rules:
   zu: "(zu|[he]runter|nach unten)"
   von_dem: "(von[ dem]|vom)"
   zu_dem: "(zu[ dem]|zum)"
-  alle: "((alle|sämtliche)[r]|jede[r|s|n]|(die|der) (ganze|komplette|sämtliche)[n])[ der]"
+  alle: "((alle|sämtliche)[r| der]|jede[r|s|n][ der]|die (ganze|komplette)|die (ganzen|kompletten)[ der]|der (ganzen|kompletten|sämtlichen)[ der])"
```
andreasbrett, line 579: Why only `sämtlichen` for the `der` case? What about e.g. `die sämtlichen`?

hecktech27: actually i thought a while about `die sämtlichen` but couldn´t really find any useful, natural context.
it always came down to 
> sämtliche der ...
> ändere die farbe der sämtlichen...

if you prefer to keep this option i can add it again

andreasbrett: If it is a deliberate decision I am fine with that. Just wanted to make sure it wasn't forgotten.

PR #3379, opened 2025-08-19
[DE] optimization of expansion rule <alle_lichter>

hecktech27:
this pr removes an unnecessary bracket from the `<alle_lichter>` expansion rule and reduces osc by ~63.000 by this


andreasbrett (approved):
Good one! BTW you've surpassed 50 merged PRs a while ago. Thank you so much for your valuable contributions.

PR #3381, opened 2025-08-19
[DE] optimization of <irgend> expansion rule to reduce overall sentence count

hecktech27:
This PR further reduces the overall sentence count by ~2 million (currently ~4%).


sentences/de/_common.yaml
```diff
@@ -619,7 +619,7 @@ expansion_rules:
   szene: "[die ]Szene"
   skript: "[das ]Skript"
   batterie: "([der ]Batterie[n]|[des ]Akku[s])"
-  irgend: "(irgend(ein[s]|eine[s]|einer|einem|einen)[ der]|(irgendwelche[s][ (der|die|des)])|(irgendwo|einige[s]|manche[s]|ein[s]|eine[s]|einer|einem|einen) <artikel_bestimmt>)"
+  irgend: "(irgend(ein[s]|eine[s]|einer|einem|einen)[ der]|(irgendwelche|einige|manche)[ (der|die|des)]|(irgendwelches|eine|eines|einer|einem|einen)[ (der|des)]|irgendwo[ (der|die|das)]|(einiges|manches|eins)[ (der|das|des)]|ein)"
```
andreasbrett: Since this PR doesn't include added tests for the added words, I have a hard time to understand why these are in there.

Is there a reason to keep `(irgendwelche|manche|einige) die` or v`(irgendwelche|manche|einige) des`? I can't find a test where this is used and can't think of a sentence where I would ask something like "irgendwelche die" or "irgendwelche des".

Do you have an example prompt for `einiges der` or `einiges das` or `eins des`?

hecktech27: since you mentioned added words: i made the deliberate decision to make the articles optional in the last part of the original sentence (that had a fixed `<artikel_bestimmt>` before) to be consistent with the first part(s) of the expansion rule where it is optional and since i can absolutely see usecases with `irgendwo/manche/...` i decided to go that way.
The cases you mentioned were all covered by the original expansion rule. I do absolutely agree with you - i couldn´t find any real usecases for those combinations as well but those actually were the cases where i thought "maybe i´m missing something" and wasn´t sure deleting those would be the right decision. but since I´m not the only one struggling i think we can go on deleting them 😉

hecktech27: i gave it another go and only kept combinations where i´m certain to find use cases. 
the new version reduces osc by ~2.6 million

PR #3386, opened 2025-08-20
[DE] reduce overall sentence count by avoiding <artikel_bestimmt> expansion rule

hecktech27:
For this PR i went through all 3 remaining files(besides common.yaml) containing the `<artikel_bestimmt>` expansion rule and exchanged the expansion rule with only the actually used variations.
OSC reduced by another ~2 million down to currently 39.7 million without loosing any tested sentences

### DIFF
--- a/sentences/de/climate_HassClimateSetTemperature.yaml
+++ b/sentences/de/climate_HassClimateSetTemperature.yaml
@@ -15,7 +15,7 @@ intents:
       - sentences:
           - "<heizung> <area_floor> auf <temperature>[ <setzen_end_of_sentence>]"
           - "<area> auf <temperature> <heat_cool>"
-          - "[<artikel_bestimmt> ]<floor> auf <temperature> <heat_cool>"
+          - "[das ]<floor> auf <temperature> <heat_cool>"
           - "<setzen> <area_floor> <heizung> auf <temperature>"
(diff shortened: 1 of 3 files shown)


sentences/de/climate_HassClimateSetTemperature.yaml
```diff
@@ -15,7 +15,7 @@ intents:
       - sentences:
           - "<heizung> <area_floor> auf <temperature>[ <setzen_end_of_sentence>]"
           - "<area> auf <temperature> <heat_cool>"
-          - "[<artikel_bestimmt> ]<floor> auf <temperature> <heat_cool>"
+          - "[(den|das) ]<floor> auf <temperature> <heat_cool>"
```
andreasbrett, line 18: This might become an issue at some point in time. I couldn't think of a floor that has a `die` article so I asked AI and indeed there are some not so common floor names that would require `die` but these are not used in typical smart homes (I guess... but you never know).

- Mansarde
- Galerie
- Kuppel

hecktech27: while you are absolutely right with your concerns regarding those options you already merged the fix for this potential issue. `<artikel_bestimmt>` is now an optional part of `<floor>` after @duhow´s latest changes ;)

PR #3398, opened 2025-08-22
[DE] split <alle> expansion rule to reduce overall sentence count

hecktech27:
This PR splits the `<alle>` expansion rule into
- `<alle>`
- `<alle_genitiv>`
- `<alle_genitiv_natural_subset>`

the last one contains only a small subset of the most commonly used genitive versions from the original <alle> rule.

all permutations of the 3 new expansion rules were covered by the original `<alle>` rule.

This drastically reduces overall sentence count by almost 30% **down to currently ~16 million.** :)


andreasbrett:
> youre right and i just tested... i would suggest to remove the natural subset(for now ;) ) since the current impact of this versus just using the genitiv rule is only -1000

Okay then let's get rid of that one.

PR #3419, opened 2025-08-27
chore: remove node_modules folder

duhow:
Some unwanted file got here, and as we don't use Javascript at the moment, I'm getting rid of it.


staticdev:
Thanks @duhow

PR #3426, opened 2025-08-29
[CA] Fix HassGetWeather spoken output with decimal and degrees

alexgraupera:
This PR fixes the handling of temperature units in the Catalan (ca) language responses for the **HassGetWeather** intent.

Previously, temperatures such as `18.8` were spoken incorrectly by the TTS engine as `"divuit d'agost"` (interpreted as a date) instead of `"divuit coma vuit"`.

The following improvements were made:

- Temperatures are now correctly formatted with a decimal comma `(18,8)` as used in Catalan.

- The temperature unit has been changed from `"ºC"` to a localized form: `"grau"` when the value is exactly 1, and `"graus"` otherwise.

- When the temperature is below zero, the suffix `"sota zero"` is appended, which is the natural way to express negative temperatures in Catalan.

This ensures more natural and accurate voice output in Catalan.

### DIFF
--- a/responses/ca/HassGetWeather.yaml
+++ b/responses/ca/HassGetWeather.yaml
@@ -21,4 +21,9 @@ responses:
           'windy': 'Vent',
           'windy-variant': 'Vent i núvols'
         } %}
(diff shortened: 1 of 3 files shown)


duhow (approved):
Moltes gràcies! 🙇🏻

PR #3430, opened 2025-09-01
[DE] HassLightSet: optimize brightness controls to reduce permutations

hecktech27:
For this PR I changed "auf" in brightness controls from optional to mandatory in sentences where it made no sense otherwise.
OSC - ~400.000 / ~2.5%
I had to remove one test-sentence that made no sense and was covered by the previous sentences.


andreasbrett (approved):
Makes sense. Thank you!

PR #3457, opened 2025-09-11
new skip words and others

timitoerner:


sentences/de/_common.yaml
```diff
@@ -696,3 +696,24 @@ skip_words:
...
+  - "vielleicht"
+  - "schnell"
+  - "gerne"
+  - "nun"
+  - "also"
+  - "na"
+  - "kannst du"
+  - "mach mal"
+  - "bitte schön"
+  - "danke schön"
```
felixschndr, line 709: I would add `dankeschön` and `bitteschön` here as well

PR #3460, opened 2025-09-13
[GA] Add HassBroadcast

AlanGreene:
Similar to the issues raised in https://github.com/OHF-Voice/intents/pull/2882 these sentences aren't as smooth / natural as I would like.

Ideally a sentence such as "Fógair go bhfuil an dinnéar réidh" would result in a broadcast `message` of "Tá an dinnéar réidh".

Since this isn't currently feasible, I've gone with the same approach agreed in #2882, which is essentially `<broadcast>: {message}`.


GGamerzz:
Sounds good, will be interesting for sure

PR #3480, opened 2025-09-18
[FR] add media player relative volume setting

FraBoCH:
Adding French sentences for HassSetVolumeRelative intent


jlpouffier (changes requested):
Hello !

Thx a lot for your PR!

We have common expansion rules for verbs `monte` and `baisse` [here](https://github.com/OHF-Voice/intents/blob/21b795018283fe4f9544e895c94a8b58a955f87a/sentences/fr/_common.yaml#L515)

Do you mind using them so that the sentences are coherent with other sentences using these verbs?

You can take inspiration from the [Absolute Volume setting](https://github.com/OHF-Voice/intents/blob/61838f179a902fb66808d1b854770956d275f0a4/sentences/fr/media_player_HassSetVolume.yaml#L8) in case you want to know how to use expansion rules!

FraBoCH:
Thanks for your review @jlpouffier! I reused the existing expansion as requested

synesthesiam:
Thanks for this, @FraBoCH, and for addressing the review feedback! Unfortunately this has been superseded — French `HassSetVolumeRelative` was added to `main` in the meantime via #3833, which covers the same relative-volume use cases (plus floor targets and additional phrasings) using the `<augmente>`/`<diminue>`/`<monte>`/`<baisse>` expansion rules. As a result this PR now conflicts on every file. Closing as superseded — sorry for the duplicated effort, and thank you for the contribution!

PR #3486, opened 2025-09-23
ci: Add intentfest validate with errors for changed files

staticdev:
Summary:
- Adds a new argument `--error-on-warn-pr` to `script/intentfest/validate.py` that converts warns into errors for newly changed files with validation errors.
- Adds a separate job on the ci workflow to do complete validation with warnings if no language has been changed, or with errors for files in language changed (PR only).
- Removes the `script/intentfest/validate.py` to remove redundant execution.

This PR was tested with an extra commit on top [here](https://github.com/staticdev/intents/pull/1) to force it to fail with the correct errors.

### DIFF
--- a/script/intentfest/validate.py
+++ b/script/intentfest/validate.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import argparse
 import itertools
 import json
+import re
 from collections import Counter, defaultdict
 from collections.abc import Callable, Collection
 from datetime import datetime
@@ -532,11 +533,6 @@ def get_arguments() -> argparse.Namespace:
         help="The language(s) to validate. Comma-separated for multiple.",
     )
     parser.add_argument(
-        "--error-on-warn-pr",
-        action="store_true",
-        help="Treat warnings in PR-changed files as errors.",
-    )
-    parser.add_argument(
         "--changed-files-json",
         type=str,
         help="JSON array of changed file paths (used by CI to pass changed files)",
@@ -688,12 +684,7 @@ def run() -> int:
         if not warnings[language]:
             warnings.pop(language)
 
-    if args.error_on_warn_pr and warnings:
-        # Require changed files to be provided when running in PR error-on-warn mode
-        if not args.changed_files_json:
-            print("error: --error-on-warn-pr requires --changed-files-json in CI")
-            return 2
-
+    if args.changed_files_json and warnings:
         try:
             changed_files = json.loads(args.changed_files_json)
         except Exception as err:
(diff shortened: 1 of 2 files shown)


staticdev:
@synesthesiam could you please check this one when you have time?

script/lint
```diff
@@ -27,7 +27,5 @@ flake8 "${python_files[@]}"
 pylint "${python_files[@]}"
 mypy "${python_files[@]}"
 
-python3 -m script.intentfest validate
```
balloob, line 30: This means that it now won't run locally anymore?

balloob (changes requested):
1 comment

script/lint
staticdev: Fixed @balloob, you are right.

script/intentfest/validate.py
```diff
@@ -667,10 +678,43 @@ def run() -> int:
...
+        except Exception as err:
+            print(f"Failed to parse changed files JSON: {err}")
+            return 2
+
+        # Check if any warning is for a changed file
+        warn_files = set()
+        for language, language_warnings in warnings.items():
+            for warning in language_warnings:
+                # Try to extract file path from warning string
+                import re
```
balloob: don't do inline import.

script/intentfest/validate.py
```diff
@@ -667,10 +678,43 @@ def run() -> int:
         # Remove language if no errors
         if not errors[language]:
             errors.pop(language)
-
         if not warnings[language]:
             warnings.pop(language)
 
+    if args.error_on_warn_pr and warnings:
```
balloob: why even have an extra flag, why don't we just treat the existense of a changed_file_json as the reason to do this.

script/intentfest/validate.py
```diff
@@ -527,6 +528,16 @@ def get_arguments() -> argparse.Namespace:
     parser.add_argument(
         "--language", type=str, choices=LANGUAGES, help="The language to validate."
     )
+    parser.add_argument(
+        "--error-on-warn-pr",
+        action="store_true",
+        help="Treat warnings in PR-changed files as errors.",
+    )
+    parser.add_argument(
+        "--changed-files-json",
```
balloob, line 536: I don't see this option being used to limit what files we're linting .

balloob:
Please don't force push so we can see what has changed.

script/intentfest/validate.py
staticdev: Fixed.

script/intentfest/validate.py
staticdev: You are right, I did a cleanup and removed this argument, it is redundant with `changed_files_json`.

script/intentfest/validate.py
staticdev: No need, you are 100% correct. Removed.

balloob (approved):

PR #3489, opened 2025-09-25
[DE] Add "Überall" as a command for light on/off

hecktech27:
This PR adds support for commands like 
- Schalte überall das Licht aus
- überall Licht an

I went with a separate expansion rule to make sure there is no unexpected behavior in HassLightSet and to keep this PR small. i will go through the other sentences containing `<alle_lichter>` and will move them to `<alle_lichter_ueberall>` where it makes sense step by step.


hecktech27:
@andreasbrett 
https://github.com/OHF-Voice/intents/pull/3423#issuecomment-3233665651

andreasbrett (approved):

hecktech27:
i´ve just removed the duplicates - don´t know if you saw it before approving

andreasbrett:
> i´ve just removed the duplicates - don´t know if you saw it before approving

I saw it shortly after hitting approve but even better that way. Auto-merge will take care of the rest. Thank you and if everything works out this should land in next week's HA release (2025.10).

PR #3492, opened 2025-09-29
[DE] new Timer sentences

tobiloeb:
Added new sentences for timer creation in german. So its possible to say "Timer for my Pizza" and "Timer for the pizza" .


hecktech27:
> Looks good but could you please add "mein" so it also allows "Stelle 5 Minuten timer für mein Essen/Auto/..."?

and maybe "unser / unsere / unseren" as well

tobiloeb:
Hi @hecktech27 @andreasbrett,

yes of course, thats a good point. Would it make sense to create a new common expantion rule for:

possessivpronom_mein: (mein|meine|meines|meiner|meinem|meinen)
possessivpronom_unser: (unser|unsere|unseres|unserem|unseren)

So it can be used in different sentences as well? Maybe this change is to big for this small PR? I tried to keep this PR simple, so it can be merged easely. What do you think?

Regards
Tobias

andreasbrett:
Very thoughtful, thank you! It's fine to have it in this PR.

PR #3499, opened 2025-09-30
[DE] restructure expansion rules in common.yaml

hecktech27:
This PR adds no features but reorders the expansion rules in the commons file.
I´ve added some comments for the respective groups of rules as well.
I did not change any expansion rules - just rearranged them by context.


tobiloeb:
I like that

andreasbrett (approved):
Indeed, this is quite helpful. Thank you!

PR #3504, opened 2025-10-01
[DE] add "stell" and "erstell" words for startTimer

tobiloeb:
Today I tried to say "Stell einen Timer für 12 Minuten" and HA answered very strange. 

With this PR you can say:

Stell einen Timer für 10 Minuten für die Nudeln
Erstell einen Timer für 5 Minuten

and so on.

The word "erstell[e]" isn't used elsewhere currently so I decided to add a new expantion_rule like it's done for "stell[e]" for it.


andreasbrett (approved):
Good one. Thank you for your contributions lately!

PR #3505, opened 2025-10-02
[NL] Add media_plater mute/unmute

TheFes:
Adds Dutch sentences to mute and unmute media players

### DIFF
--- a/sentences/nl/media_player_HassMediaPlayerMute.yaml
+++ b/sentences/nl/media_player_HassMediaPlayerMute.yaml
@@ -1,0 +1,25 @@
+language: nl
+intents:
+  HassMediaPlayerMute:
+    data:
+      - sentences:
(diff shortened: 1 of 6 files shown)


sentences/nl/media_player_HassMediaPlayerMute.yaml
```diff
@@ -0,0 +1,25 @@
+language: nl
+intents:
+  HassMediaPlayerMute:
+    data:
+      - sentences:
+          - "(demp[en]|mute)"
+          - "zet het geluid uit"
```
balloob, line 7: Tricky one. Because why not stop the media player?

TheFes: I've asked around a bit, and everyone expected that `zet het geluid uit` would mute the device, not turn it off.

PR #3506, opened 2025-10-02
[DE] add skip word "noch"

tobiloeb:
This PR adds the word "noch" to the list of skipwords. So a lot of sentences can be spoken more naturally. 
For example:

Stelle noch einen Timer für die Nulden
Welche Fenster sind noch auf?
Welches Licht ist noch an?
Füge noch Eier zur Einkaufsliste hinzu
....
and so on.

The only sentences that currently uses the word "noch" is the TimerStatus. The word could be removed from this sentences as well, but for my feeling its should stay there (the sentences are more common with the word noch).

Can be discussed, wether it sould be removed from the timerStatus sentences or stay there as "[optional]" part of the sentences. :)


tobiloeb:
@andreasbrett I decided to remove the word "noch" from TimerStatus sentences but add a comment. This will reduces the count of sentences of almost 1000.

sentences/de/homeassistant_HassTimerStatus.yaml
```diff
@@ -3,22 +3,23 @@ intents:
   HassTimerStatus:
     data:
       - sentences:
+          # The word "noch" is a skip_word and removed from sentences here to reduce the sentence count
```
mib1185:
```suggestion
```
I don't think we need this comment, as we've nowhere an similar comment that we refrained a skip work.

hecktech27:
Just a heads up: #3588 
i defined "wie lange noch" as a custom sentence for me yesterday and came across the linked issue. 
in my case i simply went with "wie lange" as a sentence definition instead, but i can imagine "noch" being a skip word to become a problem in the future.
Since the timerStatus sentences are pretty light sentencecount-wise(currently ~3500) it might make sense to implement `[noch]` in those sentences instead. what do you think?
it might be worth to think about adding things like
- "kannst du noch".....das licht im x ausschalten
- "bitte mach noch"......die musik im x aus
- "starte noch"....einen 5 minuten timer
- "Stelle noch"...einen 5 minuten timer

since those would only be skipped in a more defined context.

PR #3519, opened 2025-10-06
[MN] Update climate_HassClimateGetTemperature.yaml

chuk-a:


synesthesiam:
Thanks @chuk-a! Closing this one as it can't pass on its own: it adds a test (`"Одоо температур хэд байна?"` → `21 °C`) to `tests/mn/climate_HassClimateGetTemperature.yaml`, but `sentences/mn/climate_HassClimateGetTemperature.yaml` currently has `sentences: []` — so there's no sentence for the test to match, and validation fails with "not all sentences have tests". Also, the asserted response `21 °C` doesn't match the current mn response template (`"{{ state.state }} хэм"`). To make this work, the actual Mongolian sentences need to be added to the sentences file first, with the test response matching the rendered `HassClimateGetTemperature` response. Happy to review a combined PR that does both!

PR #3535, opened 2025-10-22
[DE] Add trigger sentences for HassGetCurrentDate and HassGetCurrentTime

felixschndr:
Adds simple trigger sentences as I (and my SO and probably others) don't like to say `Whats the current date` but only ask for the `date`


andreasbrett (commented):
I am hesitant with "Uhr" as it's just one syllable for the whole "sentence" and also basically just one vowel sound. Could be error prone for STT.

andreasbrett (approved):
Looks good to me. Thank you for your contribution!

PR #3538, opened 2025-10-24
[DE] LightTurnOn: avoid "Licht an" not being recognized due to STT failing

hecktech27:
this is more of a starting point for discussion than a "I´m sure about that"-PR.
I´ve had STT failing to recognize "Licht an" for me (and my family) in quite a lot of cases where it would always result in "Lichtern".
This happens in quiet environments when you talk a little bit faster than normal.
Since it´s a command that i think is quite important i would propose to add this edge case even though it is a little weird... 
Maybe even add a comment behind it for future reference how/why this made it´s way into the repo...
I would like to hear your thoughts on that :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-hassil==3.2.0
+hassil==3.2.1
 PyYAML==6.0.3
 voluptuous==0.15.2
 regex==2025.9.18
--- a/sentences/de/media_player_HassSetVolumeRelative.yaml
+++ b/sentences/de/media_player_HassSetVolumeRelative.yaml
@@ -4,12 +4,13 @@ intents:
     data:
       # Current area - down
       - sentences:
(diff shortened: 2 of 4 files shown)


hecktech27:
i really like that approach since it´s easy to revert in the future when it´s no longer needed!
while it is possible to fix those "dirty words" in idividual sentences/automations - i think we should implement fixes for known common misunderstandings. the need for custom modifications just to make assist understand you regularly with common sentences does not support overall acceptance of the system in my opinion ;)

hecktech27:
i implemented fixes for stt as expansion rules in the commons file similar to the french approach. i added comments for better reference as well.
in my opinion it would make sense to stick to a naming pattern for those expansion rules that directly hints yout to this being "a special" expansion rule.
i further think that it would make sense to add a comment for each expansion rule that explains which STT-error is covered by that rule. 
what do you think about that?

andreasbrett:
I like it!

hecktech27:
I´m using Homeassistant Cloud - if i had to guess probably the default for many/most users(but i might be wrong on that)

PR #3542, opened 2025-10-26
[DE] VolumeSetRelative: fix translation error

hecktech27:
i found two occurances of "percent" in the german sentences... replaced those by "Prozent" and adjusted tests accordingly.

### DIFF
--- a/sentences/de/homeassistant_HassGetCurrentDate.yaml
+++ b/sentences/de/homeassistant_HassGetCurrentDate.yaml
@@ -3,6 +3,7 @@ intents:
   HassGetCurrentDate:
     data:
       - sentences:
+          - "datum"
           - "<welche> (tag|datum) (ist|haben wir)[ heute]"
           - "was (ist|haben wir) [heute ]für ein[en] (Tag|Datum)"
           - "(nenn[e]|gib|gebe|sag[e]|verrat[e]) mir das([ heutige] Datum|[ aktuelle] Datum[ von heute])"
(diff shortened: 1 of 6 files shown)


tests/de/media_player_HassSetVolumeRelative.yaml
```diff
@@ -281,7 +281,7 @@ tests:
     response: "Lautstärke festgelegt"
   # Area by name - step_down
   - sentences:
-      - "wohnzimmer um 20% leiser"
```
mib1185, line 284: this test sentence should still work, why is it removed (_also the other ones_)?

hecktech27: My Intention was Not really to remove it but to include `Prozent` instead of `%` in one of the sentences in each Block of Test sentences. I think there are enough sentences covering `%` 
But we can add that sentence with % again of you prefer that

mib1185: yeah, removing existing and correct test sentences is always a step back in test coverage and one might miss some regressions in the future (_"you never know"_). So adding more test sentences is always great and highly appreciated 👍 but removing existing ones should be avoided :)

hecktech27: I will add it again this evening

hecktech27: @mib1185 done ;)

PR #3554, opened 2025-11-01
[SR] Update _common.yaml

PengBG:
speaks as he, strange when female voice is selected, proposal is neutral
Rest is typo fixes and adding more choices

### DIFF
--- a/sentences/sr/_common.yaml
+++ b/sentences/sr/_common.yaml
@@ -1,9 +1,7 @@
 language: sr
-language: sr
 responses:
   errors:
     # General errors
-    no_intent: "Нажалост, не разумем"
     no_intent: "Нажалост, не разумем"
     handle_error: "Дошло је до неочекиване грешке"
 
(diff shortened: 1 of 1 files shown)


sentences/sr/_common.yaml
```diff
@@ -75,59 +75,59 @@ lists:
...
+      - in: "жалузин(а|е|у)"
         out: "shade"
       - in: "ролетн[а|е|и|у]"
         out: "shutter"
       - in: "прозор[а|и|е]"
         out: "window"
   lock_states:
     values:
-      - in: "закључано"
+      - in: "закључан[а|и|о|их]|закључај"
```
ddxic: "lock" would be an action, but here we need states. I would suggest to remove the last one.

sentences/sr/_common.yaml
```diff
@@ -75,59 +75,59 @@ lists:
...
         out: "locked"
-      - in: "откључано"
+      - in: "откључан[а|и|о|их]|откључај"
```
ddxic: "unlock" would be an action, but here we need states. I would suggest to remove the last one.

PengBG: You are right, last one is action, but since I am new to intents I looked at sr-Latn and didn't give it much thought. Will remove action

PengBG:
I made some wrong click but in  [f5bb315](https://github.com/OHF-Voice/intents/pull/3554/commits/f5bb31554a34fd6fc7832537968d5255aa36e58a) only change is that I removed actions in line 128 and 130 "закључај" and "откључај"

PR #3563, opened 2025-11-04
[DE] HassStartTimer add 'nach' option alongside existing 'in' keyword

hecktech27:
this pr is heavily inspired by #3555 
it adds support for sentences like
> nach 5 Minuten die Fenster schließen

### DIFF
--- a/.github/workflows/backfill_labels.yaml
+++ b/.github/workflows/backfill_labels.yaml
@@ -20,7 +20,7 @@ jobs:
             const owner = context.repo.owner;
             const repo = context.repo.repo;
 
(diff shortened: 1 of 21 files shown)


andreasbrett:
I am hesitant with this since `after` is only used in conjunction with a reference point of time. This can be an explicit time or indirectly referenced from events that were described earlier (`I went for a walk. After 5 minutes I felt tired.`). For a voice assistant prompt there is no context and so it would be an indirect `5 minutes after (now)`. In that case though the `in 5 minutes` form is typically used. If someone said "schalte das licht ein nach 5 minuten" I would have a strong urge to ask "5 minuten nach WAS?".

hecktech27:
i get your point - i´m in the "in 5 minutes" department as well... but i can also see other people using "after" instead of "now". and it´s only +500 sentences ;) but thats up to you

PR #3564, opened 2025-11-04
[DE] add "Restzeit" and "von dem" to timer status requests

hecktech27:
this PR adds "Restzeit" in addition to "Status" for timer status requests.
i introduced 
> ...von dem...
> ...von meinem...
> ...vom...

for both "Restzeit" and "status" in this PR as well


andreasbrett (approved):
LGTM

PR #3566, opened 2025-11-04
[DE] add option to ask for media player's volume level

hecktech27:
This PR adds a response for media_player volume level requests.

### DIFF
--- a/sentences/de/homeassistant_HassDecreaseTimer.yaml
+++ b/sentences/de/homeassistant_HassDecreaseTimer.yaml
@@ -17,3 +17,4 @@ intents:
           - "[(den|meinen) ]Timer <area> um <timer_duration> <timer_decrease_end_of_sentence>"
           - "[(den|meinen) ]{timer_name:name} Timer um <timer_duration> <timer_decrease_end_of_sentence>"
           - "[(den|meinen) ]Timer (namens|für) {timer_name:name} um <timer_duration> <timer_decrease_end_of_sentence>"
+          - "[(<timer_start>|{area}|{timer_name:name}) Timer ]<timer_duration> (weniger|kürzer)"
--- a/sentences/de/homeassistant_HassIncreaseTimer.yaml
+++ b/sentences/de/homeassistant_HassIncreaseTimer.yaml
@@ -17,3 +17,4 @@ intents:
           - "[(den|meinen) ]Timer <area> um <timer_duration> verlängern"
           - "[(den|meinen) ]{timer_name:name} Timer um <timer_duration> verlängern"
           - "[(den|meinen) ]Timer (namens|für) {timer_name:name} um <timer_duration> verlängern"
+          - "[(<timer_start>|{area}|{timer_name:name}) Timer ]<timer_duration> (mehr|länger)"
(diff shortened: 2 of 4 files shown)


responses/de/HassGetState.yaml
```diff
@@ -85,3 +85,11 @@ responses:
             {%- endfor -%}
           {% endif %}
         {% endif %}
+
+      get_volume: |
+        {% if 'volume_level' in state.attributes %}
+          {% set volume = state.attributes.volume_level | int %}
```
andreasbrett: Have you tested this on a real HA instance? I don't know why the fixtures contain a `volume_level` attribute of 50. Typically the `volume_level` attribute of a `media_player` is in `float[0..1]` not `int[0..100]`.

tests/de/homeassistant_HassGetState.yaml
```diff
@@ -8,6 +8,15 @@ tests:
...
 
+  - sentences:
+      - "wie laut ist der tv"
+      - "Lautstärke TV"
+    intent:
+      name: HassGetState
+      slots:
+        name: "TV"
+    response: "Tv ist auf 50 % eingestellt"
+
```
andreasbrett, line 19: Could you please add a fixture and test for a media player entity without volume attribute?

responses/de/HassGetState.yaml
hecktech27: to be fair i sticked to the tests/provided fixtures. i just checked with my media players and you´re right - they are all float 0..1. should we adjust the fixture in the tests? i just searched the `/tests` directory for `volume_level: "50"` and there are 31 results... so more of a global problem here...

hecktech27: i went with adjusting the test fixture to the reality...

tests/de/homeassistant_HassGetState.yaml
hecktech27: done

responses/de/HassGetState.yaml
andreasbrett: Thank you. I tested it locally in `custom_sentences` to ensure `volume_level` is not converted to percent before being passed to Assist. It is not converted, so all the `volume_level` fixtures should be `float[0..1]`.

responses/de/HassGetState.yaml
```diff
@@ -85,3 +85,11 @@ responses:
             {%- endfor -%}
           {% endif %}
         {% endif %}
+
+      get_volume: |
+        {% if 'volume_level' in state.attributes %}
+          {% set volume = (state.attributes.volume_level | float(0)) * 100 %}
+          {{ slots.name | capitalize }} ist auf {{ volume | round(0) | int }} % eingestellt
+        {% else %}
+          {{ slots.name | capitalize }} hat keinen Lautstärke-Wert
```
andreasbrett, line 96:
```suggestion
        {% elif state.state == "off" %}
          {{ slots.name | capitalize }} ist im Moment ausgeschaltet
        {% else %}
          {{ slots.name | capitalize }} hat keinen Lautstärke-Wert
```

If the media player is off, there is no `volume_level` attribute, so I think we should let the user know the difference.

responses/de/HassGetState.yaml
hecktech27: can/should we/i adjust these fixtures for other languages as well in one PR or is this something that needs 31 single PRs?

andreasbrett (approved):

PR #3571, opened 2025-11-07
[DE] add option to ask for brightness levels and color temperature

hecktech27:
this pr adds a response for light-related requests for brightness and color temperature


hecktech27:
actually i think you need to have some option to ask for the current brightness in order to use set-brightness-commands in any useful way. otherwise you always need to guess the current brightness and then ask assist to set the light to a certain percentage(if you want to use percentage to control the lights)

PR #3582, opened 2025-11-10
[DE] fix/add get target temperature for climate

tobiloeb:
Today I ask the voice assist: "whats the current target temperature for my office" (Auf wieviel Grad ist die Temperatur im Büro eingestellt?). I asked this, because the room wasn't getting hot and I wanted to know which temp is set for the climate.

The response was the actual (current) temperature of the room and not whats set as target temperature. I would say, that this is a "bug". If you ask in German what is configured (was ist eingestellt), you would expect the target temperture which is configured for this room and not the actual messaured temp.

So this PR will "fix" this. 😃 

To make the different responses more readable, I changed the word "default response" to "current_temperature" and the new one as "target_temperature".

### DIFF
--- a/responses/de/HassClimateGetTemperature.yaml
+++ b/responses/de/HassClimateGetTemperature.yaml
@@ -2,7 +2,12 @@ language: de
 responses:
   intents:
     HassClimateGetTemperature:
-      default: >
+      current_temperature: >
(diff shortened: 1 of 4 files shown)


tobiloeb:
I have to thank you @andreasbrett for this ultra fast review at this time of day!!

Haha yes I took a lock at the EN as well 😄

PR #3591, opened 2025-11-20
[DE] add "wann endet mein Timer" sentence structure

hecktech27:
as described in the title this PR adds support for sentences like
- wann ist mein timer fertig
- wann endet mein timer
- wann ist mein Pizza Timer fertig
- wann läuft mein Wohnzimmer Timer ab
- ...

i do remember that at some point in the past we discussed not skipping the "timer" command für starting timers.
what do you think about the following(Not yet included) sentence:
`wann (ist|sind) [(der|die) ]{timer_name:name} fertig`
in my opinion this is specific enough in the "homeassistant-context" to be a recognized as a timer command even without the "timer" specifically mentioned. 
it would allow for very naturally asking for "wann ist die Pizza fertig?". and since it does not start any action it might be worth including that one.

### DIFF
--- a/sentences/de/homeassistant_HassTimerStatus.yaml
+++ b/sentences/de/homeassistant_HassTimerStatus.yaml
@@ -37,3 +37,4 @@ intents:
           - "wann läuft (der|mein) (<timer_start>|{area}) timer ab"
           - "wann läuft (der|mein) Timer <area> ab"
           - "wann ist [(der|mein) ]{timer_name:name} Timer (zu ende|fertig)"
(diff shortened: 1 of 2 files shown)


andreasbrett:
> what do you think about the following(Not yet included) sentence:
> `wann (ist|sind) [(der|die) ]{timer_name:name} fertig`

With the added "wann" and "fertig" pre-/suffix it works out for me. Very clear that this is a timer question. Could you please add that here as well?

hecktech27:
I already have an updated Version that covers der die das mein meine - will upload this later since i currently only have my phone with me. Just have to add the tests

andreasbrett (approved):

PR #3604, opened 2025-11-28
[DE] HassGetState: fix usage of binary_sensor-state and lock-state lists

hecktech27:
For this PR I need to explain a little more:
the current main branch does not offer the ability to ask for the state of binary sensors with "offen" or "geschlossen" - so when you have door sensors that offer a binary sensor and you name it e.g. "Küchentür" you were not able to ask "ist die Küchentür offen/geschlossen".
i tried to simply add offen/geschlossen to the `on_off_states` list and was given an error stating that this would expect a lock entity.
so the current state is: 
`ist die Küchentür geschlossen` expects a lock entity which i think is wrong. 
i changed the lists for lock states accordingly to only match sentences like
`ist die Küchentür abgeschlossen`.
while it does look like i deleted test sentences - i did not. they were moved and i replaced the lock fixture `Haustür` with the binary sensor fixture `Küchentür`.
i made sure that every sentence working previously is still working, but now responding with what was asked for - the open/closed state of the door. for this i had to add some additional sentences for binary sensors.
you can still request the lock state of a door with "ist die Küchentür abgeschlossen" but no longer using "geschlossen".

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ flake8==7.3.0
 isort==7.0.0
 mypy==1.18.2
 pylint==4.0.3
-pytest==9.0.0
+pytest==9.0.1
 pytest-xdist[psutil]==3.8.0
 types-PyYAML==6.0.12.20250915
 types-regex==2025.11.3.20251106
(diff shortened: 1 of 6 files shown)


andreasbrett (approved):
Awesome. I remember at some point in the very past I tried to ask for open doors and was confused it did not work. Thanks for providing that fix.

PR #3632, opened 2025-12-14
[DE] HassStartTimer: add support for trailing timer-set command

hecktech27:
This PR adds support for sentence structures like

- einen 5 Minuten Timer starten
- einen 5 Minuten Timer stellen
- einen Timer auf 5 Minuten Stellen
- einen Timer für die Pizza für 5 Minuten starten

this is especially useful when combined with skip words like "kannst du..." and similar

i had to split up some existing sentences to avoid unnecessary permutations


andreasbrett (approved):
I like especially when used with the skip words. Quite natural sentences ❤️

PR #3634, opened 2025-12-15
Add "depois de" to HassStartTimer

mhalano:
Add "depois de" to HassStartTimer in Portuguese.

The equivalent in English would be "after".


mhalano:
@staticdev Tests added. Please let me know if you need anything else.

mhalano:
@staticdev How is the cadence between intents releases and Home Assistant releases? I saw yesterday the version 2026.1.1 for intents was released, so I'm wondering when we will have a version of HA available that includes this intents version.

PR #3635, opened 2025-12-15
[DE] HassNevermind: add more cancel-commands

hecktech27:
this PR adds some more commands to cancel false positives.


sentences/de/homeassistant_HassNevermind.yaml
```diff
@@ -4,6 +4,9 @@ intents:
     data:
       - sentences:
           - "vergiss es"
+          - "alles (gut|ok)"
+          - "das hast du falsch verstanden"
+          - "nicht du"
```
mib1185, line 9: I like that :laughing:

hecktech27: Thats my personal Go-To 😇

PR #3655, opened 2025-12-30
[DE] MediaPlayer: "weiter" triggers MediaNext instead of MediaUnpause

hecktech27:
This PR moves `weiter` as a command from HassMediaUnpause to HassMediaNext.
In my opinion sentences like
- Weiter
- TV weiter
- Wohnzimmer weiter

should trigger MediaNext(changed in this PR) while sentences like `Musik weiter laufen lassen` should trigger MediaUnpause(as they already do)

### DIFF
--- a/sentences/de/media_player_HassMediaNext.yaml
+++ b/sentences/de/media_player_HassMediaNext.yaml
@@ -13,6 +13,7 @@ intents:
           - "[((an|auf)[ dem]|am) ]<name> <zu_dem> <naechster> <song>[ (vor[wärts]|weiter)][ ](springen|spulen)"
           - "überspring[e] ((das|dieses)[ eine]|ein) <song> [((an|auf)[ dem]|am) ]<name>"
           - "[((an|auf)[ dem]|am) ]<name> ((das|dieses)[ eine]|ein) <song> überspringen"
+          - "<name> weiter"
         requires_context:
           domain: media_player
       - sentences:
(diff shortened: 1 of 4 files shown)


andreasbrett (commented):
I disagree, for me "TV weiter" is conditional and should only play the next media item when the player is currently playing. When it is currently paused, "TV weiter" should unpause the player. At least that's my intent or expectation when saying "TV weiter". 

That logic would need to be implemented upstream by @synesthesiam.

mib1185:
I agree with @andreasbrett or at least would like to keep it as is for now, until we have the possibility for conditional intents.
I use it often in the flow of "ok nabu, pause" ... "ok nabu, weiter"

mib1185:
"we" maybe could file a feature request for the conditional intents or place it on the discord channel 🤔 😬 

About moving "weiter" from HassMediaUnpause to HassMediaNext, I still think we shouldn't do this, as it changes already existing behaviour and users might already be common with this (_at least there are no feedbacks, which would complain about it_) so this properly would need to be classified as a _breaking change_ or better should be discussed with the community what they think about it.

balloob:

German has just been migrated to a brand new sentence format in #3887. The new slot-combination format makes it much easier to **write, validate, and maintain** sentences going forward. Unfortunately, this PR is written against the old format, so it can no longer be merged as-is.

You can follow the progress and coverage for German here: https://ohf-voice.github.io/intents/language_de.html

---

PR #3665, opened 2026-01-03
[DE] ClimateGetTemperature: allow shorthand "Temperatur"

hecktech27:
This PR adds shorter temperature requests to climate entities


andreasbrett (approved):
I like it

PR #3666, opened 2026-01-03
[DE] GetWeather: add outside temperature requests

hecktech27:
This PR adds requests for only the outside temperature to HassGetWeather.

I want to take the opportunity to make everybody reading this aware of the following feature request:
https://github.com/orgs/home-assistant/discussions/495

Currently there is no way to set a "default weather entity" assist uses when answering weather requests so you might not always get the desired output when you have more than one weather entity exposed to assist.
(this applies to all weather-related requests without naming a specific entity - not only this PR)


hecktech27:
i hope so too. i think there might be more people out there who actually have e.g. their families locations as weather entities set up(and then - why not expose it to assist... IF it works correct)

PR #3668, opened 2026-01-04
[DE] HassGetState: query update entities

hecktech27:
This PR adds the option to ask for availlable updates.

When i started implementing this i only targeted `update.*` entities since(as far as i can see) most devices integrated into homeassistant that support OTA-Updates provide such an update entity.
only after i finished i realized that e.g. the EN-Repo only targets binary sensors with device_class `update`.
in my homeassistant instance i could only find 3 binary sensor update entities while i could find almost 80 update entities.

since this is a different approach than some/all other languages use i´d definitely like your opinion on this @mib1185 @andreasbrett 

all 3 binary sensor update entities were physical devices - so to add a possibility to ask for those as well i went with `Geräte Updates` for device_class update and `[(Firmware|Software) ]Updates` for the more frequently used device_class firmware.
This more or less leads to more natural requests like "Sind Updates verfügbar" defaulting to the update domain.


andreasbrett:
The update entities were introduced in 2022.4 = almost 4 years ago. I don't think we should invest time to support their predecessors. I'd rather bring up to discussion if we can globally have Assist use the "modern" entities (maybe alongside the to-be-deprecated binary sensors). I couldn't find deprecation plans for binary sensors of device class update so far - would be interesting to see what the depreciation plans are to weigh in how much efforts should be invested in that old tech.

https://developers.home-assistant.io/blog/2022/03/20/update-entity

balloob:

German has just been migrated to a brand new sentence format in #3887. The new slot-combination format makes it much easier to **write, validate, and maintain** sentences going forward. Unfortunately, this PR is written against the old format, so it can no longer be merged as-is.

You can follow the progress and coverage for German here: https://ohf-voice.github.io/intents/language_de.html

---

PR #3671, opened 2026-01-06
[DE] StartTimer: add more natural sentence variations

hecktech27:
This PR adds sentences like
- Einen Timer auf 5 Minuten stellen und ihn Pizza nennen
- stelle einen Timer auf 5 Minuten und nenn ihn Pizza

the first version especially works well when used in conjunction with skip words like `könntest|würdest du`


andreasbrett (approved):
I like it. Quite long sentence but I'm a huge proponent of having natural prompts even if I myself only use the shortest possible combinations (and I'm the only one in my household to do so).

PR #3846, opened 2026-05-04
[DE] add support for color temperatures

hecktech27:
for reference: #3743
DE-implementation of color temperature support for HassLightSet
i went with the existing de color sentences and adjusted them to support "Farbtemperatur" since both are semantically very similar.
added some tests at the end of the tests file to cover all supported color temperature names.


andreasbrett (approved):
Lgtm!

PR #3870, opened 2026-06-10
[DE] HassGetTemperature: handle unavailable and unknown states

hecktech27:
this PR adjusts response templates to not respond with a technical error message if an assigned room temperature sensor is unavailable or unknown.
this avoids `ValueError: Template error: float got invalid input.........` (e.g. if an assigned battery powered sensor goes offline) and instead responds with 'Der Temperatursensor ist nicht verfügbar.'


andreasbrett (approved):
Very nice, I like it!

PR #3914, opened 2026-06-24
Fix script/test so --language works (loads tests/conftest.py)

balloob:
## Problem

`script/test --language <lang>` fails with `pytest: error: unrecognized arguments: --language` for any PR whose changes are confined to `sentences/`, `responses/`, or `tests/`. CI's `extract-languages` step only emits a non-empty `--language` in exactly that case (a change outside those dirs makes it run full tests instead), so the bug is normally masked — most language PRs also touch `rules/`/`lists/`, taking the full-test path. A migration that creates no `rules/`/`lists/` (e.g. mn, PR #3912) hits the `--language` path and fails CI.

## Cause

`script/test` ran `pytest -vv -n auto "$@"` with **no test path**. Without a path, pytest uses `testpaths` for collection but does **not** load `tests/conftest.py` as an *initial* conftest, so `pytest_addoption("--language")` (defined there) is never registered → the argument is rejected before collection.

## Fix

Pass the `tests` path explicitly: `pytest -vv -n auto tests "$@"`. This makes pytest load `tests/conftest.py` as an initial conftest, registering `--language`. Verified locally: `pytest tests --language en` collects 357 tests, whereas `pytest --language en` errors.

No behavior change for the no-arg `script/test` case (`tests` is already the sole `testpaths` entry).

https://claude.ai/code/session_016BW5FhdPgJr9rzMhmXtpBS

---

### DIFF
--- a/script/test
+++ b/script/test
@@ -16,4 +16,8 @@ if [ -d "${venv}" ]; then
 fi
 
 python3 -m script.intentfest validate "$@"
-pytest -vv -n auto "$@"
+# Pass the "tests" path explicitly so pytest loads tests/conftest.py as an
+# initial conftest. Without a path, pytest relies on `testpaths`, which does not
+# register conftest options early enough, so `--language` (defined in
+# tests/conftest.py via pytest_addoption) is rejected as an unknown argument.
+pytest -vv -n auto tests "$@"


script/test
```diff
@@ -16,4 +16,8 @@ if [ -d "${venv}" ]; then
 fi
 
 python3 -m script.intentfest validate "$@"
-pytest -vv -n auto "$@"
+# Pass the "tests" path explicitly so pytest loads tests/conftest.py as an
+# initial conftest. Without a path, pytest relies on `testpaths`, which does not
+# register conftest options early enough, so `--language` (defined in
+# tests/conftest.py via pytest_addoption) is rejected as an unknown argument.
+pytest -vv -n auto tests "$@"
```
balloob, line 23: Wouldn't this then run the full test suite?

balloob: No — it stays scoped to the language. `testpaths = ["tests"]` in `pyproject.toml` already means a bare `pytest` collects the whole `tests/` tree, so adding the explicit `tests` path doesn't widen collection. The only thing it changes is that `tests/conftest.py` is now loaded as an *initial* conftest, so `pytest_addoption("--language")` is registered and `pytest_generate_tests` can parametrize the `language` fixture down to the requested language.

Measured on this branch:
- `pytest tests --language en --co -q` → **357 tests collected**
- `pytest tests --co -q` (no `--language`) → **22327 tests collected**

So `script/test --language mn` runs only mn's tests, exactly as before — the difference is that today it errors out with `unrecognized arguments: --language` instead of running.

If you'd prefer not to touch `script/test`, an alternative is to register the option in a repo-root `conftest.py` (or move `pytest_addoption` there) so it loads regardless of the invocation path — happy to switch to that if you like.

---

PR #3925, opened 2026-06-24
Migrate pt (Portuguese) to slot-combination format

balloob:
Migrates Portuguese (`pt`) to the new slot-combination format, mirroring `en`/`es`. All 29 intents migrated.

## What changed
- **TurnOn/Off, GetState, LightSet, Climate Get/Set, SetPosition, SetVolume**.
- **Full timer family** (Start/Cancel/CancelAll/Pause/Unpause/Increase/Decrease/Status).
- **media** (Next/Pause/Previous/Unpause), **vacuum** (Start/ReturnToBase), **weather, date/time, Respond, Nevermind, shopping-list, todo-list**.
- Built `rules/pt/` and `lists/pt/` (ASCII identifiers); shared top-level lists reused.
- Romance prepositional-phrase boundaries kept (avoid `{area}`/`{floor}` over-absorption).
- Deleted `tests/pt/_fixtures.yaml`; new tests self-contained.

## Validation
- `python3 -m script.intentfest validate --language pt` → `All good!`
- `python3 -m script.intentfest prune --language pt --check` → 0 dead rules/lists
- `python3 -m pytest tests/test_slot_combinations.py -q` → `13926 passed`

https://claude.ai/code/session_016BW5FhdPgJr9rzMhmXtpBS

---


tests/pt/HassGetState/domain_state.yaml
```diff
@@ -0,0 +1,126 @@
...
+    state: "not_home"
+
+tests:
+  # ---- lights: >4 matches -> truncated "e mais N" ----
+  - sentences: ["há alguma luz ligada"]
+    slots: { state: "on" }
+    response: "Sim, Luz A, Luz B, Luz C e mais 2"
+  - sentences: ["estão todas as luzes desligadas"]
+    slots: { state: "off" }
+    response: "Não, Luz A, Luz B, Luz C e mais 2 não estão off"
```
abmantis:
```suggestion
    response: "Não, Luz A, Luz B, Luz C e mais 2 não estão ligadas"
```
we need to translate the state, or avoid that by removing it from the response.

tests/pt/HassGetState/domain_state.yaml
```diff
@@ -0,0 +1,126 @@
...
+    slots: { state: "locked" }
+    response: "1"
+
+  # ---- people / zones ----
+  - sentences: ["está alguém em casa"]
+    slots: { state: "casa" }
+    response: "Sim, Ana"
+  - sentences: ["estão todos em casa"]
+    slots: { state: "casa" }
+    response: "Não, Joao não está casa"
```
abmantis: this one is grammatically wrong. I think we should remove the zone name and keep it just "Não, Joao is not".
Otherwise it would need "em" or "no/na" depending on the name of the zone, which is not easy to figure out automatically.

tests/pt/HassSetPosition/device_class.yaml
```diff
@@ -0,0 +1,18 @@
...
+tests:
+  - sentences:
+      - "põe as cortinas a 50%"
+      - "muda as cortinas para 50%"
+      - "abre as cortinas aqui a 50%"
+    slots:
+      domain: "cover"
+      device_class: "curtain"
+      position: 50
+    response: "Posição definida (curtain)"
```
abmantis:
```suggestion
    response: "Posição definida"
```
we should remove the device class from the response since it is not translated (and sounds weird at the end of the sentence)

tests/pt/HassSetPosition/floor_device_class.yaml
```diff
@@ -0,0 +1,24 @@
...
+tests:
+  - sentences:
+      - "põe as cortinas do piso Rés do Chão a 50%"
+      - "muda as cortinas do Rés do Chão para 50%"
+    slots:
+      floor: "Rés do Chão"
+      domain: "cover"
+      device_class: "curtain"
+      position: 50
+    response: "Posição definida (curtain)"
```
abmantis: Maybe use "Posição alterada" for consistency with the others

abmantis (changes requested):

balloob:
Thanks for the review @abmantis! Addressed all four in c5552e5:

- **HassGetState** (`all` response): the template was appending the raw English `state` token, producing things like "...não estão off" / "...não está locked" / the awkward person-zone case. Dropped the trailing `{{ slots.state }}` so it now renders grammatical Portuguese — "...e mais N não estão", "... não está", and "Não, Joao não está" (no zone name, as you suggested).
- **HassSetPosition** `cover_device_class`: removed the untranslated device_class token and unified it to `"Posição alterada"` (matching `default`), so `device_class`/`area_device_class`/`floor_device_class` are all consistent.
- Updated the affected tests (incl. the `device_class_state*` GetState files that render through the same `all` template).

`validate`, `check_overmatch` (0/0), and `pytest --language pt` (211 passed) are all green.

---

abmantis:
@claude check if there are other responses exposing the raw english state

balloob:
Checked all of `responses/pt/`. The only place that exposed the raw English **slot** state was the `HassGetState` `all` response (`{{ slots.state }}`), which is the one already fixed in c5552e5. Details:

- `HassGetState`: no remaining `{{ slots.state }}` anywhere. The other state references are `{{ state.state_with_unit }}` (`one`, `one_yesno`) and `{{ state.state }}` (`where`) — these render the **entity's** state, which Home Assistant localizes at runtime, and they mirror `en` exactly (en uses the same `state_with_unit`/`state.state` there). `slots.state` was the problem because it's the canonical English slot value from the sentence, never localized.
- `HassGetWeather`: references `state.state` but only as a key into the `weather_condition` translation map (`weather_condition.get((state.state | string).lower(), "")`), so the output is the Portuguese term, not the raw state.
- No other `responses/pt/*.yaml` file references `slots.state`.

So the `all` response was the only raw-English-state leak; everything else is either localized at runtime or already translated via a map.

---

abmantis (approved):

PR #3951, opened 2026-06-25
Migrate pt-BR (Brazilian Portuguese) to slot-combination format

balloob:
Migrates **pt-BR (Brazilian Portuguese)** from the legacy flat HassIL template files to the new per-intent slot-combination layout.

## What changed
- All 41 intents converted to `sentences/pt-BR/<Intent>/<combo>.yaml` with matching self-contained tests under `tests/pt-BR/<Intent>/`.
- `rules/pt-BR` and `lists/pt-BR` built mirroring `en` and pruned to only referenced items.
- `sentences/pt-BR/_common.yaml` reduced to `responses` / `skip_words`.
- Old flat `sentences/pt-BR/<domain>_<intent>.yaml` and `tests/pt-BR/<domain>_<intent>.yaml` files removed.

## Notes
Built mirroring `en`, using the already-migrated European `pt` as a same-language reference adapted to Brazilian spelling/vocabulary, with the distinct old pt-BR phrasings folded in. Romance preposition over-absorption avoided (verified via check_overmatch).

## Validation
- `intentfest validate --language pt-BR`: All good
- `intentfest check_overmatch --language pt-BR`: 0 cross-intent over-matches, 0 no-match sentences
- `intentfest prune --language pt-BR --check`: no dead rules/lists
- `pytest --language pt-BR`: 211 passed

---


responses/pt-BR/HassTurnOn.yaml
```diff
@@ -24,9 +24,13 @@ responses:
...
-      cover_all: "Todas as cortinas abertas"
-      lights: "Acendendo luzes"
-      light_all: "Acendendo todas as luzes"
-      fans: "Ligando ventiladores"
-      valve: "Aberta"
+      lights_area: "Luzes acesas"
+      lights_floor: "Luzes acesas"
+      light_all: "Todas as luzes acesas"
+      fans_area: "Ventiladores ligados"
+      cover: "Abrindo"
```
staticdev, line 31: Hoi @balloob (cc. @synesthesiam),

I spent a bit more than an hour reviewing this on a large monitor, comparing the previous implementation with the new structure.

From what I can see, the structure, the tests, and even the inputs and responses all changed at the same time. That makes it extremely difficult to validate the changes with confidence in more than 500 files.

As one example, today if I open my blinds the response is "Abertura completa", while with this change it becomes "Abrindo". If I saw this change in isolation, I would naturally look at the related tests to understand the reasoning. However, since the tests also changed alongside the implementation, it becomes much harder to verify whether the behavior changed intentionally or whether something was introduced accidentally.

I also noticed that many of the other language PRs have already been merged. I'm not against this approach—in general I'm very pragmatic, and I want us to move towards a better version of the voice assistant. At some point we need to make changes and validate them through testing.

With that in mind, I'm going to approve this PR and focus on testing it more thoroughly going forward, knowing that it introduces a significant amount of change.

My main feedback is about the reviewability of the PR. Even with AI helping generate the changes, I think this could have been split into smaller, more digestible phases so that reviewers could validate each step with much greater confidence. For example:

1. Change only the file structure while keeping behavior and tests unchanged.
2. Then update the lists and responses.
3. Finally, restructure the test files.

AI gives us the ability to make much larger changes than before, which is a great advantage. At the same time, we should remember that humans are still part of the development loop. Reviews should become easier, not harder, and smaller PRs make it much more likely that meaningful, in-depth reviews can happen.

Other than that, I think the slot combination format is a good direction. It helps cover blind spots we have today across languages and gives us a better way to keep track of the supported combinations. Thank you for putting in the work to bring this improvement.

staticdev (approved):

responses/pt-BR/HassTurnOn.yaml
synesthesiam: Changing the file structure while keeping the same behavior wasn't always possible, unfortunately. We enforced some new rules, such as `{lists}` not being allowed anymore inside optionals/alternatives (so no `(text|{list})` or `[{list}]`. So a lot of `<rules>` ended up needing to be refactored, and a lot of sentence templates had to be broken apart.

I think Claude did an incredible job considering the difficulty here, but I agree it would have been ideal if the test set itself didn't change. I expect some feedback during the next release about sentences that previously worked no longer working, but once the new workflow is in place it *should* just be a matter of adding more test sentences (and maybe additional test fixtures).  Famous last words :smile:

balloob: Thanks for the careful review and the approval — much appreciated.

On the blinds example: that change is intentional, not an accidental regression. In the slot-combination format the `HassTurnOn` cover response mirrors the English reference, which reports the in-progress action rather than the end state — en uses `cover: "Opening"` / `cover_device_class: "Opening the {{ slots.device_class }}"`, and pt-BR now matches with `cover: "Abrindo"` / `cover_device_class: "Abrindo {{ slots.device_class }}"`. So "Abertura completa" → "Abrindo" is the deliberate alignment with en. (Agreed, though, that this is exactly the kind of behavioral delta that's hard to spot when structure + lists + responses + tests all move together.)

On the main point — fully fair. Splitting into phases (1: restructure files only, behavior/tests unchanged → 2: lists + responses → 3: tests) would have made each step far easier to validate, and keeping humans comfortably in the review loop matters more, not less, as the changes get bigger. I'll take that to heart for the remaining work. Thanks again for the thorough look and the thoughtful feedback.

---

PR #3960, opened 2026-06-27
Add Bulgarian intents for media_player, fan, vacuum, lawn_mower and more

hristo-atanasov:
Add Bulgarian (bg) sentences for media_player, fan, vacuum, lawn_mower, todo and shopping_list intents


hristo-atanasov (commented):
All ok

PR #3981, opened 2026-07-09
[en] Add new response format for HassShoppingListCompleteItem

Lesekater:
Add new response format like in the [media search and play intent](https://github.com/home-assistant/intents/blob/main/responses/en/HassMediaSearchAndPlay.yaml).
https://github.com/home-assistant/core/pull/128565#discussion_r2091770222 in the connected, merged https://github.com/home-assistant/core/pull/128565.

Reopening https://github.com/OHF-Voice/intents/pull/3219 ontop of new slot-combination changes


synesthesiam:
Isn't "completed_items" a `ShoppingData` object in HA core and not a list of strings?

Lesekater:
> Isn't "completed_items" a `ShoppingData` object in HA core and not a list of strings?

Hmm. Yeah, you're right. Maybe that should be changed - we really only need the name of the completed item, right?
Maybe even simplify it to "completed_item" since all items in the response should have the same name.

I could create a pr for that in core as soon as I find the time.

PR #3986, opened 2026-07-13
[JA] Add get current time intent

nasa9084:
Add a GetCurrentTime intent for Japanese and `tests/ja/_fixtures.yaml` to pass the test.


tests/ja/_fixtures.yaml
```diff
@@ -0,0 +1,2 @@
+language: ja
+entities: []
```
synesthesiam: This file needs to be removed.

It was genuinely needed when you opened the PR, but `main` has since changed: [`tests/test_language_sentences.py`](https://github.com/OHF-Voice/intents/blob/main/tests/test_language_sentences.py#L46) now treats a missing `_fixtures.yaml` as empty. I deleted the file locally and all 281 `ja` tests still pass, so it's no longer load-bearing.

More importantly, it isn't harmless. `load_fixtures()` in [`script/intentfest/util.py`](https://github.com/OHF-Voice/intents/blob/main/script/intentfest/util.py#L236) *prefers* a monolithic `_fixtures.yaml` when one exists, instead of aggregating the inline fixtures from each `tests/ja/<Intent>/<combo>.yaml`. So `entities: []` here shadows all of them:

```
with tests/ja/_fixtures.yaml:    entities=0
without:                         entities=6, areas=2
```

That silently breaks `script/intentfest/sample.py` and `parse.py` for `ja`. `tests/en` is the only language that still uses the monolithic form.

responses/ja/HassGetCurrentTime.yaml
```diff
@@ -0,0 +1,16 @@
...
+    HassGetCurrentTime:
+      default: >
+        {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
+
+        {% if slots.time.hour == 0: %}
+        午前0時{{ minute_str }}分です
+        {% elif slots.time.hour < 12: %}
+        午前{{ slots.time.hour }}時{{ minute_str }}分です
+        {% elif slots.time.hour == 12: %}
+        午後12時{{ minute_str }}分です
```
synesthesiam: `午後12時` for noon isn't standard Japanese. The 12-hour clock in Japanese runs 午前0時–午前11時 / 午後0時–午後11時, so `午後12時` is ambiguous (in JIS and legal usage it reads as *midnight*, the opposite of what's intended here).

Suggest `正午` or `午後0時` instead.

This is inherited from the English template, where `12:xx PM` is correct — it just doesn't carry over. Note the midnight branch above is fine, and in fact makes the `hour == 0` case redundant: `hour < 12` already renders `午前0時`, so that first branch could be dropped.

responses/ja/HassGetCurrentTime.yaml
```diff
@@ -0,0 +1,16 @@
+language: ja
+responses:
+  intents:
+    HassGetCurrentTime:
+      default: >
+        {% set minute_str = '{0:02d}'.format(slots.time.minute) %}
```
synesthesiam: The `'{0:02d}'` zero-padding exists in the English template so the clock reads `1:02`. In Japanese it produces `午前1時02分` / `午前0時00分`, which isn't how the time is spoken, and TTS may voice the leading zero.

[`responses/ko`](https://github.com/OHF-Voice/intents/blob/main/responses/ko/HassGetCurrentTime.yaml) drops the padding for the same reason — `{{ slots.time.minute }}分` would be closer here. Ideally minute 0 omits `分` entirely (`午前1時です`).

Rendered across the hour boundaries as written:

```
00:00 → 午前0時00分です      12:00 → 午後12時00分です
01:02 → 午前1時02分です      13:00 → 午後1時00分です
23:59 → 午後11時59分です
```

sentences/ja/HassGetCurrentTime/default.yaml
```diff
@@ -0,0 +1,8 @@
+---
+language: "ja"
+data:
+  - sentences:
+      - "[<now>]何時"
+      - "[<now>の]時間は"
```
synesthesiam: These two templates cover the four sentences in the test file but not much beyond it. Probing the merged intents:

| matches | misses |
|---|---|
| 今何時, 現在何時, 今の時間は, 現在の時間は, 何時, 時間は | 今何時ですか, 何時ですか, いま何時, 今の時間, 現在時間は, 時間を教えて, 時刻を教えて, 今の時刻は, 現在の時刻は, 何時何分 |

`今何時ですか` is the one I'd most want to work. A few things would pick up most of that column:

- polite endings — `rules/ja/verbs.yaml` already enumerates these explicitly (`て|なさい|てください`), so it's in keeping with how `ja` is written
- make `の` optional so `現在時間は` and `今の時間` match
- a `時刻` alternative alongside `時間`
- a `教えて` form

For reference, [`sentences/zh-CN/HassGetCurrentTime`](https://github.com/OHF-Voice/intents/blob/main/sentences/zh-CN/HassGetCurrentTime/default.yaml) carries five patterns.

Separately: `何時` and `時間は` currently match on their own with no `<now>` prefix. That's fine today, but worth keeping in mind as a broad-match conflict risk as `ja` gains more intents.

rules/ja/getstate.yaml
```diff
@@ -0,0 +1,3 @@
+language: ja
+expansion_rules:
+  now: (今|現在)
```
synesthesiam: Very minor, take it or leave it: `<now>` will likely get reused by `HassGetCurrentDate`, the weather intents, and the climate get-temperature intents, so `rules/ja/common.yaml` might be a better home than `getstate.yaml`. zh-CN keeps its equivalent `current` rule in `common.yaml`.

synesthesiam (commented):
Thanks for this, and sorry for the slow review!

I merged this into current `main` and it applies cleanly. `python -m script.intentfest validate --language ja` reports **All good!**, and the full suite passes (14233 tests). It also closes a real declared gap — before this PR the validator warned `Missing sentences for usable slot combination: HassGetCurrentTime/default` for `ja`.

One thing does need to change before merge, plus a few Japanese-correctness suggestions — details in the inline comments.

**Needs changing**

- `tests/ja/_fixtures.yaml` should be removed. It was required when you opened this in July, but `main` now treats a missing `_fixtures.yaml` as empty, and keeping it actively blanks out `ja`'s per-combination fixtures for the `intentfest` scripts.

**Suggestions**

- `午後12時` for noon isn't standard Japanese — the 12-hour clock runs 午後0時–午後11時.
- The zero-padded minutes come from the English `1:02` clock format and don't read naturally in Japanese.
- The two sentence templates are quite narrow; `今何時ですか` and similar polite forms don't match.

You'll need a rebase on `main` in any case. Happy to help with any of the above if you'd like.

rules/ja/getstate.yaml
nasa9084: fixed.

sentences/ja/HassGetCurrentTime/default.yaml
nasa9084: fixed.
We actually won't say `現在時間は` so skipped it but added other usecases.

responses/ja/HassGetCurrentTime.yaml
nasa9084: fixed.

responses/ja/HassGetCurrentTime.yaml
nasa9084: fixed.
Honestly I would say `午後12` for noon, however following JIS/legal usage sounds reasonable.

tests/ja/_fixtures.yaml
nasa9084: fixed.

nasa9084:
Sorry I mistakenly made commits containing other languages during rebase/fixing conflicts and many reviewers are assigned...
Now I fixed conflicts and squashed my commits but I don't have permission to remove unnecessary reviewers.

hecktech27:
> Sorry I mistakenly made commits containing other languages during rebase/fixing conflicts and many reviewers are assigned... Now I fixed conflicts and squashed my commits but I don't have permission to remove unnecessary reviewers.

i reverted that for you :)

synesthesiam (approved):

synesthesiam:
Thanks @nasa9084! Let me know if you'd be interested in being a language leader :slightly_smiling_face:

PR #3991, opened 2026-07-19
[DE] Fix area rule preposition handling

### LINKED ISSUE #3990: [DE] rule <preposition> is missing from common.yaml
The rule "preposition" is referenced in:

```
area:
    "[(<preposition>|<artikel_bestimmt>) ]{area}|(<preposition> der|(an|in) die|in
    dem|im|am|<von_dem>) {area}|des {area}s"

```

but is never defined at least in DE.

This should be necessary for the sentence:

"Licht auf dem Balkon einschalten"

Not sure how the sentence 

*Licht im Garten einschalten"

Since preposition and article are merged.

mstu01:
Fixes #3990.

The `preposition` rule referenced by the `area` rule in `rules/de/common.yaml` was misplaced into `rules/de/covers.yaml` during the modern-syntax migration (#3887). It belongs to the shared grammar rules, so this moves it back next to `artikel_bestimmt` and removes the now empty covers.yaml. All rule files are merged per language, so the move itself is a readability fix, not a behavior change.

The actual gap behind the issue: the area rule only combined a preposition with the article "der" ("in der Küche", "auf der Terrasse"), while the dative article "dem" was limited to "in dem", "im" and "am". "Schalte das Licht auf dem Balkon ein" was not recognized. The rule now allows "dem" after any preposition (auf dem, an dem, bei dem), covered with new test sentences.

"Licht im Garten einschalten" from the issue already worked via the explicit "im" alternative.

Local checks: intentfest validate, pytest --language de (324 passed), check_overmatch (0 over-matches, 0 no-match), prune --check clean.

### DIFF
--- a/rules/de/common.yaml
+++ b/rules/de/common.yaml
@@ -3,10 +3,11 @@ expansion_rules:
   artikel_bestimmt: (der|die|das|dem|den|des)
   possessivpronom_mein: (mein|meine|meines|meiner|meinem|meinen)
   possessivpronom_unser: (unser|unsere|unseres|unserer|unserem|unseren)
+  preposition: (in|an|bei|auf)
   name: "[<artikel_bestimmt> ]{name}"
   area:
-    "[(<preposition>|<artikel_bestimmt>) ]{area}|(<preposition> der|(an|in) die|in
-    dem|im|am|<von_dem>) {area}|des {area}s"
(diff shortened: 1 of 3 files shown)


andreasbrett (approved):
Thank you, good catch!

PR #3993, opened 2026-07-19
[DE] Add vacuum area slot combinations

mstu01:
Adds the two missing German area_only combinations for HassVacuumStart and HassVacuumReturnToBase.

HassVacuumCleanArea already owns the "clean this area" phrasing in German ("die Küche saugen"), so the new HassVacuumStart sentences target the vacuum located in an area instead ("starte den Staubsauger in der Küche"), and HassVacuumReturnToBase sends it back to its base ("schicke den Staubsauger in der Küche zurück zur Basis").

Local checks: intentfest validate, pytest --language de (324 passed), check_overmatch (0 over-matches, 0 no-match), prune --check clean.


andreasbrett (approved):
LGTM, thank you for helping out here. Great to see the last puzzle pieces coming together.

PR #3994, opened 2026-07-20
[DE] Add remaining HassSetPosition slot combinations

mstu01:
German was missing three declared slot combinations, all in HassSetPosition: name_area, name_floor and floor_device_class. This PR adds them, which brings German to full coverage of the declared slot combinations.

The new sentence files reuse the position phrasing already established in the German name_only and area_device_class files (öffnen/schließen/fahren/stellen plus "auf X Prozent") and the shared area and floor rules from rules/de/common.yaml. No new rules, lists or responses were needed and no existing file was changed. Following intents.yaml, the name combos are limited to the cover domain and floor_device_class infers the cover domain from the device class words, same as the existing area_device_class file.

Each combination comes with a test file whose sentences cover every sentence template, using the same fixtures as the neighbouring German tests (Vorhang links, Wohnzimmer, Obergeschoss, Erdgeschoss, Dachboden).

Verified locally: script.intentfest validate for de passes, the full pytest run for de passes (324 tests), check_overmatch reports 0 cross-intent over-matches and 0 no-match sentences, prune reports no dead rules or lists, and report_unparsed_examples shows no new unparsed examples (the three new example sentences all parse). write_missing_slot_combinations.py now reports 0 missing combinations for German.

### DIFF
--- a/sentences/de/HassSetPosition/floor_device_class.yaml
+++ b/sentences/de/HassSetPosition/floor_device_class.yaml
@@ -1,0 +1,13 @@
+---
+# HassSetPosition - floor_device_class
+# slots: {floor}, {device_class}, {domain}, {position}
+
+language: "de"
+data:
+  - sentences:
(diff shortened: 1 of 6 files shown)


andreasbrett (approved):
Thank you, looks good!

PR #3995, opened 2026-07-20
[DE] Support natural word order in HassGetState domain queries with area and floor

mstu01:
The German domain_state_area and domain_state_floor files for HassGetState only recognized the area or floor in two positions: before the noun ("welche Lichter in der Küche sind an") or trailing after the state ("welche Lichter sind an in der Küche"). The most natural word order in German puts the prepositional phrase between the verb and the state: "welche Lichter sind in der Küche an", "ist irgendein Ventilator im Schlafzimmer an", "wie viele Türen sind im Erdgeschoss abgeschlossen". None of these parsed, and the report_unparsed_examples script listed all 24 documented examples of the two files as no match.

This PR adds one alternative sentence per any/which/how_many block (lights, fans, switches and locks) covering that middle position, for both area and floor. The new lines reuse the existing are_any/which/how_many and area/floor expansion rules and keep the tight on_off_states/lock_states constraints, no rules or lists were changed. The area test fixtures gain the Schlafzimmer, Büro and Garage areas because the documented examples reference them and example recognition resolves areas from the combination's own fixtures.

Every affected block got new test sentences in the existing test groups, written first and confirmed failing with "Sentence was not recognized" before the grammar change. Verified afterwards: intentfest validate is clean, the full German test suite passes (324 tests), check_overmatch reports 0 cross-intent over-matches and 0 no-match sentences, prune reports no dead rules or lists, and report_unparsed_examples for German drops from 48 to 24 problems with all 24 domain_state entries gone.

### DIFF
--- a/sentences/de/HassGetState/domain_state_area.yaml
+++ b/sentences/de/HassGetState/domain_state_area.yaml
@@ -11,6 +11,7 @@ data:
   # ---- lights ----
   - sentences:
       - "(ist|sind|gibt es|stehen)[ <irgend>] <area> (Licht[er]|Lampe[n]|Beleuchtung[en]) {on_off_states:state}"
+      - "<are_any> (Licht[er]|Lampe[n]|Beleuchtung[en]) <area> {on_off_states:state}"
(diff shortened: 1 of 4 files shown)


hecktech27 (approved):
Thanks for this - looks good!

PR #3996, opened 2026-07-20
[DE] Fix unparsed HassGetState examples for device class and name queries

mstu01:
The unparsed-examples report for German currently lists 16 documented examples from the HassGetState slot combinations device_class_state_area, device_class_state_floor, name_state, name_state_area and name_state_floor as not recognizable. They fall into two groups.

The first group is a word order gap. Questions like "ist irgendein Fenster im Schlafzimmer offen", "welche Fenster sind im Obergeschoss offen" or "ist das Garagentor in der Garage offen" put the area or floor between the subject and the state, which is the most natural way to ask this in German. The templates only covered "sind im Schlafzimmer Fenster offen" and "ist irgendein Fenster offen im Schlafzimmer". This change adds the missing alternative to the affected blocks, reusing the existing area, floor, are_any, which and how_many expansion rules. No rules or lists were changed.

The second group is a fixture mismatch. The examples "ist das Fenster offen", "ist das Büro belegt", "ist das Fenster im Schlafzimmer offen" and "ist das Schlafzimmerfenster im Obergeschoss offen" use names that had no matching entity in the combination's test fixtures, so the example could never resolve. The English fixtures back every example name with a test entity, so the German test files now do the same: binary_sensor entities named Fenster, Büro and Schlafzimmerfenster were added, existing entities and tests were left untouched.

Every one of the 16 examples is now also a test sentence, added before the fix and observed failing with "Sentence was not recognized", then passing after the change.

Verification: intentfest validate for de ends with "All good!" and no new warnings, the full German pytest run passes (324 tests), check_overmatch reports 0 cross-intent over-matches and 0 no-match sentences, prune reports no dead rules or lists, and report_unparsed_examples for de drops from 48 to 32 no-match entries with no new entries. The remaining 32 belong to other slot combinations.

### DIFF
--- a/sentences/de/HassGetState/device_class_state_area.yaml
+++ b/sentences/de/HassGetState/device_class_state_area.yaml
@@ -13,6 +13,7 @@ data:
   - sentences:
       - "(ist|sind|gibt es|stehen)[ <irgend>] <area> {cover_classes:device_class} {cover_states:state}"
       - "<are_any> {cover_classes:device_class} {cover_states:state} <area>"
+      - "<are_any> {cover_classes:device_class} <area> {cover_states:state}"
     example: "ist irgendein Fenster im Schlafzimmer offen"
     inferred_domain: "cover"
     response: "any"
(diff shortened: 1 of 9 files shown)


hecktech27 (approved):
looking good as well!

PR #3997, opened 2026-07-20
[DE] Fix unparsed volume control examples

mstu01:
The unparsed examples report listed eight German volume control examples as no-match:

- HassSetVolume/name_only: "Stelle die Lautstärke von dem Fernseher auf 50 Prozent", "Stelle den Fernseher auf maximale Lautstärke"
- HassSetVolumeRelative/name_only: "den Fernseher leiser", "Fernseher um 20 Prozent lauter", "Fernseher um 20 Prozent leiser"
- HassSetVolumeRelative/floor_only: "die Lautstärke im Erdgeschoss leiser", "die Lautstärke im Erdgeschoss um 20 Prozent lauter", "die Lautstärke im Erdgeschoss um 20 Prozent leiser"

The grammar already covers all eight sentences. The problem was in the test fixtures: example recognition resolves name and floor slots from the slot combination's own test file, and those fixtures only contained a media_player named "TV" and a floor named "Obergeschoss", while the examples use "Fernseher" and "Erdgeschoss".

This PR adds a "Fernseher" media_player entity and an "Erdgeschoss" floor to the fixtures of the three affected test files and adds test blocks that cover the example sentences plus a few close variants. No sentence templates were changed and no existing tests were modified, the diff is insertions only. This follows the English files, where examples use the exact fixture names and these combos have zero problems.

Verification, all run for language de: intentfest validate passes with no new warnings, the full test suite passes (324 passed), check_overmatch reports 0 cross-intent over-matches and 0 no-match sentences, prune reports no dead rules or lists, and report_unparsed_examples drops from 48 to 40 no-match with exactly these eight entries gone and no new ones.

### DIFF
--- a/tests/de/HassSetVolume/name_only.yaml
+++ b/tests/de/HassSetVolume/name_only.yaml
@@ -5,6 +5,8 @@ language: de
 entities:
   - name: TV
     domain: media_player
+  - name: Fernseher
+    domain: media_player
 tests:
   - sentences:
       - Stelle TV auf maximum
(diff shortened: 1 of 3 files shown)


hecktech27 (approved):
Thanks for these adjustments as well!

PR #3998, opened 2026-07-20
[DE] Speak German state names in HassGetState responses

mstu01:
The German HassGetState responses mixed English words into spoken German answers. Asking "sind alle Türen abgeschlossen" answered "Nein, Hintertür ist nicht locked", and "ist die Flurlampe an" answered "Nein, off". The "all" response inserted the raw slot state and the "one_yesno" response inserted the raw entity state, so every non-numeric state came out in English.

This change adds a small state name mapping inside the two affected templates in responses/de/HassGetState.yaml, following the pattern already used by the weather condition mapping in responses/de/HassGetWeather.yaml. The mapping covers the closed set of state values reachable through the German sentence lists (on_off_states, cover_states, lock_states, bs_open_states, bs_occupancy_states) plus the person states home and not_home, using the same words the German sentences and the existing "where" response already use: an, aus, offen, geschlossen, am Öffnen, am Schließen, abgeschlossen, aufgeschlossen, zu Hause, unterwegs. Unmapped values, for example zone names or sensor values with units, pass through unchanged. The wording is invariant for singular and plural, so the existing ist/sind handling keeps working.

The 16 test expectations that asserted the mixed language output were updated to the German wording first and failed against the old templates, then passed after the template change. Verified with intentfest validate, the full German pytest suite (324 passed), check_overmatch (0 over-matches, 0 no-match), prune in check mode, and report_unparsed_examples (unchanged at the upstream baseline).

### DIFF
--- a/responses/de/HassGetState.yaml
+++ b/responses/de/HassGetState.yaml
@@ -13,10 +13,22 @@ responses:
         {% endif %}
 
       one_yesno: |
+        {% set state_names = {
+          'on': 'an',
+          'off': 'aus',
+          'open': 'offen',
+          'closed': 'geschlossen',
+          'opening': 'am Öffnen',
+          'closing': 'am Schließen',
+          'locked': 'abgeschlossen',
+          'unlocked': 'aufgeschlossen',
+          'home': 'zu Hause',
+          'not_home': 'unterwegs'
+        } %}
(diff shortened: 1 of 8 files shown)


hecktech27 (approved):
thanks for these additions as well! :)

PR #4031, opened 2026-07-30
[DE] Add imperative templates for HassVacuumCleanArea

mstu01:
German `HassVacuumCleanArea` only matched the verb-final word order, for example "das Wohnzimmer reinigen". The imperative that people actually say to a voice assistant, "sauge das Wohnzimmer" or "sauge hier", did not match at all.

This adds one imperative template to `area_only` and one to `context_area`, both reusing the existing `<saugen>` rule from `rules/de/common.yaml`, plus test sentences for them. English has the mirror image: it ships the imperative and not the infinitive, so this gives German both.

### DIFF
--- a/sentences/de/HassVacuumCleanArea/area_only.yaml
+++ b/sentences/de/HassVacuumCleanArea/area_only.yaml
@@ -7,5 +7,6 @@ language: "de"
 data:
   - sentences:
       - "<area> (reinigen|putzen|[staub[ ]]saugen|wischen)"
+      - "<saugen> <area>"
     example: "das Wohnzimmer reinigen"
     response: "default"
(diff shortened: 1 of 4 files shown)


andreasbrett (changes requested):
As mentioned in your previous PRs I won't spend time reading through large manifests when the PR has only 10 lines of changed code. I assume you've spent drastically less time "writing" this PR than is required processing it.

Please read https://developers.home-assistant.io/blog/2026/07/20/ai-policy/

mstu01:
Fair point, sorry for the wall of text. I have shortened the description to what actually matters and will keep the length proportionate to the change from now on.

Short version: German only matched "das Wohnzimmer reinigen", not the imperative "sauge das Wohnzimmer" or "sauge hier". I added one imperative template to `area_only` and one to `context_area`, both reusing the existing `<saugen>` rule, plus tests for them.

andreasbrett (approved):
Thank you, I think that really helps getting better UX

PR #4033, opened 2026-07-30
[DE] Add "Welche Temperatur" phrasings for HassClimateGetTemperature

mstu01:
Asking "Welche Temperatur hat das Wohnzimmer?" did not match anything in German. The intent already covers "Wie hoch ist die Temperatur ...", "Wie warm ist es ..." and the bare "Temperatur ...", but not the "Welche Temperatur ..." question form, which is just as common in spoken German.

This adds one template to each of the four slot combinations, plus test sentences for them. The target temperature block is untouched, so "Auf welche Temperatur ist die Heizung eingestellt?" still answers with the target temperature and not the current one.


andreasbrett (approved):
LGTM

PR #4004, opened 2026-07-20
update test script to support new file/folder structure as well

hecktech27:
the current test script enforces files to be located in e.g. `<sentences>/<lang>/<file>`
this pr updates the script to allow another subfolder as well.

this is used by e.g. "vscode tasks run tests for current file/language"

PR #4034, opened 2026-07-30
add rules and lists directory to test script

hecktech27:
follow-up for #4004 
this PR adds the rules and lists directory to the test_file script.
before the transition to the new structure it was possible to test lists and rules with this script so I think it makes sense to add this functionality again.


synesthesiam:
Thanks!

PR #4036, opened 2026-07-31
CI: update language labeler to support tests/ and lists/ directory changes

hecktech27:
currently the language labeler does not assign language labels when there are changes in the lists/ and/or rules/ directory only.


andreasbrett (approved):
LGTM. Hope I'm not overstepping here @synesthesiam but this seems like a no-brainer to me. Non invasive, easily reversed if it fails, obvious fix.

PR #4050, opened 2026-08-04
[fi] Add HassVacuumCleanArea context_area sentences

synesthesiam:
Adds the `context_area` slot combination for `HassVacuumCleanArea` in Finnish.

This is one of the combinations marked **usable** in `intents.yaml`, and it was the only usable combination still missing for `fi`.

`HassVacuumCleanArea` did not exist for Finnish at all, so this also adds the response file.

## Changes

| File | Change |
| --- | --- |
| `sentences/fi/HassVacuumCleanArea/context_area.yaml` | new |
| `tests/fi/HassVacuumCleanArea/context_area.yaml` | new |
| `responses/fi/HassVacuumCleanArea.yaml` | new |

## Sentences

The template reuses the existing `<here>` expansion rule from `rules/fi/common.yaml`, so no rule changes are needed:

```
(imuroi|siivoa) <here>
```

This covers imuroi täällä, siivoa tässä, imuroi tässä huoneessa, siivoa tässä tilassa.

## Scope

Only the usable combination is added here. `area_only` and `name_area` are `complete`-tier and remain unimplemented.

## Verification

- `python -m script.intentfest validate --language fi` → All good!
- `pytest tests --language fi` → 249 passed

### DIFF
--- a/responses/fi/HassVacuumCleanArea.yaml
+++ b/responses/fi/HassVacuumCleanArea.yaml
@@ -1,0 +1,5 @@
+language: "fi"
+responses:
+  intents:
+    HassVacuumCleanArea:
+      default: "Imuroidaan {{ slots.area }}"
--- a/rules/fi/common.yaml
+++ b/rules/fi/common.yaml
@@ -37,6 +37,7 @@ expansion_rules:
   jotkut: (jotkut|mitkään|yhdetkään)
   kaikki: (kaikki|jokainen|jokikinen)
   here: (täällä|tässä|tässä huoneessa|tässä tilassa)
+  this_area: (tämä (huone|tila))
   alueen: (alueen|paikan|huoneen|tilan)
   alueessa: (alueessa|paikassa|huoneessa|tilassa)
   alueesta: (alueesta|paikasta|huoneesta|tilasta)
--- a/sentences/fi/HassVacuumCleanArea/context_area.yaml
+++ b/sentences/fi/HassVacuumCleanArea/context_area.yaml
@@ -1,0 +1,11 @@
+---
+# HassVacuumCleanArea - context_area
+# example: imuroi täällä
+# context_area: true
+
+language: "fi"
+data:
+  - sentences:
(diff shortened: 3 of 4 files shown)


salleq:
I would add "tämä huone" and "tämä tila" as in "siivoa tämä tila" and "imuroi tämä huone" - requires additions to rules/fi/common.yaml "here" rule - just seeing if claude does this automatically ;) 

otherwise looks good to me.

salleq (approved):

synesthesiam:
Done in 694f40a — with one deliberate deviation from the suggestion.

I added the forms in a separate rule instead of extending `here`:

```yaml
here: (täällä|tässä|tässä huoneessa|tässä tilassa)
this_area: (tämä (huone|tila))
```

and the template is now `(imuroi|siivoa) (<here>|<this_area>)`.

The reason: `<here>` is locative and is also used in `HassTurnOn/domain_only.yaml` and `HassTurnOff/domain_only.yaml`. Adding the nominative/accusative forms there would make those templates also generate ungrammatical sentences like *"kytke valot päälle tämä huone"*, which matters because these templates are expanded into training sentences, not only used for matching. `imuroi`/`siivoa` take a direct object, so the object form seemed to belong in its own rule.

Happy to fold it into `here` instead if you'd rather keep a single rule — your call as fi leader.

Tests added: `imuroi tämä huone`, `siivoa tämä tila`. `validate --language fi` → All good!, `pytest tests --language fi` → 249 passed.

salleq:
> Done in [694f40a](https://github.com/OHF-Voice/intents/commit/694f40acd2a4e2b11509eb27ab9c3501bc14cd51) — with one deliberate deviation from the suggestion.
> 
> I added the forms in a separate rule instead of extending `here`:
> 
> ```yaml
> here: (täällä|tässä|tässä huoneessa|tässä tilassa)
> this_area: (tämä (huone|tila))
> ```
> 
> and the template is now `(imuroi|siivoa) (<here>|<this_area>)`.

I think your approach is better. Looks good to me!

PR #4085, opened 2026-08-07
Render the real area name in context_area test responses

synesthesiam:
## Problem

Slot combinations with `context_area: true` get their area from the voice satellite. Home Assistant puts the area's **name** in the intent context:

```python
# homeassistant/components/conversation/default_agent.py
return {"area": {"value": satellite_area.name, "text": satellite_area.name}}
```

but the test harness injected the sentinel string `__context_area__`. Every expected response that references `{{ slots.area }}` therefore had to spell out that placeholder instead of what the user actually hears:

```yaml
response: "Cleaning __context_area__"
```

The response templates were always right — only the tests read wrong. A few test files had already worked around it by literally naming an area `__context_area__` so their fixture entities would match.

## Fix

The satellite's area is now the **first area declared in the test file**, falling back to the sentinel for test files that declare none:

```yaml
areas:
  - name: "Living Room"

tests:
  - sentences:
      - "vacuum in here"
    response: "Cleaning Living Room"
```

45 test files updated:

- Expected responses spell out the real name (`Cleaning Living Room`, `正在打掃客廳`, `Ha d Chuchi-Lampe uusgschaute.`).
- 16 files that declared no areas got one, reusing a name that language's other tests already use (`Cuina`, `Salon`, `Bucatarie`, `客厅`, …). For `ru` I followed the existing convention of naming the area in the grammatical case the response needs — `Кухне` for "Свет в Кухне включен", `Кухня` after the colon in the vacuum response.
- 5 files that used `__context_area__` as an area name now use real names.
- Because the satellite's area now filters entities the same way a spoken area does, `he/HassClimateGetTemperature` and both `pl` `domain_only` files needed their entities assigned to that area — otherwise `{{ state.state }}` renders on `None`.

`check_overmatch.py` and `example_highlight.py` also inject the sentinel, but only for recognition — they never render a response, so it stays invisible there.

## Testing

- `pytest tests/` — 13715 slot-combination tests + 102 others pass
- `python -m script.intentfest validate` — "All good!" (only pre-existing missing-sentence warnings)


davefx (approved):

andreasbrett (approved):

staticdev (approved):

TheFes (approved):

tetele (approved):

tests/ru/HassSetPosition/device_class.yaml
```diff
@@ -4,13 +4,13 @@
 language: "ru"
 
 areas:
-  - name: "__context_area__"
+  - name: "Гостиная"
 
 entities:
   - name: "Штора в спальне"
```
HepoH3:
```suggestion
  - name: "Штора"
```

HepoH3: This suggestion fixes an inconsistency between the entity name and the area name that was updated in this PR.

dontinelli (approved):
ok for de-CH

LubosKadasi (approved):
sk-SK OK. Thank You

v1k70rk4 (approved):

MTrab (approved):

leranp (approved):

duhow (approved):

piitaya (approved):

HepoH3 (approved):

PR #4088, opened 2026-08-08
[he] gender-neutral sentences and responses

ofirsnb:
In Hebrew, there are different pronounces when addressing male or female.
So, if my wake-word is "Alexa" (female), I should be able to:
1. speak to her in a female form.
2. get responses from her as gender-neutral (preferably fixed-nouns) - since the responses are static and may be returned by either male or female assistant models.

This PR is an expansion to #2152  ,  that includes:
1. adjustments to the new intents that were added overtime (`חדש|המשך` to `חדש|חדשי|המשך|המשיכי` , etc.) .
2. changing the responses to be neutral-gender (`מצטער, לא הצלחתי להבין את הבקשה` to `סליחה, לא הצלחתי להבין את הבקשה` , etc.) .

Thank you very much!


leranp:
Thanks for the help! However, there's a fundamental issue with this approach in Home Assistant:
​As far as I know, Home Assistant's native voice assistant currently cannot distinguish whether a specific wake word (hotword) was meant for a female persona (e.g., "Alexa") or a male persona.

​Because HA doesn't pass hotword metadata or context down to the pipeline, the system cannot determine the target persona based on who/what triggered it. In a gender-inflected language like Hebrew, when the input itself lacks explicit gender markers, it ends up dynamically swapping or guessing between masculine and feminine responses—regardless of which hotword was used.

​To properly fix this, it seems like a significant change in the Home Assistant Core pipeline would be required so that wake words can route to specific persona settings or prompt context.

ofirsnb:
Thanks for looking at this!

We agree on the facts - but I'd argue they are the reason for this PR rather than an objection to it.

You're right that HA has no notion of persona gender - nothing tells the pipeline whether the assistant should be male or female. That is exactly why both halves of this PR are needed, because currently the assistant's gender is baked into both directions of the conversation. The "dynamically swapping or guessing between masculine and feminine" you describe is exactly the behavior it eliminates - afterwards there is nothing to guess, because the responses no longer inflect for the speaker (`מצטער` to `סליחה` and so on) ,  while keeping the inputs (sentences) addressee-aware by supporting both male and female (`המשך` and `המשיכי`,  and so on). 

It's also forward-compatible. If Core later routes persona context down the pipeline, accepting both forms on input
stays correct, neutral responses stay correct, and per-persona variants can be layered on top - nothing here would
need undoing.

BTW,  it follows the exact solution other languages (with identical issue) implemented - like `de` / `it` / `es` etc.
I don't believe a Core change like this is going to be implemented,  just for a few languages - especially if it can be solved by the current mechanism.   Could be wrong of course,  but that's the current status,  so I'm just adapting and aligning. 

I believe we should have it here as well,  since there is already a fully working mechanism to overcome the described issue, and of course to improve the support and conversation-fluidity for the Hebrew speakers:)   One more BTW - it won't be a merge-and-forget (at least by me), since my family uses the assist daily in Hebrew,  so I would like to further improve and keep it up-to-date overtime. 

Eagerly waiting for your reply! Thank you

leranp:
Thanks for the detailed explanation! That makes a lot of sense.
Looking at it from the perspective of pragmatism and feature parity with other languages (like DE/IT/ES), using gender-neutral phrasing (like סליחה instead of מצטער) to avoid speaker-inflection issues while accepting both gender forms for inputs is definitely the best immediate solution.
I also appreciate that this approach is forward-compatible and won't block any potential future Core updates regarding persona routing.
Thanks for taking the initiative and committing to maintaining this for the Hebrew-speaking community. Looking forward to seeing this merged!

leranp (approved):

ofirsnb:
Thank you very much!  Glad to hear and fully agree.
I really appreciate you taking the time to dig into the issue,  and very thankful for the review and approval ! 
Looking forward as well:)

PR #4114, opened 2026-08-18
[pt-BR] Add "depois de" to the timer

mhalano:
I already added in the PR #3634, but it seems there was a restructure and my change was lost, so I'm adding again following the new structure.


staticdev (approved):
Thanks again @mhalano ! Yeah, last change broke some stuff (partly my fault since I reviewed).. I am also on fixes.

PR #4115, opened 2026-08-18
[pt-BR] IncreaseTimer missing slot combinations

staticdev:


sentences/pt-BR/HassIncreaseTimer/area_hours.yaml
```diff
@@ -2,7 +2,7 @@
 language: pt-BR
 data:
   - sentences:
-      - "(aumenta[r]|adiciona[r]|acrescenta[r]|junta[r]) <nb_hours> hora[s] (ao|no|ao meu|no meu) <timer> (em|na|no|da|do) {area}"
```
staticdev, line 5: @synesthesiam do you see a way I could use lists for adding the possible list of verbs in a sentence? If yes, is there some example in other language for this use-case?

I am basically repeating the same in all the slot combination files of the intent, not really DRY.

PR #4116, opened 2026-08-19
[fi] Add sytytä and sammuta for lights

jay-jay-s:
Add natural Finnish light control commands:

- "sytytä {name}" for HassTurnOn
- "sammuta {name}" for HassTurnOff

Both commands are limited to the light domain.

### DIFF
--- a/rules/fi/common.yaml
+++ b/rules/fi/common.yaml
@@ -12,6 +12,7 @@ expansion_rules:
   kytkimet: (kytkimet|vivut)
   kerro: (kerro|näytä|sano)
   kytke: (kytke|laita|aseta)
+  sytytä: sytytä
   käynnistä: (käynnistä|starttaa|aja)
   pysäytä: (pysäytä|sammuta|stoppaa)
   avaa: (avaa|aukaise|nosta)
--- a/rules/fi/common.yaml
+++ b/rules/fi/common.yaml
@@ -12,8 +12,6 @@ expansion_rules:
   kytkimet: (kytkimet|vivut)
   kerro: (kerro|näytä|sano)
   kytke: (kytke|laita|aseta)
-  sytytä: sytytä
-  sammuta: sammuta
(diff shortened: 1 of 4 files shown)


rules/fi/common.yaml
```diff
@@ -12,6 +12,8 @@ expansion_rules:
   kytkimet: (kytkimet|vivut)
   kerro: (kerro|näytä|sano)
   kytke: (kytke|laita|aseta)
+  sytytä: sytytä
```
salleq: Expansion rule isn't required since it's only one word. I would either:
- remove the expansion rule totally and just use "sytytä {name}" instead or
- add it to "kytke" expansion rule. "kytke valot" makes sense too. However I don't know if 'sytytä' is applicable to anything else but lights. Maybe a furnace or something like that

rules/fi/common.yaml
```diff
@@ -12,6 +12,8 @@ expansion_rules:
...
+  sammuta: sammuta
```
salleq: Same thing here, only one word. Either for lights I would just use "sammuta {name}" or use expansion rule "pysäytä" instead - however the rest of the words in the rule don't fit lights at all.

salleq: actually, there is already:
 - "<pysäytä> {name}"
sentence for lights (<pysäytä> includes sammuta). so it would be duplicate and cannot be added and shouldn't pass tests.

salleq:
Forgot to mention that test sentence need to be added as well - as it seems you're doing your first commit here just shout out if you need help @jay-jay-s

jay-jay-s:
Thanks! I've removed the unnecessary expansion rules and the duplicate sammuta sentence, changed sytytä to a literal light-only sentence, and added a test sentence. Thanks for the guidance!

synesthesiam:
Merging so this can get into the next beta :slightly_smiling_face:

PR #4132, opened 2026-08-25
[nl] Add Speech-to-Phrase lean sentence blocks

synesthesiam:
Speech-to-Phrase builds a constrained FST grammar instead of matching with
hassil, so it consumes the `speech_to_phrase`-tagged subset of each slot
combination rather than the full block. Only English had those blocks, so the
add-on could not serve Dutch even though a Dutch acoustic model exists.

This adds **53 tagged blocks** covering the same 26 intents as English.

### How the sentences were chosen

Every sentence is a **restriction of Dutch's own untagged templates** — taken
from what the rich blocks already accept, not newly invented phrasing. Home
Assistant's grammar is therefore unchanged: `partition_speech_to_phrase` drops
the tagged block whenever an untagged sibling exists, and the subset check
proves the lean language is contained in the rich one.

Dutch ships no `HassTurnOn/floor_domain` or `HassTurnOff/floor_domain` combo
file, so those two are not covered here.

### Verification

Each block's `example` was synthesized with the Dutch Home Assistant Cloud TTS
voice and decoded through the real Speech-to-Phrase grammar (`stt_nl_citrinet_256`), then
scored on whether hassil resolves the transcript back to the command that was
spoken — not on string equality, since the decoder legitimately picks a
different in-grammar realization (articles drop, numbers are spelled out).

**53/53 examples resolve to the correct command.**

`script/test` and `python3 -m script.intentfest validate --language nl` both pass.

### Note on the shared test change

`tests/test_speech_to_phrase.py` now discovers languages from the tagged blocks
instead of a hard-coded `["en"]`, and checks containment by *matching* each lean
phrasing against the rich blocks rather than enumerating both sides. Enumeration
does not scale past English — the Spanish `HassTurnOn` blocks expand far enough
to exhaust memory — while matching is linear in the (small) lean side and gives
the same answer. Verified against a planted violation to confirm it still fails
when it should.

That change is **byte-identical in all seven of these language PRs**, so
whichever merges first makes it a no-op for the rest.

### DIFF
--- a/sentences/nl/HassCancelTimer/default.yaml
+++ b/sentences/nl/HassCancelTimer/default.yaml
@@ -8,3 +8,8 @@ data:
       - "[<my>] <timer> (stoppen|annuleren|[op] stop zetten)"
     example: "annuleer timer"
     response: "default"
+  - sentences:
+      - "stop de timer"
+    example: "stop de timer"
+    response: "default"
+    speech_to_phrase: true
--- a/sentences/nl/HassClimateGetTemperature/area_only.yaml
+++ b/sentences/nl/HassClimateGetTemperature/area_only.yaml
@@ -10,3 +10,8 @@ data:
       - "Wat is de {area}[ ]temperatuur"
     example: "wat is de temperatuur in de woonkamer"
     response: "default"
+  - sentences:
+      - "Wat is de temperatuur in [de|het] {area}"
+    example: "Wat is de temperatuur in de Keuken"
+    response: "default"
+    speech_to_phrase: true
--- a/sentences/nl/HassClimateGetTemperature/default.yaml
+++ b/sentences/nl/HassClimateGetTemperature/default.yaml
@@ -10,3 +10,8 @@ data:
       - "hoe (warm|koud|heet|koel) is het [<here>]"
     example: "wat is de temperatuur"
     response: "default"
+  - sentences:
+      - "wat is de temperatuur"
+    example: "wat is de temperatuur"
+    response: "default"
+    speech_to_phrase: true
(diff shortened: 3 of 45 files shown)


synesthesiam:
### Merging ahead of a full native-speaker review

These sentences are needed for the upcoming Home Assistant beta, so this is
going in now on the understanding that **the wording deserves a closer look from
the language leaders afterwards**. Please treat it as a starting point rather
than a finished translation, and open follow-ups freely — refining phrasing here
is expected, not a sign something went wrong.

**What is already established and should not need re-checking:**

- **No new phrasing was invented.** Every tagged sentence is a *restriction* of
  this language's own untagged templates, chosen from phrasings the rich blocks
  already accept. `tests/test_speech_to_phrase.py` proves that containment, so
  Home Assistant's grammar is unchanged and Speech-to-Phrase cannot recognise
  anything Home Assistant would not.
- **It actually decodes.** Every block's `example` was synthesized with this
  language's Home Assistant Cloud TTS voice, run through the real
  Speech-to-Phrase FST grammar, and checked that hassil resolves the resulting
  transcript back to the command that was spoken.
- CI is green: lint, the full test suite, and `intentfest validate` in the
  stricter changed-files mode.

**What genuinely would benefit from a second pass:**

- **Which phrasing is most idiomatic.** The lean set is deliberately tiny — one
  or two ways to say each command, because every extra phrasing widens the FST
  and costs recognition accuracy. *Which* one to keep is a judgement call a
  native speaker should make.
- **The generated `example:` strings.** They are valid sentences the block
  accepts, but they are rendered mechanically from the first template, so some
  have agreement warts (a singular noun after a plural number, for instance).
- Any trade-off called out in the PR description above.

**One extra note for this PR:** it also localizes two `example:` fields on
*pre-existing* groups (`HassNevermind`, `HassTimerStatus`) that still held the
English strings. CI requires every example in a touched file to be localized, so
they had to be filled in here rather than left for a separate PR.

PR #4137, opened 2026-08-26
[DE] Add imperative templates for media player mute and unmute

mstu01:
German only has verb-final templates for muting, so "Fernseher stummschalten" is recognized but "schalte den Fernseher stumm" is not. The imperative with a separable verb is the phrasing people actually use, and English already has it ("mute the TV"), so this adds it to both mute and unmute using the same pattern HassTurnOn uses in German.


andreasbrett (approved):
LGTM, thank you very much for your contribution!

PR #4138, opened 2026-08-27
Require a media noun in Speech-to-Phrase media commands

synesthesiam:
## Problem

Speech-to-Phrase's constrained STT grammar cannot score very short utterances reliably. Bare one-word transport commands — `pause`, `mute`, `skip`, `resume` — are routinely misrecognised, and they act as attractors for out-of-vocabulary audio.

## Approach

Split the affected media combos into two data blocks:

- an untagged **rich** block that Home Assistant keeps using, which still accepts the bare `pause`;
- a `speech_to_phrase: true` **lean** block that requires a media noun (`pause the music`).

The subset check in `tests/test_speech_to_phrase.py` keeps the two in sync, so the lean grammar can never accept a phrasing HA would reject.

## English

New `rules/en/media.yaml` defines `<track>` = `(song|track|item)` and `<media>` = that plus `music|movie|[tv] show[s]|media [player[s]]`. Applied to `HassMediaPause`, `HassMediaUnpause`, `HassMediaNext`, `HassMediaPrevious`, `HassMediaPlayerMute`, `HassMediaPlayerUnmute`.

Two rich blocks widen as a side effect — `pause the music` and `resume the music` previously only resolved with an area (`pause music in the living room`), never in the `default` combo.

## Other languages

Only eight languages have tagged blocks. The remaining seven get the same treatment, preferring phrasings each language's own files already contained over new grammar:

| lang | was | now |
|---|---|---|
| ca | `pausa`, `continua`, `silencia` | `pausa la música`, `continua la música`, `silencia la música` |
| cs | `pauza` | `zastav písničku` |
| de | `pause`/`pausiere`/`stopp`, `fortsetzen` | `pausiere die Musik`, `musik fortsetzen` |
| es | `pausa`, `continúa`, `silenciar` | `pausa la música`, `continúa la música`, `silencia la música` |
| fr | `pause`, `sourdine` | `arrête la lecture`, `mets le son en sourdine` |
| it | `pausa`, `riprendi` | `metti in pausa la musica`, `riprendi la musica` |
| nl | `pauzeer`, `hervat` | `pauzeer de muziek`, `hervat de muziek` |

Catalan's `default` `HassMediaPause`/`HassMediaUnpause` blocks gain the noun phrasing lifted verbatim from their `area_only` siblings, plus the two missing `tests/ca/HassMedia{Pause,Unpause}/default.yaml`.

## What is deliberately *not* changed

`HassNevermind` is an interjection, not a command. `never mind` / `egal` / `niente` / `annuleer` is what people reflexively say to abort, and there is no natural longer paraphrase — substituting one would trade recognition accuracy for phrasings users never utter. Length here is better handled by acoustic confidence thresholds on the Speech-to-Phrase side.

The single exception is Catalan's `res`, one syllable, replaced by the `deixa-ho` and `oblida-ho` the rich block already offered.

## Needs a native speaker

`cs` `HassMediaUnpause`: `pokračuj` is the only phrasing anywhere in the Czech files — `default`, `area_only` and `name_only` all lack a noun variant to select. Something like `pokračuj v přehrávání` would close it, but I did not want to invent Czech grammar.

## Testing

Full suite passes (14555 tests); `python -m script.intentfest validate` is clean for all eight touched languages.

### DIFF
--- a/rules/en/media.yaml
+++ b/rules/en/media.yaml
@@ -1,0 +1,10 @@
+---
+language: "en"
+
+expansion_rules:
+  # Media nouns, used to give transport commands something to hang on.
+  # Speech-to-Phrase blocks require one of these so its constrained STT grammar
+  # never has to score a bare one-word command ("pause", "skip", "resume"),
+  # which is too short to recognise reliably.
+  track: "(song|track|item)"
+  media: "(music|movie|[tv] show[s]|media [player[s]]|<track>)"
--- a/sentences/ca/HassMediaPause/default.yaml
+++ b/sentences/ca/HassMediaPause/default.yaml
@@ -6,12 +6,15 @@
 language: "ca"
 data:
   - sentences:
-      - "pausa[r]"
+      - "pausa[r] [[la|la meva] (música|reproducció|sèrie|reproductor)]"
       - "(atura|para|deté)[r] [la reproducció|la música]"
     example: "pausa"
     response: "default"
+  # Speech-to-Phrase: lean subset of the block above, with the media noun
+  # required so the STT grammar never has to recognise a bare "pausa".
   - sentences:
-      - "pausa"
-    example: "pausa"
+      - "pausa la (música|reproducció|sèrie)"
+      - "atura la (reproducció|música)"
+    example: "pausa la música"
     response: "default"
     speech_to_phrase: true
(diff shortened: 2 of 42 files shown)


synesthesiam:
Language leaders: I'm merging speech-to-phrase sentence changes for right now until I can get a release out of the app.

PR #4139, opened 2026-08-27
Add climate to HassTurnOn/HassTurnOff for all languages

### LINKED ISSUE #3989: Domain "climate" missing from HassTurnOn and HassTurnOff
The subject says it all:
I can no longer just turn on/off my HVAC with a voice command, because the domain "climate" is missing from any sentence in the intents HassTurnOn and HassTurnOff

synesthesiam:
Fixes #3989.

## The bug

A sentence block's `name_domains` becomes a `requires_context: {domain: [...]}` filter at load time ([`util.py:resolve_domain_context`](https://github.com/OHF-Voice/intents/blob/main/script/intentfest/util.py)), so a climate entity's name cannot match unless the block lists `climate`.

`climate` was already in the `optional`/`complete` allowlist tiers in `intents.yaml`, and #4005 added it to the shared `default` `name_domain_group` — but only **en**, **es** and **ru**'s `name_floor` referenced that group. The other 62 languages hardcoded their own domain list, so "turn on the thermostat" didn't resolve there. That matches the Russian follow-up report on the issue.

Before / after, same grammar files:

```
[en] 'turn on the thermostat'   -> HassTurnOn     (already worked)
[ru] 'включи термостат'         -> None    ->  HassTurnOn
[ru] 'выключи термостат'        -> None    ->  HassTurnOff
```

## The change

Rather than duplicate the domain list into 62 more files, this points the generic on/off-able block at the shared group wherever the two are equivalent — so the diff is mostly a deletion (361 insertions, 1581 deletions):

- **263 blocks** whose literal list matched `default` now use `name_domains: "default"`. Future domain additions become a one-line change in `intents.yaml` instead of a 260-file sweep. This also covers **cy**'s `HassTurnOff`, which spelled out the full list while its `HassTurnOn` already used the group.
- **18 blocks** keep an explicit list because their domain set genuinely differs, and gain `climate`:
  - **cs / de / de-CH** — generic block excludes `light`; they have a separate light block with different verbs
  - **mn** — `light`/`switch` only
  - **th** — same verb also opens covers and valves
- **Speech-to-Phrase blocks are untouched.** Their list is `default` minus `climate`, so collapsing them would have silently widened them.

Also adds the English climate coverage that was missing entirely: a `Thermostat` entity and on/off test groups in all six `tests/en/HassTurn{On,Off}/name_{only,area,floor}.yaml` files, expecting "Turned on/off the thermostat".

## Verification

- `pytest tests/` — **14555 passed**
- `python3 -m script.intentfest validate` — exit 0, output **byte-identical to `main`** (no new warnings)
- `pre-commit run prettier --all-files` — passed; `sr-Latn` confirmed byte-identical to freshly regenerated output
- `climate` now present in `requires_context.domain` for **all 65 languages** × both intents
- Sampled **619 sentences** from the climate-carrying blocks across all 65 languages and recognized each back against the full grammar with a climate entity — all resolved to the correct intent

## Reviewer note

This widens `{name}` matching to climate entities in 62 languages, so a climate entity can now win a name collision against a similarly-named entity in another domain. That's the same tradeoff en/es/cy already carry, and it's what the issue asks for — but it is a behavior change in every language, so it's worth a deliberate look.

### DIFF
--- a/sentences/af/HassTurnOff/name_only.yaml
+++ b/sentences/af/HassTurnOff/name_only.yaml
@@ -6,12 +6,7 @@ data:
       - "<switch> af [<the>] {name}"
       - "<deactivate> [<the>] {name}"
     example: "skakel die plafonlig af"
-    name_domains:
-      - "light"
-      - "switch"
-      - "fan"
-      - "media_player"
-      - "input_boolean"
+    name_domains: "default"
     response: "default"
 
   # lights named with a trailing "light(s)/lamp(s)" word
--- a/sentences/af/HassTurnOn/name_only.yaml
+++ b/sentences/af/HassTurnOn/name_only.yaml
@@ -7,12 +7,7 @@ data:
       - "<activate> [<the>] {name}"
       - "[<the>] {name}"
     example: "skakel die plafonlig aan"
-    name_domains:
-      - "light"
-      - "switch"
-      - "fan"
-      - "media_player"
-      - "input_boolean"
+    name_domains: "default"
     response: "default"
 
   # lights named with a trailing "light(s)/lamp(s)" word
--- a/sentences/ar/HassTurnOff/name_area.yaml
+++ b/sentences/ar/HassTurnOff/name_area.yaml
@@ -12,12 +12,7 @@ data:
       - "<turn_off> [<light> | <switch>] [ال]{name} <in> [ال]{area}"
       - "<turn_off> [<light> | <switch>] [ال]{area} [ال]{name}"
     example: "اطفئ الضوء العلوي في المطبخ"
-    name_domains:
-      - "light"
-      - "switch"
-      - "fan"
-      - "media_player"
-      - "input_boolean"
+    name_domains: "default"
     response: "default"
 
   # covers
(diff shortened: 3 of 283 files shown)


zubir2k (approved):
Reviewed on ms. Thanks

tetele (approved):

MTrab (approved):

andrejs2 (approved):
thanks @synesthesiam

dontinelli (approved):

HepoH3 (approved):

leranp (approved):

TheFes (approved):

ddxic (approved):

apo-mak (approved):

LubosKadasi (approved):

piitaya (approved):

aman-kumar-maersk (approved):

amankrokx (approved):

Rrp13 (approved):

andreasbrett (approved):

makstech (approved):

v1k70rk4 (approved):

AlanGreene (approved):
Looks good for `ga`, thanks

LaStrada (approved):

PR #4143, opened 2026-08-28
Stop English examples leaking into non-English sentence files

synesthesiam:
Two related changes: fix the generator that seeds English examples into new languages, and remove the historical residue. Split into two commits so they can be reviewed (or landed) separately.

## 1. `add_language` no longer seeds English examples

`add_language` copies each English combo file and empties its sentence lists, but kept the English `example:` value. A new language therefore starts with examples that *look* filled in, read as English, and survive untranslated — surfacing much later as `validate_localized_examples` warnings. That's the origin of the 9 "example is identical to the English example" warnings on `main` (nl, sk, th).

Now blanked. An empty example says plainly it still needs writing, and the matching `sentences/en/` file remains the reference for what the group means.

Verified by scaffolding a throwaway language: 42 intent directories, every `example:` emitted empty, no English text leaked.

## 2. Removing 358 stale English header comments

358 non-English sentence files carry a `# example:` header comment still byte-identical to the English one — English prose at the top of a Slovak or Romanian file, describing a sentence that isn't in it:

```diff
 ---
 # HassTurnOn - name_only
-# example: turn on the overhead light
 # slots: <name>

 language: "ro"
```

Nothing parses these comments and every group carries a validated `example:` key a few lines below, so nothing is lost.

**Only the untranslated leftovers go:**

| | count | action |
|---|---|---|
| header identical to English | 358 | removed |
| header localized | 1788 | **left untouched** |
| no English counterpart to compare | 328 | left untouched |

Header comments are already absent from 4440 of the 6914 non-English combo files, so this doesn't introduce an inconsistency. Removals span 24 languages, led by ro (77), ar (42), hy (33), he (24), it (20), kw (20).

Worth noting these comments are *not* the source of problem #1 — `add_language` round-trips through `yaml_dump`, which strips comments, so the generator never propagated them. They're older residue.

## Verification

- `intentfest validate` output is **byte-identical to main** — confirming the comments are inert
- `pytest tests/` — **14555 passed**
- prettier clean; `black`/`isort`/`flake8`/`pylint` clean on the changed script (the one `mypy` note about `hassil` stubs reproduces identically on `main`)

## Related

- #4141 fills the 300 missing `example:` values these warnings are about
- #4139 fixed the 30 in its own path
- The 9 identical-to-English examples that already exist still need a native speaker; this PR stops *new* ones appearing but doesn't translate the existing ones.

### DIFF
--- a/script/intentfest/add_language.py
+++ b/script/intentfest/add_language.py
@@ -54,8 +54,14 @@ def run() -> int:
     # Create sentence files based off English. Sentences live in the
     # per-slot-combination format: sentences/<lang>/<Intent>/<combo>.yaml. Each
     # English combo file is copied with its sentence lists emptied — the
-    # language-independent scaffolding (slots, name/inferred domains, response
-    # keys, and the example hint) is kept for the translator to work from.
+    # language-independent scaffolding (slots, name/inferred domains and
+    # response keys) is kept for the translator to work from.
+    #
+    # `example` is deliberately blanked rather than copied. An English example
+    # left in place looks filled-in, survives untranslated, and only surfaces
+    # later as a `validate_localized_examples` warning; an empty one says plainly
+    # that it still needs writing. Use the matching sentences/en/ file as the
+    # reference for what the group means.
     english_sentences = SENTENCE_DIR / "en"
 
     for intent_dir in sorted(p for p in english_sentences.iterdir() if p.is_dir()):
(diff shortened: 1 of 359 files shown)


TheFes (approved):

WaterInTheLake:
Wouldn't it make sense to delete all `# example:`s (except from the English files, of course)?

Surely it's more confusing to have them there than not. When I made the PR for my language, I thought maybe something was parsing those comments. But if that's not the case, I'd say it's better to remove them all.

PR #4174, opened 2026-09-07
[it] Stop matching climatizzatore/condizionatore as fan

### LINKED ISSUE #3054: [IT] Improve climate common names to avoid clashes with fan
Hi, thanks for all the great work for the italian language support

When i try to switch on or off a climate using the term "climatizzatore" or "condizionatore" HA keep trying to use a fan entity which never existed (I don't actually have any fan entity)

So for clarifying, when i ask:
`Accendi il condizionatore in camera`
Assist respond with:
`Non conosco nessun dispositivo di tipo 'fan'` (or something similar)

By looking in the default intents I understood that the terms "condizionatore" e "climatizzatore" are used either for `fan` entities and `climate` entities, which is not what I was expecting at all as I'm not sure I'll ever call a `fan` with these terms (but maybe it's a subjective thing, I don't know). In any case It seems odd to me that HA default to fail on first entity class assumption

For now I solved by removing the two terms from `fan` domain in my local `_common.yaml` file, a procedure which is hinted but basically not documented at all in the official documentation (https://www.home-assistant.io/voice_control/custom_sentences_yaml/)

Here are my configuration (in case I'm not the only one struggling with this)
The file should be saved in `config/custom_sentences/it/_common.yaml`, create sub-folders if needed
```yaml
# config/custom_senteces/it/_common.yaml
language: it
expansion_rules:
  fan: "(ventol(a|e) | ventilator(e|i) | ventilazione)"
```

sappafrancesco:
Fixes #3054.

`climatizzatore` and `condizionatore` mean air conditioner / climate
control unit, not fan. They were listed as synonyms in the shared
`<fan>` expansion rule used by `HassTurnOn`/`HassTurnOff`, so a
sentence like "accendi il condizionatore in camera" was parsed as a
request to turn on a `fan` entity. Since no fan entity exists on a
typical setup, Assist replied with a misleading "non conosco nessun
dispositivo di tipo fan" instead of doing anything useful — exactly
what the issue reports.

No language (Italian included) has a generic `inferred_domain: climate`
block for `HassTurnOn`/`HassTurnOff` — climate entities are only
reachable there by name. So the correct fix is to stop these two words
from claiming the `fan` domain, matching the workaround the reporter
already validated locally in their own `_common.yaml` override.

Also removes the two now-fixed sentences from `tests/it/_test_failures.yaml`,
per that file's own "must be removed when fixed" convention.

Checked with:
```
python3 -m script.intentfest validate --language it
pytest tests --language it                                   # 332 passed, no regressions
python3 -m script.intentfest check_overmatch --language it    # 0 cross-intent mismatches
```

### DIFF
--- a/rules/it/fans.yaml
+++ b/rules/it/fans.yaml
@@ -1,3 +1,3 @@
 language: it
 expansion_rules:
-  fan: (ventol(a|e) | ventilator(e|i) | ventilazione | climatizzator(e|i) | condizionator(e|i))
+  fan: (ventol(a|e) | ventilator(e|i) | ventilazione)
--- a/tests/it/_test_failures.yaml
+++ b/tests/it/_test_failures.yaml
@@ -1,6 +1,4 @@
 language: it
 sentences:
-  - "accendi in soggiorno il climatizzatore"
-  - "spegni in soggiorno il climatizzatore"
   - "spegni il ventilatore in soggiorno"
   - "accendi il ventilatore in cucina"


auanasgheps (approved):
Good catch, thanks!

PR #4176, opened 2026-09-07
[it] Add HassListRemoveItem, fix verb overlap with HassListCompleteItem

sappafrancesco:
Adds Italian sentences for HassListRemoveItem (removes an item from a
todo list), which currently has no Italian coverage at all. Only one
slot combination exists in intents.yaml for this intent: name_item.

Verbs: elimina/elimini/eliminare, rimuovi/rimuovere, cancella/cancelli/
cancellare, togli/togliere, leva/levi/levare, butta/butti(via). Includes
the optional "mi " politeness prefix used elsewhere in this language
(matching #4154), with the correct indicative form for -are verbs
("mi elimini", not "mi elimina").

While building this I found that the existing HassListCompleteItem/it
sentence already claimed "elimina" and "rimuovi" as synonyms for
"complete", which collides with the natural verbs for "remove". Ran
check_overmatch before touching anything to confirm: both words were
being stolen by HassListCompleteItem instead of matching the new
intent.

English keeps these fully separate (HassListCompleteItem only uses
check/complete, HassListRemoveItem only uses delete/remove), so this
fixes the Italian sentence to match that same separation: "elimina"
and "rimuov(i|ere)" are dropped from HassListCompleteItem, keeping
only genuine completion verbs (controlla, completa, spunta). The two
test sentences that relied on the old wording move to
HassListRemoveItem, where they actually belong.

Also tested against a few real-world phrasings with plural articles
and multi-word items ("rimuovi i cavolfiori dalla lista della spesa",
"togli lo yogurt alle pesche dalla lista della spesa") to make sure
the wildcard item slot and the shared "<of>" rule ("della") combine
correctly; these are now permanent test cases.

Checked with:
```
python3 -m script.intentfest validate --language it
pytest tests --language it                                   # 332 passed, no regressions
python3 -m script.intentfest check_overmatch --language it   # 0 cross-intent mismatches (was 2 before the HassListCompleteItem fix)
```


auanasgheps (approved):

auanasgheps:
Great work, thank you @sappafrancesco!

PR #4212, opened 2026-09-17
[eu] Complete HassLightSet colour, temperature and floor coverage

urtzai:
Adds the nine remaining HassLightSet slot combinations, so the intent now covers every combination declared in intents.yaml:

- colour: name_color, color_only, area_color, floor_color
- colour temperature: name_temperature, temperature_only, area_temperature, floor_temperature
- floor_brightness

Adds lists/eu/lights.yaml with the Basque colour names and the colour temperature names, plus the color and temperature responses.

The colour list carries the inflected forms (gorri, gorria, gorrira) the way fi and pl do it, so the sentence templates stay free of case suffixes and either form is understood. The numeric colour temperature uses the shared color_temperature range as {color_temperature:temperature}: written as {temperature} it silently binds to the climate list in lists/climate.yaml (0-100), which the tests caught.

Floor sentences use the locative ({floor}n) for the same reason as the HassTurnOn/HassTurnOff work: floor names are stored with their article ("Lehen solairua"), so adding the genitive -ko would not match what a user says.


urtzai:
Hi @duhow! This PR is also waiting for approval... Could you please merge it?

duhow (approved):

PR #4213, opened 2026-09-17
Fix decimal separator in Slovenian HassGetState response

RomanLipa:
Use a comma as the decimal separator in the `one` response, consistent with the default Slovenian response.

### Reproduction

With a temperature sensor reporting `16.3 °C`, Slovenian `HassGetState` currently returns:

`Temperatura zunaj na drogu je 16.3 °C`

When this response is sent to Home Assistant Cloud TTS in Slovenian, `16.3 °C` is pronounced character by character.

Testing the same TTS directly with:

`Temperatura zunaj na drogu je 16,3 °C`

produces the correct decimal pronunciation.

This change applies the existing `replace('.', ',')` conversion to the `one` response as well.


andrejs2 (approved):

RomanLipa:
Thanks for the detailed explanation and follow-up suggestions!24. sep. 2026 16:45 je oseba Andrej Sersen ***@***.***> napisala:
@andrejs2 commented on this pull request.

Adding the notes that went missing from my approval above:
Thanks for catching this! 👍
This restores behaviour we already had. In #2383, replace('.', ',') was added to the default, sensor_value and sensor_value_temperature responses. The migration to the slot-combination format (#3954) removed the sensor_value* responses, so sensor queries now resolve to one, which never had the replacement. This PR fixes that regression and makes one consistent with default.
It is also needed on the backend side: state_with_unit is built as f"{state} {unit}" from the rounded state and always uses a decimal point. Only the frontend applies locale formatting. Whatever the template renders goes straight to TTS, so the Slovenian decimal comma has to come from the response template.
Suggestion (non-blocking, for a follow-up): replace('.', ',') replaces every dot, not only the decimal separator. Non-numeric states would be affected too, e.g. an IP address 192.168.1.10 → 192,168,1,10 or a firmware version v1.2.3 → v1,2,3. The default response has the same issue. A more precise version would only replace a dot between two digits:
{{ state.state_with_unit | regex_replace('(\d)\.(\d)', '\1,\2') }}
regex_replace exists in Home Assistant, but the test harness here uses a plain jinja2.Environment, so the filter would first have to be registered in the tests. That is out of scope for this PR, so I'm happy to merge it as is and handle default and one together later.
It would be nice to add a decimal case to tests/sl/HassGetState/name_only.yaml (currently only 15 °C is covered), but that is not a blocker.

PR #4216, opened 2026-09-18
[PL] Add missing intents

Uriziel01:
This one was a huge refactoring, initial work is done, to be refined in further commits.


Uriziel01:
**While I was at it I've introduced a breaking change and phrase `usuń` can no longer be used to finish list item, it's no longer matched agains `CompleteItem` and rather matches `HassListRemoveItem`, it was never intended to do so and was oversight, this change puts it in line with other languges.**

